PR #5, opened 2025-07-04
Make MCP server configurable

mughetto:
Trivial changes to allow users to configure the preferred host, port and transport method for the MCP server.
Also patched the generator to reflect the change there.


gasvn:
Hey, we have a new ToolUniverse version that already supports this function, so we are closing this pull request. Thank you again for your contribution, and we look forward to your future contributions.

PR #19, opened 2025-10-14
Add batch processing text embedding

suyccc:


SufianTA:
Thanks for the contribution. This PR is now stale/conflicting with current main. If this is still needed, could you please rebase and add current validation results? Otherwise we may close it and revisit batch text embedding in a fresh PR.

PR #23, opened 2025-10-19
adding vllm batch inference support

RichardZhu52:


SufianTA:
Thanks for the contribution. This PR is now stale against current main, has failing checks, and conflicts with the current LLM client/provider code. Closing for now; a fresh smaller PR for vLLM batch inference would be welcome if this is still needed.

PR #26, opened 2025-10-22
feature: add EMBL-EBI Ontology Lookup Service (OLS) tools

lee-t:
Sort of an exploration into contributing tools from existing MCP servers.

The Ontology Lookup Service (OLS) is a repository for biomedical ontologies that aims to provide a single point of access to the latest ontology versions. This allows for standardization of biomedical metadata and provides a grounded mapping crucial for meta-analysis across different biomedical datasets.

My team at [CCB ](https://dbmi.hms.harvard.edu/about-dbmi/core-computational-biomedicine) has found moderate success at using [ols-mcp-server ](https://github.com/seandavi/ols-mcp-server)by Sean Davis with claude desktop. I converted the tool functions into a local tool in ToolUniverse. 

Another approach could be to simply use the [remote mcp server ](https://www.ebi.ac.uk/ols4/api/mcp) but I think this is having hosting issues.

### DIFF
--- a/examples/ols_tool_example.py
+++ b/examples/ols_tool_example.py
@@ -1,0 +1,309 @@
+#!/usr/bin/env python3
+"""
+Example usage of OLSTool for EMBL-EBI Ontology Lookup Service (OLS).
+
+This example demonstrates all major operations of the OLSTool:
+- Searching for terms
+- Getting ontology information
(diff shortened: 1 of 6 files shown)


gasvn:
Thank you so much for your contribution!

Issue #8, opened 2025-10-01
[New feature request] Add backend support to Claude and other LLM API providers


gasvn:
At ToolUniverse, we’re building agentic tools powered by diverse LLM backends. We already support GPT, Gemini, and Azure APIs, but this is just the beginning. 

We’re looking for passionate contributors to add more LLM backends and help expand the ecosystem. If you’re excited about pushing the boundaries of what agentic tools can do, this is your chance to make a real impact.

Join us in shaping the future of “AI scientists”

gasvn:
PR #411 now provides a current-main, tested generic OpenAI-compatible adapter (including buffered and SSE streaming paths), while draft #412 tracks AWS Bedrock. #411 intentionally does not close this issue: this request covers a broader multi-provider backend strategy, including native providers such as Claude, so the umbrella issue should remain open after #411 merges.

PR #160, opened 2026-04-09
Add direct OpenAI support via a separate LLM client

lnairGT:
In relation to https://github.com/mims-harvard/ToolUniverse/issues/8 This PR adds support for OpenAI (without needing Azure) via the `OpenAIClient` using `OPENAI_API_KEY` and `api_type=OPENAI`.


SufianTA:
Thanks for contributing direct OpenAI support. We implemented this in #184 with an OpenAI-compatible provider, docs, and tests, so this PR is now superseded. Closing to avoid duplicate provider implementations.

PR #168, opened 2026-04-25
fix: always open PR for upstream sync

d33disc:
## Summary
- Remove fast-forward merge path from sync-upstream workflow
- Always open a PR for upstream changes so they can be reviewed before merging
- Prevents orphan `sync/upstream-*` branches from accumulating when fork has diverged

## Context
The fork has local commits not in upstream, so `git merge --ff-only` fails every daily run, creating a new branch each time. This simplifies to a single path: check for existing PR, create one if needed.

## Test plan
- [ ] Trigger workflow manually with dry-run to verify diff summary still works
- [ ] Trigger workflow manually without dry-run to verify PR creation
- [ ] Verify duplicate PR detection still skips when one is open


d33disc:
Opened against wrong repo — recreating on fork.

PR #195, opened 2026-05-25
fix: build filter.advanced clause for filter_study_type

### LINKED ISSUE #189: `ClinicalTrials_search_studies` sends invalid `filter.studyType` param to CT.gov v2 API → 400 Bad Request
### Summary

`ClinicalTrials_search_studies` (and the underlying `ClinicalTrialsTool._run_search`) maps the `filter_study_type` argument to a query-string param `filter.studyType`, which is not a valid CT.gov v2 API parameter. Any call that passes `filter_study_type` hard-fails with HTTP 400 from CT.gov.

This appears to be a half-completed refactor: the same file already implements the correct `filter.advanced` AREA-expression handling for `filter_phase` (line 178: *"CTG API v2 uses filter.advanced for phase, not filter.phase"*), but the same treatment was not extended to `filter_study_type`.

### Affected version

`tooluniverse==1.1.11` (latest on PyPI as of 2026-05-22), `main` branch HEAD as of commit pushed 2026-05-21.

### Reproducible call

Via the package's MCP server, with `tooluniverse` 1.1.11 installed:

```json
{
  "tool": "ClinicalTrials_search_studies",
  "arguments": {
    "query_cond": "asthma",
    "filter_study_type": "INTERVENTIONAL",
    "page_size": 1
  }
}
```

### Observed behavior

CT.gov v2 returns HTTP 400 with body:

```text
`filter.studyType` is unknown parameter
```

The outgoing URL embedded in the error response is:

```text
https://clinicaltrials.gov/api/v2/studies?format=json&countTotal=true&query.cond=asthma&filter.studyType=INTERVENTIONAL&pageSize=1
```

CT.gov's v2 API has no `filter.studyType` param. The documented way to filter on study type is via the `filter.advanced` expression syntax, e.g. `filter.advanced=AREA[StudyType]INTERVENTIONAL`.

### Expected behavior

Either:

(a) `filter_study_type` is removed from the public schema (it is currently advertised as a valid param of `ClinicalTrials_search_studies`), or
(b) `filter_study_type` is mapped to a `filter.advanced` AREA clause, the same way `filter_phase` already is.

(b) is preferable — `filter_study_type` is a useful filter and the surrounding code already implements the pattern.

### Suggested one-line fix

In `src/tooluniverse/ctg_tool.py` around line 177-188, add a sibling branch to the existing `filter_phase` handling:

```python
for key, value in arguments.items():
    if value is None:
        continue
    if key == "filter_phase":
        # CTG API v2 uses filter.advanced for phase, not filter.phase
        phases = [p.strip() for p in value.split(",")]
        phase_clause = " OR ".join(f"AREA[Phase]{p}" for p in phases)
        if len(phases) > 1:
            phase_clause = f"({phase_clause})"
        advanced_clauses.append(phase_clause)
    elif key == "filter_study_type":
        # CTG API v2 uses filter.advanced for studyType, not filter.studyType
        types = [t.strip() for t in value.split(",")]
        type_clause = " OR ".join(f"AREA[StudyType]{t}" for t in types)
        if len(types) > 1:
            type_clause = f"({type_clause})"
        advanced_clauses.append(type_clause)
    elif key in _SEARCH_PARAM_MAP:
        params[_SEARCH_PARAM_MAP[key]] = value
```

The line 152 entry `"filter_study_type": "filter.studyType"` in `_SEARCH_PARAM_MAP` can be removed at the same time (the new `elif` intercepts before the dict lookup, so leaving it is harmless but misleading).

### Confirmed working workaround

Callers can bypass the broken `filter_study_type` arg today by emitting the AREA expression directly through `query_term`:

```json
{
  "query_cond": "asthma",
  "query_term": "AREA[StudyType]INTERVENTIONAL",
  "page_size": 3
}
```

This call returns interventional studies and works against the current 1.1.11 code path.

### Note: `filter_phase` enum constraint

While testing this, also noting (not part of this bug, but related): CT.gov v2 enforces enum values `NA, EARLY_PHASE1, PHASE1, PHASE2, PHASE3, PHASE4` for the phase field, and rejects combo labels like `PHASE1|PHASE2` or `PHASE2|PHASE3` as enum-invalid. Studies that span multiple phases still come back fine via the per-phase enum query; their `phases` field is an array. The current `filter_phase` implementation works correctly when callers pass only enum values; combo labels need to be split locally. If it would help, this could be called out in the docstring.

### How we hit this

We attempted to use `filter_study_type` to constrain a search to interventional studies and received the 400 above. Because the param is advertised in the MCP schema, the failure surfaces only at the network layer rather than at validation, so a caller who doesn't inspect the upstream error body could easily mistake the empty result for a legitimate zero-match query.

mvanhorn:
Fixes #189.

`ClinicalTrials_search_studies` was mapping `filter_study_type` to a query-string parameter `filter.studyType`. CT.gov v2 does not accept that key, so any call passing `filter_study_type` hard-failed with HTTP 400 (`filter.studyType is unknown parameter`).

The repro from the issue:

```json
{
  "tool": "ClinicalTrials_search_studies",
  "arguments": {"query_cond": "asthma", "filter_study_type": "INTERVENTIONAL", "page_size": 1}
}
```

led to the outgoing URL `…&filter.studyType=INTERVENTIONAL&pageSize=1`, which CT.gov rejects.

### Fix

Apply the same `filter.advanced` `AREA[…]` treatment that `filter_phase` already uses in this method. The comment one branch up already calls this out (`# CTG API v2 uses filter.advanced for phase, not filter.phase`); this PR extends the pattern to study type.

After this PR the same call sends:

```
…&filter.advanced=AREA[StudyType]INTERVENTIONAL&pageSize=1
```

Multiple comma-separated values (`"INTERVENTIONAL,OBSERVATIONAL"`) are joined with `OR` and wrapped in parentheses, mirroring how phase values are handled.

### Scope

Only `src/tooluniverse/ctg_tool.py::ClinicalTrialsTool._run_search` is touched - that is the path the issue identifies. The other `clinicaltrials_tool.py` file uses a different mechanism (`aggFilters=phase:…`) and is out of scope for this PR.

### DIFF
--- a/src/tooluniverse/ctg_tool.py
+++ b/src/tooluniverse/ctg_tool.py
@@ -149,7 +149,6 @@ class ClinicalTrialsTool(RESTfulTool):
             "query_intr": "query.intr",
             "query_term": "query.term",
             "filter_status": "filter.overallStatus",
-            "filter_study_type": "filter.studyType",
             "page_size": "pageSize",
             "next_page_token": "pageToken",
             "intervention": "query.intr",
@@ -181,6 +180,15 @@ class ClinicalTrialsTool(RESTfulTool):
                 if len(phases) > 1:
                     phase_clause = f"({phase_clause})"
                 advanced_clauses.append(phase_clause)
+            elif key == "filter_study_type":
+                # CTG API v2 uses filter.advanced for study type, not filter.studyType
+                study_types = [s.strip() for s in value.split(",")]
+                study_type_clause = " OR ".join(
+                    f"AREA[StudyType]{s}" for s in study_types
+                )
+                if len(study_types) > 1:
+                    study_type_clause = f"({study_type_clause})"
+                advanced_clauses.append(study_type_clause)
             elif key in _SEARCH_PARAM_MAP:
                 params[_SEARCH_PARAM_MAP[key]] = value
 
--- a/tests/unit/test_ctg_tool.py
+++ b/tests/unit/test_ctg_tool.py
@@ -1,0 +1,90 @@
+"""Unit tests for ClinicalTrialsTool search-param translation.
+
+The tool sits on CT.gov's v2 API, which has no `filter.studyType` (and no
+`filter.phase`) — both filters must be expressed via `filter.advanced`
+AREA clauses. These tests pin the outgoing URL shape so a future refactor
+can't silently regress the AREA-clause construction.
+"""
+
+from unittest.mock import Mock, patch
+
+import pytest
+
+from tooluniverse.ctg_tool import ClinicalTrialsTool
+
+
+def make_search_tool():
+    """Construct a search-operation ClinicalTrialsTool with minimal config."""
+    return ClinicalTrialsTool(
+        {
+            "name": "ClinicalTrials_search_studies",
+            "type": "ClinicalTrialsTool",
+            "fields": {"operation": "search"},
+            "query_schema": {},
+            "parameter": {"type": "object", "properties": {}},
+        }
+    )
+
+
+def make_empty_studies_response():
+    response = Mock()
+    response.status_code = 200
+    response.raise_for_status.return_value = None
+    response.json.return_value = {"studies": [], "totalCount": 0}
+    return response
+
+
+@pytest.mark.unit
+@patch("requests.get")
+def test_filter_study_type_builds_area_clause(mock_get):
+    """filter_study_type must NOT appear as a query param; it must go into filter.advanced as AREA[StudyType]<value>."""
+    mock_get.return_value = make_empty_studies_response()
+
+    make_search_tool().run(
+        {"query_cond": "asthma", "filter_study_type": "INTERVENTIONAL"}
+    )
+
+    params = mock_get.call_args.kwargs["params"]
+    assert "filter.studyType" not in params, (
+        "Regression: CT.gov v2 has no filter.studyType param; passing it 400s."
+    )
+    assert params["filter.advanced"] == "AREA[StudyType]INTERVENTIONAL"
+
+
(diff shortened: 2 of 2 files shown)


gasvn:
Thanks for the diagnosis and the clean fix, @mvanhorn 🙏 — verified the original CT.gov 400 reproduction in your report and that the patch produces the expected `AREA[StudyType]<value>` clause.

Two small follow-ups pushed onto this branch:

1. **Rebased onto current `main`** so the diff applies cleanly to the latest tree (was a few commits behind).
2. **Added `tests/unit/test_ctg_tool.py`** (3 tests) pinning the new behaviour so future refactors can't silently regress:
   - single value emits `AREA[StudyType]<value>` and does NOT add a `filter.studyType` query param
   - comma-separated values join with `OR` inside parentheses
   - combined with `filter_phase`, both clauses join with `AND`

`ctg_tool.py` previously had no unit tests at all, so this is mostly net-new coverage — credit fully yours on the fix commit.

Ready for review/merge from my side. CI should kick off on the new push (may need a one-click approval in the Actions tab since this is the workflow's first run for this fork).

mvanhorn:
Appreciate the merge, @gasvn. Building the filter.advanced clause for filter_study fixes the ClinicalTrials search mapping.

mvanhorn:
Appreciate you landing this @gasvn. Building the filter.advanced clause properly should make the advanced filters behave.

mvanhorn:
Thank you @gasvn for landing the filter.advanced clause fix for filter_study_type.

mvanhorn:
Thanks for the merge @gasvn! Glad the filter_study_type advanced clause is in.

PR #197, opened 2026-05-25
Rename variant-predictor-dms-benchmarking → variant-predictor-dms-validation (re-target to main)

gasvn:
## Summary

Reapplies the rename from PR #196 onto `main`. PR #196 was merged into `feature/sae-inference-and-interpretation` (its base) **after** that branch was already merged into `main` via PR #192 — so the rename ended up on a stale branch and never reached `main`. This PR re-targets the same rename at `main`.

- `tooluniverse-variant-predictor-dms-benchmarking` → `tooluniverse-variant-predictor-dms-validation`
- The skill's job is to check whether a variant-effect predictor's scores correspond to measured functional effects on a specific protein — i.e. validate the predictor on its target. "Benchmarking" implied cross-method comparison on a standard set, which is not what the skill does.

## Scope

11 references rewritten across 5 files (directory rename + frontmatter + 3 cross-skill refs + CHANGELOG + 6 test mentions). 0 stale references remain.

## Test plan

- [x] `pytest tests/tools/test_dms_skill_snippets.py tests/tools/test_alphamissense_tool.py` — 35 passed
- [x] `ruff check .` — all checks passed
- [x] `grep -r "variant-predictor-dms-benchmarking" plugin/ tests/` — no remaining matches

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -31,7 +31,7 @@ covering single-variant mechanistic interpretation through whole-protein DMS ana
   AlphaMissense + AlphaFold + UniProt + DynaMut2) with 6-category decision rule.
 - `tooluniverse-protein-structural-annotation-pdb` — wraps Structure_annotate_per_residue.
 - `MaveDB_get_effect_matrix` — orchestrates `MaveDB_*` tools + HGVS parsing.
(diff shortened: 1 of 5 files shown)


gasvn:
Subsumed by #200, which adds the source /skills/ directories and applies the dms-benchmarking → dms-validation rename via sync-skills.sh round-trip.

PR #225, opened 2026-06-03
Pin MCP server to released version so plugin upgrades also upgrade tooluniverse

gasvn:
### Problem
`plugin/.mcp.json` launches the MCP server with `uvx tooluniverse` — no version, no `--refresh`. `uvx` reuses its cached resolved environment, so it never picks up newer PyPI releases (it kept running an old 1.1.2 locally). The file is also byte-identical across plugin versions, so bumping the plugin (e.g. 1.2.1 → 1.2.4) does nothing to the MCP launch command. Result: the Claude Code plugin and the `tooluniverse` pip package were fully decoupled — a plugin auto-update did **not** upgrade tooluniverse. (No hook did it either; the only SessionStart hook just cleans up legacy skills.)

### Fix
- **Pin** `plugin/.mcp.json` to the released version: `uvx tooluniverse@1.2.4`. uvx fetches that exact version, so when the plugin auto-updates to a new version that ships a new pin, uvx pulls the matching tooluniverse.
- **Keep it in lockstep**: `scripts/release-plugin.sh` now rewrites the `tooluniverse@X.Y.Z` pin in `.mcp.json` on every bump (alongside the manifests) and stages it. So every release keeps plugin ↔ pip in sync automatically.

### Tested
- `bash -n` clean; `.mcp.json` valid JSON with `args: ["tooluniverse@1.2.4"]`.
- End-to-end: `release-plugin.sh 1.2.5 --no-push` bumped `plugin.json`, `marketplace.json`, **and** pinned `.mcp.json → tooluniverse@1.2.5` in one shot (then reverted).
- Runtime: a version-pinned `uvx` resolves to the pinned version (vs. bare `uvx tooluniverse` reusing the stale cache).

### DIFF
--- a/plugin/.mcp.json
+++ b/plugin/.mcp.json
@@ -2,7 +2,7 @@
   "mcpServers": {
     "tooluniverse": {
       "command": "uvx",
-      "args": ["tooluniverse"],
+      "args": ["tooluniverse@1.2.4"],
       "env": {
         "PYTHONIOENCODING": "utf-8",
         "PYTORCH_MPS_HIGH_WATERMARK_RATIO": "0.0",
--- a/scripts/release-plugin.sh
+++ b/scripts/release-plugin.sh
@@ -136,6 +136,7 @@ if [ "$DRY_RUN" -eq 1 ]; then
              .claude-plugin/marketplace.json; do
         [ -f "$f" ] && echo "  $f   $CURRENT → $NEW" || echo "  $f   (absent, skipped)"
     done
+    [ -f plugin/.mcp.json ] && echo "  plugin/.mcp.json   pin tooluniverse@$NEW"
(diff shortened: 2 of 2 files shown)


gasvn:
Closing — verified this is the wrong fix.

Definitive uvx testing (isolated clean env) showed that for normal users `uvx tooluniverse` already re-resolves to the latest published version on each launch (even over a cached older ephemeral env), so the plugin already auto-upgrades tooluniverse. The 'stuck on old version' symptom was a dev-machine artifact: a local `uv tool install tooluniverse==1.1.2` shadows uvx (`uvx <name>` prefers an installed tool over the ephemeral latest).

Pinning would FREEZE the version (breaking the auto-upgrade that already works), couple plugin↔pip (fighting independent small updates), and crash if the pinned pip version isn't published. Status quo (bare `uvx tooluniverse`) is correct. The release-script robustness fixes already landed separately in #224.

PR #238, opened 2026-06-04
Add primer-design skill (PCR/qPCR primers + oligo QC)

gasvn:
Discovery harness — **task-gap** (Sub-loop B).

## Gap

Four primer/oligo tools exist — `DNA_primer_design` (SantaLucia NN), `NEB_Tm_calculate`, `IDT_analyze_oligo`, `DNA_calculate_gc_content` — but **no skill** for one of the most common molecular-biology tasks.

## Skill

`tooluniverse-primer-design`:

- **`DNA_primer_design` orchestration** + the **non-obvious constraint** I reverse-engineered: `product_size_min ≤ target span ≤ product ≤ product_size_max` — with the two common errors ("target smaller than product_size_min", "product does not span target") and their fixes.
- **`NEB_Tm_calculate`** (Tm + polymerase-specific annealing temp) and **`IDT_analyze_oligo`** (GC, MW, hairpin/dimer) usage + the *don't mix Tm methods/conditions* rule.
- **Design-rule table**: length 18–24, Tm 58–62, ΔTm<3, GC 40–60, GC clamp, 3′-end, runs, qPCR amplicon 70–150 bp.
- **Explicit specificity warning**: the tools do NOT check genome specificity — always BLAST / Primer-BLAST before ordering (the #1 cause of a 'good' primer failing).
- **`scripts/primer_qc.py`**: dependency-free QC of a primer pair (length, GC%, GC clamp, 3′ self/cross-dimer, runs, rough Tm + Tm match) → PASS/WARN.

## Validation

- `DNA_primer_design` verified live → forward/reverse primers (Tm, GC, position, product_size); documented its constraint quirk from the actual errors.
- QC script verified: a designed pair → all PASS (Tm match 0.0); a deliberately bad pair → flags long runs, GC%, GC clamp, ΔTm 19°C.
- Routing rows added (both `skills/` + `plugin/skills/` mirrors).

### DIFF
--- a/plugin/skills/tooluniverse-primer-design/SKILL.md
+++ b/plugin/skills/tooluniverse-primer-design/SKILL.md
@@ -1,0 +1,79 @@
+---
+name: tooluniverse-primer-design
+description: PCR / qPCR primer and oligo design — design forward/reverse primers for a target region (SantaLucia nearest-neighbor thermodynamics), compute melting temperature (Tm) and annealing temperature (Ta), check GC content, and screen an oligo for hairpins and primer-dimers. Use when you need primers for a sequence, want to QC an existing primer pair, or need the Tm of an oligo. Covers the primer-design rules (Tm matching, GC clamp, 3'-end, length) and the tools' constraint quirks.
+disable-model-invocation: true
+---
+
+# PCR / qPCR Primer & Oligo Design
+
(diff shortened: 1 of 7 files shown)


gasvn:
Superseded by #242, which consolidates this discovery round (5 tools + 8 skills) into a single PR to avoid the router-row merge conflicts. Same content, all 62 tests passing.

PR #240, opened 2026-06-04
Add drug-synergy skill (Bliss/Loewe/HSA/ZIP/Combination Index)

gasvn:
Discovery harness — **task-gap** (Sub-loop B).

## Gap

Five `DrugSynergy_calculate_*` tools exist (Bliss, HSA, Loewe, ZIP, Chou-Talalay Combination Index) but **no skill** — only the SYNERGxDB *database* lookup is referenced (cell-line-profiling), not the **compute-from-your-own-data** tools.

## Skill

`tooluniverse-drug-synergy`:

- **Model-selection table keyed by the DATA you measured** (the core value): single effects → Bliss; effect arrays → HSA; dose-response curves + combo → Loewe / CI; full dose×dose viability matrix → ZIP — with each tool's exact (and inconsistent) param names.
- What each model's **additivity null** means + when to use it.
- The **fractional-inhibition vs %-viability scale rule** (the #1 error: `inhibition = 1 − viability/100`).
- Interpretation table including the **Combination Index's opposite direction** (CI<1 = synergy), dose-dependence, and ceiling effects.
- Gotchas: scale mismatch, ZIP/Loewe need ≥3 non-zero dose points, **model disagreement is normal** (pre-specify + report the model), synergy ≠ efficacy.
- **`scripts/synergy_reference.py`**: Bliss/HSA/additive expectations side-by-side from one dose pair.

## Validation

- Tools verified live: Bliss (expected 0.58, score 0.12, 'Strong synergy'), HSA (per-point + mean scores), Loewe (status success). ZIP returns a clear actionable error when dose points are insufficient.
- Script matches the tool exactly (Bliss expected 0.580, score 0.120) and demonstrates model disagreement (additive vs synergy call).
- Routing rows added (both `skills/` + `plugin/skills/` mirrors).

### DIFF
--- a/plugin/skills/tooluniverse-drug-synergy/SKILL.md
+++ b/plugin/skills/tooluniverse-drug-synergy/SKILL.md
@@ -1,0 +1,79 @@
+---
+name: tooluniverse-drug-synergy
+description: Drug-combination synergy analysis — quantify whether two drugs together are synergistic, additive, or antagonistic using the standard reference models (Bliss independence, HSA / highest single agent, Loewe additivity, ZIP, and the Chou-Talalay Combination Index). Use when you have measured single-drug and combination effects (inhibition/viability) and need a synergy score. Explains which model to use, what data each one needs, and how to read the score. NOT for looking up pre-computed synergy in a database (use the SYNERGxDB tool / cell-line-profiling skill).
+disable-model-invocation: true
+---
+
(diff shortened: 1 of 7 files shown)


gasvn:
Superseded by #242, which consolidates this discovery round (5 tools + 8 skills) into a single PR to avoid the router-row merge conflicts. Same content, all 62 tests passing.

Issue #247, opened 2026-06-05
Gemini provider: `return_json` agentic tools fail; deterministic tools default to temp 1.0


freddymonteiro:
[submited with claude-code]

**Version:** tooluniverse 1.2.4 (pip/uv)
**Provider under test:** Gemini (free tier), `gemini-flash-latest`
**Config:** `TOOLUNIVERSE_LLM_DEFAULT_PROVIDER=GEMINI`, `TOOLUNIVERSE_LLM_MODEL_DEFAULT=gemini-flash-latest`, `TOOLUNIVERSE_LLM_CONFIG_MODE=env_override`

I ran the full `agents`-category battery end to end on the Gemini provider (both the CLI `tu run` path and the MCP `execute_tool` path). Most tools work well. Two issues surfaced that specifically affect anyone using Gemini, which is the documented free-tier entry point. They are independent; happy to split into two issues if you prefer.

---

## 1. Bug: every `return_json: true` agentic tool fails on the Gemini provider

Any agentic tool whose config sets `return_json: true` raises immediately on Gemini:

```
$ tu run MedicalTermNormalizer '{"raw_terms": "asprin, hypertention, alzheimers"}'
❌ Error executing MedicalTermNormalizer: Gemini JSON mode not supported here
```

**Root cause.** The Gemini client hard-raises whenever JSON output is requested, in both the sync and streaming paths:

- `tooluniverse/llm_clients.py:442` (`infer`) → `raise ValueError("Gemini JSON mode not supported here")`
- `tooluniverse/llm_clients.py:514` (`infer_stream`) → same

`agentic_tool.py:122` reads `return_json` from the tool config, so the tool is dead on arrival under Gemini regardless of input.

**Tools affected (9), all in `data/agentic_tools.json` with `configs.return_json: true`:**

| Tool | Category |
|---|---|
| MedicalTermNormalizer | agents |
| QuestionRephraser | agents |
| AdvancedCodeQualityAnalyzer | agents |
| DomainExpertValidator | agents |
| ToolQualityEvaluator | agents |
| LabelGenerator | agents |
| ToolMetadataGenerator | agents |
| ToolMetadataStandardizer | agents |
| ToolRelationshipDetector | agents |

Tools with `return_json: false` (ScientificTextSummarizer, HypothesisGenerator, ExperimentalDesignScorer, the reviewer suite) all run fine on Gemini, so this is isolated to the JSON path.

[...]

PR #273, opened 2026-07-05
fix: honor return_json in Gemini client instead of hard-raising

mvanhorn:
## Summary

`GeminiClient.infer` and `GeminiClient.infer_stream` used to raise `ValueError("Gemini JSON mode not supported here")` as soon as `return_json=True`. That guard fired for every agentic tool configured with `return_json`, so the nine agents-category tools (MedicalTermNormalizer, QuestionRephraser, AdvancedCodeQualityAnalyzer, DomainExpertValidator, ToolQualityEvaluator, LabelGenerator, ToolMetadataGenerator, ToolMetadataStandardizer, ToolRelationshipDetector) failed on the Gemini provider. This change removes the two raises and teaches `_build_config` to request structured output, so those tools now work on the Gemini path.

## Why this matters

Item 1 of #247 reports that every `return_json` agentic tool fails immediately on Gemini, which is the documented free-tier entry point. The Google GenAI SDK already supports structured output via `response_mime_type="application/json"` on `GenerateContentConfig`, and the client was already building that config object, so the raise read as an unimplemented stub rather than a real platform limit. The fix mirrors the OpenAI client, which sets `response_format={"type": "json_object"}` and returns the raw text for the caller to parse: `_build_config` now accepts `return_json` (and an optional `custom_format` schema) and adds `response_mime_type="application/json"` (plus `response_schema` when a schema is supplied) to the config kwargs. `infer` and `infer_stream` forward those values and no longer raise. The text-return path and the `test_api` call site are unchanged, so JSON parsing stays with the caller and default behavior is preserved.

## Testing

Extended `tests/unit/test_gemini_client.py` (no live Gemini API call, matching the existing offline pattern with a fake key):
- `_build_config(..., return_json=True)` sets `response_mime_type == "application/json"`.
- `_build_config(..., return_json=False)` leaves `response_mime_type` unset (default text behavior).
- Existing positional `_build_config(0.7, 128)` still sets temperature and `max_output_tokens` (backward compatibility).
- Source-scan regression guard asserting the old `Gemini JSON mode not supported here` raise stays gone.

All tests in `tests/unit/test_gemini_client.py` pass.

Refs #247 (partial scope: this addresses only the `return_json` item; the per-tool temperature-default and default model-id items in the same issue are deferred to follow-ups).

### DIFF
--- a/src/tooluniverse/llm_clients.py
+++ b/src/tooluniverse/llm_clients.py
@@ -413,12 +413,22 @@ class GeminiClient(BaseLLMClient):
         self.model_name = model_name
         self.logger = logger
 
-    def _build_config(self, temperature: Optional[float], max_tokens: Optional[int]):
+    def _build_config(
+        self,
+        temperature: Optional[float],
+        max_tokens: Optional[int],
+        return_json: bool = False,
+        custom_format: Any = None,
+    ):
         kwargs: Dict[str, Any] = {
             "temperature": (temperature if temperature is not None else 0)
         }
         if max_tokens is not None:
             kwargs["max_output_tokens"] = max_tokens
+        if return_json:
+            kwargs["response_mime_type"] = "application/json"
+            if custom_format is not None:
+                kwargs["response_schema"] = custom_format
         return self._types.GenerateContentConfig(**kwargs)
 
     def test_api(self) -> None:
@@ -438,8 +448,6 @@ class GeminiClient(BaseLLMClient):
         max_retries: int = 5,
         retry_delay: int = 5,
     ) -> Optional[str]:
-        if return_json:
-            raise ValueError("Gemini JSON mode not supported here")
         contents = ""
         for m in messages:
             if m["role"] in ("user", "system"):
@@ -450,7 +458,12 @@ class GeminiClient(BaseLLMClient):
                 resp = self._client.models.generate_content(
                     model=self.model_name,
                     contents=contents,
-                    config=self._build_config(temperature, max_tokens),
+                    config=self._build_config(
+                        temperature,
+                        max_tokens,
+                        return_json=return_json,
+                        custom_format=custom_format,
+                    ),
                 )
                 return getattr(resp, "text", None) or getattr(resp, "candidates", [{}])[
                     0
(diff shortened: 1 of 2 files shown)


eeshsaxena (commented):
Dropping the hard raise is right, but this leaves `custom_format` meaning two different return types depending on the backend.

In `AzureOpenAIClient.infer`, a non-None `custom_format` switches to `chat.completions.parse` and the method returns `resp.choices[0].message.parsed.model_dump()`, so callers get a dict. The Gemini path here returns `getattr(resp, "text", None) or ...`, so callers get the raw JSON string. `AgenticTool._run` passes that value straight back (`result` in the metadata dict, or the bare return), so the same tool config yields a dict on Azure and a string on Gemini. Anything doing `result["field"]` breaks when the model is switched, and before this PR that case failed loudly instead.

google-genai populates `response.parsed` when `response_schema` is set, so the Gemini branch can match the OpenAI shape:

```python
if custom_format is not None and getattr(resp, "parsed", None) is not None:
    parsed = resp.parsed
    return parsed.model_dump() if hasattr(parsed, "model_dump") else parsed
```

Worth deciding one way or the other and stating it in the docstring, because right now `return_json=True` with no `custom_format` also differs: OpenAI asks for `{"type": "json_object"}` and returns content, Gemini sets `response_mime_type` and returns text, which happen to line up, while the `custom_format` case does not.

Same applies to `infer_stream`, where a `response_schema` on a streamed call gives you JSON fragments per chunk rather than a parseable object per chunk.

mvanhorn:
6b066c2 takes your snippet as written, so a `custom_format` call returns the dumped parsed object and Gemini matches `AzureOpenAIClient.infer`. Switching backends no longer changes what `result["field"]` does.

For `infer_stream` I went with buffering rather than per-chunk parsing: a `custom_format` call now defers to `BaseLLMClient.infer_stream`, which yields one complete result. Streaming JSON fragments that only become parseable at the end seemed worse than giving up the incremental delivery for that case.

The docstring on `infer` now spells out all three shapes, including the `return_json` case you noted lines up between backends by coincidence rather than by contract.

The six failures in `tests/unit/test_gemini_client.py` here are a missing `google-genai` in my environment and reproduce on the clean base; the two new tests mock the client and pass.

gasvn:
Thank you, @mvanhorn — your original Gemini JSON-mode commits (`17973251` and `6b066c2c`) were preserved unchanged and have now landed on `main` through #407. The final integration also updated the retired default model, handled Gemini 3.6 sampling-parameter changes, and added live/API regression coverage. GitHub has therefore recognized this PR as merged, and #247 closed automatically. Really appreciate the careful fix and follow-up on consistent `custom_format` return shapes.

PR #287, opened 2026-07-16
Fix: correct return_schema envelope mismatch for 9 tools

### LINKED ISSUE #246: tu test reports return_schema mismatch on tools returning valid live data (UniProt, RCSB, ClinicalTrials) — v1.2.4
[Submitted via claude-code]

## Summary

Several tools report `return_schema mismatch` under `tu test` even though they successfully retrieve correct, well-formed live data. The declared `return_schema` for these tools appears narrower or staler than what the live upstream APIs now return, so `tu test` produces false negatives. `tu run` / MCP `execute_tool` are unaffected (they do not enforce `return_schema`), so this is a validation-metadata bug rather than a functional one, but it makes `tu test` unreliable as a health check.

## Environment

- tooluniverse 1.2.4 (installed via `uv tool install tooluniverse`)
- Python 3.12, Linux x86_64
- CLI path (`tu test` / `tu run`)

## Affected tools (confirmed)

| Tool | `tu run` result | `tu test` result |
|---|---|---|
| `UniProt_get_entry_by_accession` | `status: success`, full P04637 / P53_HUMAN entry | ✗ `return_schema mismatch` |
| `RCSBData_get_entry` | `status: success`, 4HHB hemoglobin metadata | ✗ `return_schema mismatch` |
| `ClinicalTrials_search_studies` | `status: success`, 106 matching studies | ✗ `return_schema mismatch` |
| `ClinicalTrials_get_database_stats` | `status: success`, populated `largest_studies` | ✗ (`total_studies: null` vs schema int) |

## Reproduction

```bash
tu test RCSBData_get_entry
# ✗ example 1
#   return_schema mismatch: {...valid 4HHB payload...} is not valid under any of the given schemas (at [])

tu run RCSBData_get_entry '{"pdb_id":"4HHB"}'
# {"status": "success", "data": {"pdb_id": "4HHB", "title": "THE CRYSTAL STRUCTURE OF HUMAN DEOXYHAEMOGLOBIN ...", "resolution": 1.74, ...}}
```

Same pattern for the other three tools: the payload returned is correct and complete, but fails the declared `return_schema`.

## Example of the mismatch

`ClinicalTrials_get_database_stats` returns `total_studies: null` (and `average_byte_size: null`) while still returning a fully populated `largest_studies` array. If the schema marks `total_studies` as a required integer, the null trips validation even though the response is usable.

For `RCSBData_get_entry` and `ClinicalTrials_search_studies` / `UniProt_get_entry_by_accession`, the live payloads contain valid, expected fields but fail `is not valid under any of the given schemas (at [])`, suggesting the top-level schema definition is too strict or out of date relative to the current API responses.

## Expected behavior

`tu test` should pass for tools that return correct live data, i.e. the `return_schema` definitions should match current upstream API responses (or tolerate nullable/optional fields where the API legitimately returns null).

## Suggested fixes

- Relax/refresh the `return_schema` for the affected tools to match current API output.
- Allow nullable fields (e.g. `total_studies`) where upstream can return null.
- Consider having `tu test` distinguish "retrieval failed" from "return_schema mismatch on otherwise-valid data," so schema drift does not read as a functional failure.

Happy to provide full captured payloads for any of these tools if useful.

gasvn:
## Summary

Fixes #246. `tu test` validates a tool's response via `jsonschema.validate(result.get("data"), return_schema)` (`cli.py`) — i.e. it validates only the inner `data` payload, not the full response. The documented convention (`skills/tooluniverse-custom-tool/references/json-tool.md`) is explicit that `return_schema` must describe `result["data"]`'s shape directly, not the envelope.

The 9 tools reported in the issue (and its follow-up comment) had `return_schema` authored as the full `{"status", "data", "metadata"}` / `{"status", "error"}` envelope instead. A bare `data` payload can never satisfy a schema requiring a top-level `status` key, so `tu test` failed even on fully correct, successfully-retrieved live data — a validation-metadata bug, not a functional one, matching the issue's own diagnosis.

## Changes

- Unwrapped `return_schema` for all 9 reported tools (`UniProt_get_entry_by_accession`, `RCSBData_get_entry`, `ClinicalTrials_search_studies`, `ClinicalTrials_get_database_stats`, `gnomad_get_gene_constraints`, `GTEx_get_tissue_sites`, `gwas_search_associations`, `PharmGKB_search_drugs`, `ENCODE_search_histone_experiments`) to describe the data payload directly, per the documented convention.
- Added a disambiguating `required` field on each object-shaped success branch (e.g. `["primaryAccession"]`, `["studies"]`, `["gene"]`) so it can no longer also match the `{"error": ...}` branch — without this, `oneOf` matched both branches for any object payload and raised a different validation error.
- Widened gnomAD's `exac_constraint` from a null-only type to nullable object: the GraphQL query requests it as an object, and gnomAD legitimately returns `null` for many genes since ExAC was superseded by gnomAD, but the schema previously forbade it from ever being anything but null.
- Added `tests/tools/test_return_schema_data_shapes.py`: for each of the 9 tools, validates a live-shaped payload and an error payload both pass, and that re-wrapping the payload in the old envelope shape now correctly fails — pinning down the regression this issue described.

## Scope note

While auditing this, the same envelope-vs-data-only `return_schema` authoring mistake was found in ~554 other tools across 233 files (vs. ~2000 tools using the correct data-only convention) — same root cause, same false-positive `tu test` failure. Fixing all of those is a much larger, separate effort and out of scope here; this PR only covers the 9 tools explicitly reported in #246. Filing a follow-up issue to track the rest.

## Test plan

- [x] `tu test` passes for all 9 tools against live API data (was failing before the fix)
- [x] New regression suite (`pytest tests/tools/test_return_schema_data_shapes.py`) passes — 28 tests
- [x] Existing related tests (`test_uniprot_compact_entry.py`, `test_gwas_tool_params.py`, `test_gtex_v2_tool.py`) still pass
- [x] Full `tests/unit` suite passes with no new failures

### DIFF
--- a/src/tooluniverse/data/clinicaltrials_gov_tools.json
+++ b/src/tooluniverse/data/clinicaltrials_gov_tools.json
@@ -1084,94 +1084,86 @@
         {
           "type": "object",
           "properties": {
-            "data": {
-              "type": "object",
-              "properties": {
-                "studies": {
-                  "type": "array",
-                  "items": {
-                    "type": "object",
-                    "properties": {
-                      "nct_id": {
-                        "type": [
-                          "string",
-                          "null"
-                        ]
-                      },
-                      "brief_title": {
-                        "type": [
-                          "string",
-                          "null"
-                        ]
-                      },
-                      "status": {
-                        "type": [
-                          "string",
-                          "null"
-                        ]
-                      },
-                      "study_type": {
-                        "type": [
-                          "string",
-                          "null"
-                        ]
-                      },
-                      "phases": {
-                        "type": "array"
-                      },
-                      "enrollment": {
-                        "type": [
-                          "integer",
-                          "null"
-                        ]
-                      },
-                      "conditions": {
-                        "type": "array"
-                      },
-                      "interventions": {
-                        "type": "array"
-                      },
-                      "sponsor": {
-                        "type": [
-                          "string",
-                          "null"
-                        ]
-                      },
-                      "start_date": {
-                        "type": [
-                          "string",
-                          "null"
-                        ]
-                      },
-                      "completion_date": {
-                        "type": [
-                          "string",
-                          "null"
-                        ]
-                      }
-                    }
+            "studies": {
+              "type": "array",
+              "items": {
+                "type": "object",
+                "properties": {
+                  "nct_id": {
+                    "type": [
+                      "string",
+                      "null"
+                    ]
+                  },
+                  "brief_title": {
+                    "type": [
+                      "string",
+                      "null"
+                    ]
+                  },
+                  "status": {
+                    "type": [
+                      "string",
+                      "null"
+                    ]
+                  },
+                  "study_type": {
+                    "type": [
+                      "string",
+                      "null"
+                    ]
+                  },
+                  "phases": {
+                    "type": "array"
+                  },
+                  "enrollment": {
+                    "type": [
+                      "integer",
+                      "null"
+                    ]
+                  },
+                  "conditions": {
+                    "type": "array"
+                  },
+                  "interventions": {
+                    "type": "array"
+                  },
+                  "sponsor": {
+                    "type": [
+                      "string",
+                      "null"
+                    ]
+                  },
+                  "start_date": {
+                    "type": [
+                      "string",
+                      "null"
+                    ]
+                  },
+                  "completion_date": {
+                    "type": [
+                      "string",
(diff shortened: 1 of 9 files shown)


gasvn:
Filed #288 to track the ~554 other tools sharing this same return_schema envelope-vs-data-only mismatch, with a detection script and fix recipe.

PR #290, opened 2026-07-17
Fix #288: correct envelope-style return_schema (batch 1/5, files A-D)

### LINKED ISSUE #288: return_schema envelope-vs-data-only mismatch affects ~554 more tools beyond #246/#287
## Background

| File | Affected tools |
|---|---|
| gtex_v2_tools.json | 13 |
| esm_tools.json | 10 |
| gnomad_tools.json | 10 |
| cpic_tools.json | 9 |
| cryoet_tools.json | 7 |
| ebi_search_tools.json | 7 |
| oma_tools.json | 7 |

## Detection script

```python
import json, glob

envelope_tools = []
    tools = json.load(open(path))
        continue
    for t in tools:
            continue
        for branch in rs["oneOf"]:
            ):
                break

```

For each affected tool:

## Suggested longer-term fix

gasvn:
## Summary

Part 1 of 5 batches fixing #288 — 111 of the 552 tools with envelope-style `return_schema` beyond the 9 already fixed in #287.

## Root cause

`tu test` validates `jsonschema.validate(result.get("data"), return_schema)` — only the inner `data` payload, never the full `{"status","data","metadata"}` envelope. These tools declared `return_schema` as a `oneOf` with a branch requiring/declaring `status`, which can never be satisfied by a bare `data` payload, so `tu test` always failed regardless of how correct the live API response was.

## Fix

For each affected tool:
- Extracted the nested `oneOf[...].properties.data` sub-schema (or, where the success branch had no `data`/`error` wrapper at all, just the fields flattened alongside `status`) and promoted it to the top level.
- Normalized the error branch to the documented plain form `{"type":"object","properties":{"error":{"type":"string"}},"required":["error"]}`.
- Added a disambiguating `required` field on object-shaped success branches that had none, so the two `oneOf` branches stay disjoint from the error branch (verified programmatically: a synthetic `{"error": "..."}` payload now matches exactly one branch for every tool in this batch).

Files in this batch (alphabetically A-D): `aging_cohort`, `allen_brain`, `allen_cell_types`, `alphafill`, `alphagenome`, `alphamissense`, `ampsphere*`, `antibody_registry`, `aopwiki`, `artic`, `bgpt`, `bigg_models`, `bindingdb`, `bioimage_archive`, `biosamples`, `bmrb`, `cancerppd2`, `cbioportal`, `cdc`, `cellpose`, `classyfire`, `clinical_calculators`, `clinical_guidelines`, `clinical_tables`, `clinical_trial_stats`, `clinicaltrials_gov`, `clinvar_submitted`, `cluspro`, `coexpression_module`, `col`, `colocalization`, `complex_portal`, `compound_disease`, `compound_gene_disease`, `compound_variant`, `conoserver`, `cpic`, `crossref`, `cryoet`, `ctd`, `ctis`, `dailymed`, `dbaasp`, `dblp`, `decoupler_ulm`, `deseq2`, `dfam`, `dn_ds`.

## Validation

- All 232 changed files (across all 5 planned batches) parse as valid JSON.
- Every changed tool's `return_schema.oneOf` compiles under `jsonschema.Draft7Validator` and stays disjoint (checked programmatically with a synthetic error payload against every branch).
- `tu list` loads the full registry cleanly (2599 tools, no load errors) with this batch applied.
- Live-tested a representative sample from this batch with `tu test <ToolName>` against real APIs, spanning both the common "wrap `data` under `oneOf`" shape and the trickier "bare open-object success branch" shape that needed the disambiguating `required`/`not` fix:
  - `AlphaFill_get_transplants`, `AMPSphere_search_amps`, `Crossref_get_member`, `CTIS_get_trial`, `Sequence_dn_ds`, `CoL_get_taxon`, `BiGG_download_model`, `CPIC_get_gene_info`, `CTD_get_chemical_gene_interactions`, `DBLP_search_authors`, `ComplexPortal_search_complexes`, `ClinicalCalc_CHA2DS2_VASc`, `ArtIC_search_artworks`, `Dfam_search_families` — all pass.
  - Three other tools tested surfaced pre-existing, unrelated failures (not caused by this change, confirmed by diffing against the current `main` schema before this fix): `run_deseq2_analysis` (stale test_examples file path), `CryoET_list_datasets` (a `relatedDatabaseEntries` field pre-typed as `"type": "null"` on `main`, unrelated to the envelope wrapping), `ConoServer_get_conopeptide` (upstream site returning 403). None are return_schema-envelope issues; flagging for a separate follow-up rather than bundling into this fix.

## Follow-up (not included here)

- `CELLxGENE_get_census_versions` / `CELLxGENE_get_presence_matrix` (in `cellxgene_census_tools.json`) share the same detection signature but have a different root cause: `CELLxGENECensusTool.run()` doesn't wrap results under a `data` key at all, so their current `return_schema` relies on a `{"type": "null"}` branch as a workaround and currently passes `tu test`. Fixing them properly needs a code change to the tool's `run()` method, not just a schema edit, so they're intentionally excluded from this batch.
- The `CryoET_list_datasets` nullable-field bug found during live testing above.

## Test coverage

Schema-only change; correctness is validated via the programmatic disjointness check plus live `tu test` runs above, matching the fix recipe from #288 (no new unit test needed — existing `tu test` machinery is exactly what this PR fixes to pass).

### DIFF
--- a/src/tooluniverse/data/aging_cohort_tools.json
+++ b/src/tooluniverse/data/aging_cohort_tools.json
@@ -11,93 +11,159 @@
           "description": "Keyword search across study names, descriptions, variable categories, and topics. Examples: 'iron intake longitudinal', 'grip strength aging Europe', 'biomarkers elderly China'."
         },
         "country": {
-          "type": ["string", "null"],
+          "type": [
+            "string",
+            "null"
+          ],
           "description": "Filter by country or region. Examples: 'USA', 'UK', 'China', 'Europe', 'Netherlands'. Case-insensitive substring match."
         },
         "design": {
-          "type": ["string", "null"],
+          "type": [
+            "string",
+            "null"
+          ],
           "description": "Filter by study design. One of: 'longitudinal', 'cross-sectional', 'both'.",
-          "enum": ["longitudinal", "cross-sectional", "both", null]
+          "enum": [
+            "longitudinal",
+            "cross-sectional",
+            "both",
+            null
+          ]
         },
         "min_sample_size": {
-          "type": ["integer", "null"],
+          "type": [
+            "integer",
+            "null"
+          ],
           "description": "Minimum sample size threshold. Only returns cohorts with sample_size >= this value."
         },
         "has_variable": {
-          "type": ["string", "null"],
+          "type": [
+            "string",
+            "null"
+          ],
           "description": "Filter for cohorts that include a specific variable category. Substring match against variable categories. Examples: 'iron', 'grip_strength', 'walking_speed', 'nutrition', 'cognitive', 'biomarkers', 'genetics'."
         }
       },
-      "required": ["query"],
+      "required": [
+        "query"
+      ],
       "additionalProperties": false
     },
     "return_schema": {
       "oneOf": [
         {
-          "type": "object",
-          "properties": {
-            "status": {"type": "string", "enum": ["success"]},
-            "data": {
-              "type": "array",
-              "items": {
-                "type": "object",
-                "properties": {
-                  "name": {"type": "string"},
-                  "full_name": {"type": "string"},
-                  "country": {"type": "string"},
-                  "design": {"type": "string"},
-                  "sample_size": {"type": "integer"},
-                  "sample_size_note": {"type": "string"},
-                  "age_range": {"type": "string"},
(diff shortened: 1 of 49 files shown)


gasvn:

PR #292, opened 2026-07-17
Fix #288: correct envelope-style return_schema (batch 3/5, files G-N)

### LINKED ISSUE #288: return_schema envelope-vs-data-only mismatch affects ~554 more tools beyond #246/#287

gasvn:
## Summary

Part 3 of 5 batches fixing #288 — 113 of the 552 tools with envelope-style `return_schema` beyond the 9 already fixed in #287. Same root cause and fix recipe as batch 1 (#290); see that PR for the full write-up.

## Files in this batch (alphabetically G-N)

`gtopdb`, `harmonizome`, `hemolytik2`, `hgnc`, `hlaligandatlas`, `hpa`, `hubmap_sample`, `hubmap`, `huggingface_inference`, `icite`, `idigbio`, `idr_searchengine`, `idr`, `iedb_prediction`, `imgt`, `impc`, `intact`, `interpro_domain_arch`, `interpro_entry`, `interpro_ext`, `interpro_member_db`, `interpro`, `interproscan`, `ipd_imgt_hla`, `iptmnet`, `isrctn`, `iupred3`, `kegg`, `lipidmaps_gene`, `lipidmaps`, `marine_regions`, `marrvel`, `massbank`, `massive`, `mavedb`, `mendelian_randomization`, `metabolights`, `metabolite`, `metabolomics_workbench`, `mgnify_expanded`, `mgnify`, `mhcmotifatlas`, `mibig`, `mpd`, `msigdb`, `nasa_sbdb`, `ncbi_datasets`.

## A real bug this batch caught (beyond the mechanical fix)

`Harmonizome_get_gene_set_members` has two mutually exclusive success shapes depending on `mode`: `gene_set` mode returns `members`, `gene` mode returns `associations`, never both. The mechanical fix initially picked a single disambiguating `required: ["members"]`, which broke `gene` mode. Live-testing this tool's full example set (rather than just checking schema validity statically) caught the regression before push; fixed by changing to `anyOf: [{"required":["members"]}, {"required":["associations"]}]`, confirmed live against both modes.

That finding prompted a targeted sweep across all 552 tools in #288 for the same risk pattern (a test_examples parameter that varies a mode/operation/type-like key across examples, combined with a single-field `required` on the promoted success branch). It surfaced 4 more candidates; all 4 were live-tested and confirmed already correct (`BiGG_download_model`, `OpenAIRE_search_publications`, `web_search`, `EBI_msa_align`, `IDR_list_metadata_keys` — the last two required checking `EBI_msa_align` from the already-pushed batch 2 PR (#291) as well, which also passed).

## Validation

- Every changed tool's `return_schema.oneOf` compiles under `jsonschema.Draft7Validator` and stays disjoint (synthetic error-payload check).
- Live-tested a representative sample against real APIs: `GtoPdb_get_interactions`, `Harmonizome_get_gene_set_members` (both modes), `HGNC_fetch_gene_family_members`, `HuBMAP_search_samples`, `iCite_get_publications`, `iDigBio_search_records`, `IMPC_get_gene_summary`, `InterPro_search_entries`, `ISRCTN_search_trials`, `kegg_search_pathway`, `LipidMaps_get_compound_by_xref`, `MarineRegions_get_record`, `MARRVEL_get_gene`, `MassBank_get_record_by_accession`, `HMDB_search`, `MetabolomicsWorkbench_get_study`, `MGnify_get_samples`, `MSigDB_get_gene_set_members`, `NASASBDB_get_body`, `NCBIDatasets_get_genome_assembly`, `IDR_list_metadata_keys` — all pass.

## Test coverage

### DIFF
--- a/src/tooluniverse/data/gtopdb_tools.json
+++ b/src/tooluniverse/data/gtopdb_tools.json
@@ -237,420 +237,351 @@
       }
     ],
     "return_schema": {
-            "oneOf": [
-                {
-                    "type": "object",
-                    "properties": {
-                        "status": {
-                            "type": "string",
-                            "enum": [
-                                "success"
-                            ]
-                        },
-                        "data": {
-                            "type": "array",
-                            "items": {
-                                "type": "object"
-                            }
-                        }
-                    },
-                    "required": [
-                        "status",
-                        "data"
-                    ]
-                },
-                {
-                    "type": "object",
-                    "properties": {
-                        "status": {
-                            "type": "string",
-                            "enum": [
-                                "error"
-                            ]
-                        },
-                        "error": {
-                            "type": "string"
-                        }
-                    },
-                    "required": [
-                        "status",
-                        "error"
-                    ]
-                }
-            ]
-        }
-  },
- {
-  "name": "GtoPdb_search_diseases",
-  "description": "Search the Guide to Pharmacology (GtoPdb / IUPHAR-BPS) for diseases by name, and retrieve their curated disease IDs, names, synonyms, and cross-references to external databases (OMIM, Orphanet, Disease Ontology). GtoPdb links diseases to the drug targets and approved/experimental ligands relevant to them. Example: 'epilepsy' returns disease entries such as 'Epilepsy and Intellectual Disability'. No authentication required.",
-  "parameter": {
-   "type": "object",
-   "properties": {
-    "name": {
-     "type": [
-      "string",
-      "null"
-     ],
-     "description": "Disease name to search. Examples: 'epilepsy', 'asthma', 'hypertension', 'breast cancer'."
-    },
-    "query": {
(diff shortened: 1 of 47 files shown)


gasvn:

PR #293, opened 2026-07-17
Fix #288: correct envelope-style return_schema (batch 4/5, files N-R)

### LINKED ISSUE #288: return_schema envelope-vs-data-only mismatch affects ~554 more tools beyond #246/#287

gasvn:
## Summary

Part 4 of 5 batches fixing #288 — 111 of the 552 tools with envelope-style `return_schema` beyond the 9 already fixed in #287. Same root cause and fix recipe as batch 1 (#290); see that PR for the full write-up.

## Files in this batch (alphabetically N-R)

`ncbi_sra`, `ncbi_variation`, `nci_cactus`, `network_proximity`, `neuromorpho`, `norine`, `npatlas`, `obis`, `odphp`, `oma`, `oncokb`, `oncotree`, `open_genes`, `openaire_dataset`, `openaire`, `openfda_label`, `openfoodfacts`, `opengwas`, `opentarget_genetics`, `opentarget`, `opentopodata`, `opsin`, `orcid`, `orthodb`, `package_discovery`, `panglaodb`, `pathoplexus`, `pathwaycommons`, `pdb_redo`, `pdbe_sifts`, `pdc`, `pepcalc`, `peplife2`, `peptideatlas`, `pgs_catalog`, `pharmgkb`, `pharos`, `pheweb_phewas`, `phykit`, `pride`, `progenetix`, `proteomexchange`, `proteomicsdb_meltome`, `proteomicsdb`, `pubchem`, `pubtator3_ext`, `replicate`, `retraction`, `rgd_strain`, `rhea_reaction`, `rnacentral_genome`.

## Two real bugs this batch caught (beyond the mechanical fix)

Live-testing the full example set of the sampled tools (rather than just checking schema validity statically) caught two pre-existing, unrelated bugs in files this batch already touches:

1. **`RGD_search_strains` / `RGD_get_strain_annotations` (`rgd_strain_tools.json`) and `PanglaoDB_markers_for_cell_type` / `PanglaoDB_cell_types_for_gene` / `PanglaoDB_list_cell_types` (`panglaodb_tools.json`)** — 5 tools whose `return_schema` was shaped `{"type": "object", "oneOf": [...]}`, with a stray top-level `type` sibling to `oneOf`. Since these tools' real success payload is an array, the redundant root-level `"type": "object"` made the schema unsatisfiable no matter how the `oneOf` branches were fixed (jsonschema requires the instance to match *both* the root type and one of the `oneOf` branches). This predates #288 and is a different bug class (not the "envelope declares status" pattern), but it lives in files this batch already modifies, so it's fixed here as a separate, clearly-labeled commit — removing the redundant `type` key, since `oneOf` alone already fully constrains the schema. Confirmed live against all 5 tools (they were failing 100% before this fix, now pass).
2. A sweep for the same "root `type` conflicts with an array-typed `oneOf` branch" pattern was run across *all* `src/tooluniverse/data/*_tools.json` (not just the 552 in scope for #288) — these 5 were the only hits anywhere in the codebase.

## Other pre-existing issues found (not fixed here, out of scope)

- `OpenFoodFacts_get_product`: one field 4 levels deep (`product.ecoscore_data.adjustments.packaging.packagings[].food_contact`) is typed `["integer","number"]` but the live API sometimes returns a numeric string. Pre-existing on `main`, unrelated to the envelope pattern, and touching this ~500-property auto-generated-looking schema is out of scope here.

## Validation

- Every changed tool's `return_schema.oneOf` compiles under `jsonschema.Draft7Validator` and stays disjoint (synthetic error-payload check).
- Live-tested a representative sample against real APIs, including the two multi-mode disambiguation cases (`Replicate_run_prediction`/`Replicate_get_prediction` skipped live — they require an API key not configured in this environment, but were validated statically): `NCBIVariation_spdi_to_rsids`, `NeuroMorpho_get_literature`, `Norine_get_peptide`, `NPAtlas_get_compound`, `OMA_get_protein`, `OncoKB_get_gene_info`, `OncoTree_search`, `OpenGenes_get_gene`, `FDA_search_drug_labels`, `OpenTopoData_get_elevation`, `OPSIN_name_to_structure`, `PathwayCommons_paths_between`, `PGSCatalog_get_score`, `PharmGKB_get_drug_label_annotations`, `Pharos_get_target_ligands`, `PubChem_get_substance_by_SID`, `PubTator3_GetEntityRelations`, `Rhea_get_reaction`, `RNAcentral_get_sequence`, `Network_proximity`, `RGD_search_strains`, `RGD_get_strain_annotations`, `PanglaoDB_markers_for_cell_type`, `PanglaoDB_cell_types_for_gene`, `PanglaoDB_list_cell_types` — all pass.

## Test coverage


gasvn:

PR #294, opened 2026-07-17
Fix #288: correct envelope-style return_schema (batch 5/5, files R-Z)

### LINKED ISSUE #288: return_schema envelope-vs-data-only mismatch affects ~554 more tools beyond #246/#287

gasvn:
## Summary

Part 5 of 5 batches fixing #288 — 100 of the 552 tools with envelope-style `return_schema` beyond the 9 already fixed in #287 (one tool originally in this batch, `WAQI_get_air_quality`, turned out to be a detection false positive and was left unchanged — see below). Same root cause and fix recipe as batch 1 (#290); see that PR for the full write-up.

## Files in this batch (alphabetically R-Z)

`rnacentral`, `roc_analysis`, `ror`, `rxclass`, `rxn_chemistry`, `rxnorm_extended`, `sabdab`, `sasbdb`, `scanprosite`, `screen`, `scxa`, `semantic_scholar_ext`, `sgd_protein`, `signor`, `spredixcan`, `ssgsea`, `stitch`, `structure_annotation`, `sunrise_sunset`, `swiss_target`, `swisslipids`, `tark`, `tdc_dataset`, `tdc_oracle`, `therasabdab`, `togoid`, `tumorhope2`, `unibind`, `uniprot_idmapping`, `uniprot_locations`, `uniprot_proteomes`, `uniprot_ref`, `uniprot_taxonomy`, `uniprot`, `usda_plants`, `variant_fraction`, `vcf_stats`, `veupathdb`, `waqi`, `web_search`, `wikipathways`, `worms`, `zfin`, `zinc`, `zooma`.

## Two more real bugs this batch caught (beyond the mechanical fix)

1. **`UniBind_search_datasets` / `UniBind_list_tfs` (`unibind_tools.json`) and `ZOOMA_annotate_text` / `ZOOMA_list_datasources` (`zooma_tools.json`)** — the same stray root-level `type` sibling to `oneOf` found in batch 4's PanglaoDB/RGD fix (their success payload is an array, and the redundant root `"type": "object"` made the schema unsatisfiable). A full codebase sweep for this pattern (across all `src/tooluniverse/data/*_tools.json`, not just the 552 tools in #288) found 11 total instances: 5 already fixed in batch 4 (#293), 4 fixed here, and 2 in `LipidMaps_get_gene`/`LipidMaps_get_protein` (`lipidmaps_gene_tools.json`, batch 3) fixed as a follow-up commit already pushed to #292. All 11 confirmed live.
2. **`WAQI_get_air_quality`** was a genuine detection false positive. Its real `result["data"]` is shaped `{"status": "ok"/"error", "data": {...}}` because the tool passes through the raw upstream WAQI API response verbatim, which happens to use `status`/`data` as its own top-level keys — coincidentally matching the buggy envelope pattern's signature. The original schema already correctly described this shape (notably: no `required` was ever set on that branch, unlike every genuine envelope-bug tool). The mechanical fix incorrectly promoted the wrong nesting level, breaking a tool that was already passing `tu test`. Reverted this one tool's `return_schema` to its original form; confirmed live it still passes. A full sweep for this exact signature (branch with no `required` and properties exactly `{status, data}`, nothing else) across all 554 originally-flagged tools found this to be the only instance.

## Validation

- Every changed tool's `return_schema.oneOf` compiles under `jsonschema.Draft7Validator` and stays disjoint (synthetic error-payload check) — except `WAQI_get_air_quality`, intentionally left in its original (pre-#288) form; see above.
- Live-tested a broad sample against real APIs: `RNAcentral_get_xrefs_and_pubs`, `ROC_analysis`, `ROR_match_affiliation`, `ROR_get_organization`, `RxClass_get_drug_classes`, `RxNorm_get_ndc_properties`, `SemanticScholar_search_authors`, `SGD_get_protein_domains`, `SIGNOR_connect_proteins`, `STITCH_get_chemical_protein_interactions`, `SunriseSunset_get_times`, `SwissLipids_search`, `Tark_get_transcript`, `TogoID_convert`, `UniBind_search_datasets`, `UniBind_list_tfs`, `UniProtIDMap_convert_ids`, `UniProtLocations_get_location`, `UniProtRef_search_diseases`, `UniProtTaxonomy_get_taxon`, `UniProt_search`, `USDA_plants_get_profile`, `VEuPathDB_search_genes_by_organism`, `WAQI_get_air_quality`, `WikiPathways_search`, `WoRMS_get_classification`, `ZFIN_get_gene_expression`, `ZOOMA_annotate_text`, `ZOOMA_list_datasources` — all pass.
- A few sampled tools hit unrelated, pre-existing issues (not caused by this change): `SAbDab_get_structure_summary` and `VCF_summary_stats`/`VCF_count_variants`/`VCF_normalize` fail on stale test_examples (a missing local file path / a structure ID SAbDab no longer serves as tabular data), `ZINC_get_compound` and `ScanProsite_find_proteins_with_motif` hit slow/timing-out upstream APIs. None are return_schema-envelope issues.

## Test coverage

### DIFF
--- a/src/tooluniverse/data/rnacentral_tools.json
+++ b/src/tooluniverse/data/rnacentral_tools.json
@@ -15,10 +15,12 @@
           "default": 10,
           "minimum": 1,
           "maximum": 100,
-          "description": "Number of records per page (1–100)."
+          "description": "Number of records per page (1\u2013100)."
         }
       },
-      "required": ["query"]
+      "required": [
+        "query"
+      ]
     },
     "fields": {
       "endpoint": "https://rnacentral.org/api/v1/rna/",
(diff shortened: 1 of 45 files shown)


gasvn:

PR #295, opened 2026-07-18
Fix Round: test-harness sweep across ChEMBL, ClinicalTrials, OpenTargets, ClinVar-annotation, Proteins API

gasvn:
## Summary

Internal test-driven improvement round (not tied to a specific GitHub issue). Ran the repo's regression suite + 3 role-play researcher-persona agents (pharmacologist/rivaroxaban, cancer genomicist/BRAF, clinical microbiologist/E. coli) exercising ~50 tool calls across ~35 distinct tools, then CLI-verified every reported issue before touching code (per project convention, roughly half of persona-reported issues turn out to be false positives — this round was no exception).

### Verified and fixed (all reproduced via `python3 -m tooluniverse.cli run` before and after)

- **`OpenTargets_get_target_expression_by_ensemblID`** — completely broken (HTTP 400, even on its own documented example) since OpenTargets renamed the GraphQL `expressions` field to `baselineExpression` with a different row-based shape. Rebuilt the query and `return_schema` against the live schema (confirmed via GraphQL introspection).
- **`ChEMBL_search_targets`** — `target_chembl_id` was silently dropped whenever combined with `fields` (a prior fix, Feature-120B-001, over-broadly excluded this tool alongside `ChEMBL_get_target` from the id→`__exact` filter mapping, even though only `ChEMBL_get_target` consumes the id via a URL path template). Was returning the entire ~18,552-row target table instead of the one requested target.
- **`ChEMBL_search_mechanisms`** — parameter schema advertised `target_chembl_id` as a working filter ("Filter by target ChEMBL ID") when the endpoint has always rejected it by design (`/mechanism.json` ignores target-based filters). Updated the description to state this and point to the actual workaround, matching the tool's own runtime behavior.
- **`annotate_variant_multi_source`** — when a queried variant (e.g. BRAF V600E) had no exact ClinVar match, `_parse_clinvar`'s fallback-context list (unrelated gene-level variants, intentionally returned for context) was being blindly summarized as `clinvar_classification`, silently attributing an unrelated variant's classification to the query. Now returns `null` + an explicit `clinvar_note` instead.
- **`search_clinical_trials`** — the tool explicitly documented as "the PRIMARY tool for finding clinical trials" has silently required `AREA[HasResults]true` by default since the very first release (an untested placeholder marked "NOTE: Can change this one" in the original commit), hiding real trials that haven't posted results yet. Reproduced with rivaroxaban+renal-impairment (11 real trials exist, tool returned 0). Removed the undocumented HasResults restriction; kept the documented phase≥2 filter. Also added `additionalProperties: false` so passing an unrecognized param (e.g. `query_intr`, valid on a sibling tool but not this one) now errors clearly instead of silently running an unfiltered query and returning `"status": "success"` with unrelated results.
- **`proteins_api_search`** — UniProt accessions (the tool's own documented example, `P05067`) always misrouted to a gene-name search (`gene=P05067`) and returned 0 results, because a prior fix (Feature-81B-007) made *all* short alphanumeric queries route to `gene=` to fix gene-symbol misclassification, with no accession detection at all. Added canonical UniProt accession-format matching before the gene-name heuristic.

### Regression suite hygiene

`pytest` was fully blocked at collection time by `tests/tools/test_hmdb_tool.py`, which imports `tooluniverse.hmdb_tool` — a module moved to `data/broken_apis/` in an earlier round (HMDB is Cloudflare-blocked) and replaced with `MetaboliteTool`. Deleted the stale test, added `tests/tools/test_metabolite_tool.py` covering the current tool + its `HMDB_*` aliases.

Once collection was unblocked, found and fixed 9 more stale tests left over from an earlier return-schema envelope migration (tools now correctly nest all data under `data`, but these tests still asserted duplicate top-level keys or a bare-list return) — all false test failures, not product bugs:
- `test_dailymed_parser.py` (5 tests)
- `test_fda_orange_book_tool.py` (8 tests, same pattern)
- `test_bindingdb_tool.py` (1 test — mock never set `status_code`, so the tool's correct HTTP-status check always treated it as an error)
- `test_europe_pmc_auto_snippets.py` (6 tests) + `test_europe_pmc_enrich_missing_abstract.py` (2 tests, one also had a missing `isOpenAccess: "Y"` in its mock payload that skipped the very code path it was meant to test)
- `test_esm_tool.py` (2 tests — one asserted every ESM tool's test_examples include an `operation` field, but none of the 10 ESM tools use operation-based dispatch; the other needed a config with no fixed `fields.operation` to actually reach the "unknown operation" branch, since real ESM_* configs always pin one)
- `test_dna_digest_enzyme_coverage.py` (2 tests — missing a `biopython`-installed skip guard, following the existing pattern used elsewhere in the suite)

### Known false positives (CLI-verified, not bugs)

- `HPA_get_protein_interactions_by_gene` — reported as "dead," but it already returns a clear, actionable error for a documented upstream HPA API deprecation, pointing to two working alternatives. No fix needed.

### Not fixed (real, verified, out of scope for this round)

- Deeper `ClinVar_search_variants` limitations (no relevance ranking / no protein-change search parameter, no oncogenicity-classification filter) — a real gap for cancer-variant lookups but requires reworking the underlying search strategy, not a quick fix.
- `NCBIDatasets_suggest_taxonomy` not surfacing exact species matches, `BVBRC_search_amr` missing an organism filter, `get_taxonomy_by_pdb_id` not extracting a clean scientific name, a few other LOW-severity ranking/UX gaps — all noted by personas, all real, all deferred as smaller-value / higher-effort items for a future round.
- The full pytest suite surfaced ~30+ *additional*, pre-existing failures unrelated to this round's tools (`test_metacyc_tool.py`, `test_researcher_simulation_fixes.py`, `test_literature_tools_error_shape.py`, `test_paper_search_tools.py`, `test_ppi_tools.py`, `test_pubmed_search_include_abstract.py`) — confirmed via targeted reruns that none of these are caused by this PR's changes. This is a separate, larger body of test debt worth a dedicated round.

### Verification

- All 18 changed/added test files pass together (98 passed, 2 skipped for missing `biopython`).
- Every fix above was reproduced live via `python3 -m tooluniverse.cli run <ToolName> '<args>'` before the change and re-verified after.
- JSON configs validated with `python3 -c "import json; json.load(...)"`.

Note: running the local test suite reliably regenerates ~450 files under `src/tooluniverse/tools/*.py` (an existing stub-sync side effect of `ToolUniverse()`/`load_tools()`, unrelated to this change) — that drift was reverted before every commit in this branch and is not part of this diff.


gasvn:

PR #296, opened 2026-07-18
Fix Round 2: test-harness sweep across structural biology, pharmacovigilance, variant-annotation

gasvn:
## Summary

Second internal test-harness round (see #295 for round 1), covering domains not exercised previously: structural biology/PDB, genomics/variant-calling, epidemiology, drug safety/pharmacovigilance, and imaging. Ran 5 role-play researcher-persona agents against ~85 tool calls across ~40 distinct tools, then CLI-verified every reported issue before touching code.

### A note on a session-local environment issue

Several persona reports were dominated by a stale MCP server process in this session, holding a reference to a `uv` cache directory (`~/.cache/uv/archive-v0/...`) that had since been garbage-collected. This caused two consistent symptoms across almost every network-calling tool call made *through the MCP interface*: a "Could not find a suitable TLS CA certificate bundle" error, and/or "Tool type 'X' not found in registry" for more recently-added tool families. This is an environment/process-lifecycle issue specific to this session (the MCP server needs a restart against a fresh install), **not a ToolUniverse code defect** — confirmed by re-running the exact same tool calls via the local `tu`/`python3 -m tooluniverse.cli` (a fresh process each time), which worked correctly. All findings below were verified via that CLI path, not the broken MCP session.

### Verified and fixed

- **`Foldseek_search_structure`** — the code's own default search `mode` (`"tmalign"`) didn't match the JSON schema's documented default (`"3diaa"`), so any call that omitted `mode` (the common case) silently ran the wrong Foldseek search mode and returned noise-level hits: e-value ≈0.3 (not significant), 2–5% sequence identity, structurally unrelated proteins. Reproduced on both the persona's real query (SARS-CoV-2 Mpro, PDB 6LU7) and a trivial control case (hemoglobin, PDB 4HHB) — the same garbage-result pattern on both confirmed this was systemic, not query-specific. After the fix, 4HHB correctly returns other hemoglobin structures at e-value 4.985e-18 / 100% identity, and 6LU7 correctly returns other SARS-CoV-2 Mpro structures at e-value ~1e-60 / ~99% identity. Also fixed a copy-paste typo in the `mode` parameter description that named `'tmalign'` twice instead of `'3diaa'` and `'tmalign'`, and clarified that the `sequence`/`query` parameters (documented as an alternative to `pdb_id`) always fail server-side, since Foldseek's public ticket API requires an actual 3D structure.
- **`PDBe_get_structure_ligands` / `PDBe_get_bound_molecules`** (`ligand_monomers` and `residue_listing` PDBe endpoints) — these endpoints return HTTP 404 whenever an entry simply has no data there (e.g. its bound inhibitor is modeled as a polymer/peptide chain rather than a discrete non-polymer ligand), which the tool unconditionally reported as `"PDB entry '{id}' not found"` — including for real, extremely well-known entries. Reproduced on SARS-CoV-2 Mpro (6LU7, N3 inhibitor) and 3 other Mpro structures with peptidomimetic inhibitors (7BQY, 6M03, 6Y2E). Added a lightweight follow-up check against PDBe's summary endpoint to disambiguate "entry doesn't exist" from "entry exists but has no data at this specific endpoint"; the latter now returns a clean empty success with an explanatory note instead of a false error.
- **`FAERS_search_reports_by_drug_and_reaction`** and 15 sibling `FAERS_search_*`/`FAERS_count_*` tools — `medicinalproduct` only matched openFDA's raw, literal `patient.drug.medicinalproduct` free-text field, not the FDA-normalized `patient.drug.openfda.generic_name` field. Reproduced with semaglutide: searching the generic name returned zero gastroparesis reports, while searching a brand name (Ozempic) found the same real report — since ~92% of FAERS reports for this drug are filed under brand names (Ozempic/Wegovy/Rybelsus), not the generic name. Fixed by adding the normalized field as an OR-matched alternative, reusing the multi-field search-mapping mechanism the same tool classes already support and use elsewhere (no new code paths). This also resolves a related, previously-noted-but-unfixed inconsistency from round 1 (rivaroxaban generic-name vs. brand-name FAERS death counts no longer diverge for the wrong reason).
- **`annotate_variant_multi_source`** — the four sub-source calls (ClinVar, gnomAD, CIViC, UniProt) signal failure by returning a `{"status": "error", ...}` dict, which is the normal way tools in this framework report failure — but the aggregator's `try/except` only catches *raised* exceptions, so a failed sub-call was silently parsed as if it were valid (empty) data and never recorded in `sources_failed`. Reproduced independently of the session's environment issue via a clean mock: with every sub-call failing, the tool returned `"status": "success"` with an empty result set for BRCA1 — a gene with thousands of real ClinVar entries — giving no indication anything had gone wrong. Now correctly reports each failed source in `sources_failed` and leaves `annotations` empty for it, rather than fabricating an empty-but-successful result.

### Regression suite hygiene

While working in the FAERS tool family for the `medicinalproduct` fix, found `tests/tools/test_faers_analytics_tool.py` (6 tests) failing due to the same class of stale test debt already cleaned up in round 1 — assertions written against a flat/duplicate-top-level response shape from before an earlier return-schema envelope migration, when the tools now correctly nest everything under `data`. Fixed the assertions to match current (correct) tool behavior.

### Known false positives / environment noise (not fixed, not real code bugs)

- The MCP-server TLS/registry staleness described above — confirmed environment-specific via CLI cross-check, not reproducible in a fresh process.
- `HPA_get_protein_interactions_by_gene`-style "already handled correctly" cases were not re-flagged this round.

### Verified-real but out of scope this round (lower value / bigger design changes, deferred)

- PDBe/RCSB free-text search tools (`RCSBAdvSearch_search_structures`, `PDBeSearch_search_structures`) rank results by resolution/deposition date rather than textual relevance, so a query like "nirmatrelvir" without an organism filter surfaces unrelated 2010-era HIV protease structures above the real SARS-CoV-2 hits. Same class of issue as ClinVar's ranking gap noted in round 1 — needs a relevance-sort option, not a quick fix.
- `RCSBData_get_nonpolymer_entity` reports "entry not found" for an entity that exists but is polymer-typed (same root cause family as the PDBe fix above, but in a different tool with a different response shape).
- `NeuroVault_list_collections`/`list_atlases` have no keyword/tag search parameter (17,586 total collections, no practical way to filter) — a coverage gap, not a bug.
- TCIA's `TCGA-GBM` example in tool descriptions silently returns empty results — confirmed via direct upstream API check to be a real NBIA v1 API limitation (the collection isn't served there anymore), not a ToolUniverse bug; the example should be swapped for a currently-live collection (e.g. `UPENN-GBM`) in a docs-focused pass.
- `TCIA_get_body_part_values` returns one spurious empty `{}` row alongside real data — cosmetic, low severity.
- `get_cellpose_info` reports PyPI's latest published version rather than checking whether the package is actually installed locally, which can mislead a user into thinking segmentation is ready to run.

### Verification

- All 4 fixed/touched source files have new or updated unit tests (10 files total changed/added), all passing: `test_compound_variant_tool.py` (2 new + 5 existing), `test_foldseek_tool.py` (new, 2 tests), `test_pdbe_ligands_tool.py` (new, 3 tests), `test_faers_analytics_tool.py` (11, all fixed).
- Every fix was reproduced live via `python3 -m tooluniverse.cli run <ToolName> '<args>'` before the change and re-verified after; the Foldseek and FAERS fixes were also cross-checked against their originally-broken real-world queries (6LU7, 4HHB, semaglutide/gastroparesis).
- JSON configs validated with `python3 -c "import json; json.load(...)"`.

Branch created fresh off `origin/main` (not building on the still-open #295) via an isolated worktree, per project convention.

### DIFF
--- a/src/tooluniverse/compound_variant_tool.py
+++ b/src/tooluniverse/compound_variant_tool.py
@@ -6,7 +6,7 @@ pathogenicity classification, population frequencies, and clinical evidence.
 """
 
 import re
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 from .base_tool import BaseTool
 from .tool_registry import register_tool
@@ -104,7 +104,11 @@ class CompoundVariantAnnotationTool(BaseTool):
                         "arguments": {"query": clinvar_query, "limit": 20},
                     }
                 )
-                annotations["clinvar"] = self._parse_clinvar(r, variant_token)
+                error = self._sub_call_error(r)
+                if error:
+                    sources_failed.append(f"ClinVar: {error[:100]}")
+                else:
+                    annotations["clinvar"] = self._parse_clinvar(r, variant_token)
             except Exception as e:
                 sources_failed.append(f"ClinVar: {str(e)[:100]}")
 
@@ -117,7 +121,11 @@ class CompoundVariantAnnotationTool(BaseTool):
                         "arguments": {"gene_symbol": gene_for_gnomad},
                     }
                 )
-                annotations["gnomad"] = self._parse_gnomad(r)
+                error = self._sub_call_error(r)
+                if error:
+                    sources_failed.append(f"gnomAD: {error[:100]}")
+                else:
+                    annotations["gnomad"] = self._parse_gnomad(r)
             except Exception as e:
                 sources_failed.append(f"gnomAD: {str(e)[:100]}")
 
@@ -130,7 +138,11 @@ class CompoundVariantAnnotationTool(BaseTool):
                         "arguments": {"gene_symbol": gene_for_gnomad},
                     }
                 )
-                annotations["civic"] = self._parse_civic(r, variant_token)
+                error = self._sub_call_error(r)
+                if error:
+                    sources_failed.append(f"CIViC: {error[:100]}")
+                else:
+                    annotations["civic"] = self._parse_civic(r, variant_token)
             except Exception as e:
                 sources_failed.append(f"CIViC: {str(e)[:100]}")
 
@@ -147,7 +159,11 @@ class CompoundVariantAnnotationTool(BaseTool):
                         },
                     }
                 )
-                annotations["uniprot"] = self._parse_uniprot(r)
+                error = self._sub_call_error(r)
+                if error:
+                    sources_failed.append(f"UniProt: {error[:100]}")
+                else:
+                    annotations["uniprot"] = self._parse_uniprot(r)
             except Exception as e:
                 sources_failed.append(f"UniProt: {str(e)[:100]}")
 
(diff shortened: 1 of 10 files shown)


gasvn:

PR #298, opened 2026-07-18
Test-harness round 4: cBioPortal, GDC, USDA PLANTS, DailyMed domain-sweep fixes

gasvn:
## Summary

Round 4 of the ongoing test-harness sweep. Persona domains this round: oncology, microbiome, cardiology, ophthalmology, agricultural science. Every issue below was CLI-verified (`python3 -m tooluniverse.cli run <Tool> '<args>'`) before fixing — roughly half of raw persona reports turned out to be false positives, consistent with prior rounds.

## Verified bugs fixed

- **Fix-R4A-001** (`cbioportal_tool.py`) — `_build_url` only substitutes `{placeholder}` tokens for keys present in the call `arguments`, so an omitted optional param (e.g. `limit`, schema default=20) left its literal `{limit}` unfilled in the endpoint template, sending a broken query string and causing a live 400 on `cBioPortal_get_cancer_studies`. Now fills in schema-declared defaults for any omitted param before building the URL.

- **Fix-R4A-002** (`gdc_tool.py`) — `GDCMutationFreqByProjectTool` read `case_summary.doc_count` from the GDC analysis endpoint as a project's total case count, but that field is not reliable (confirmed against the raw API and cross-checked with `GDC_search_cases`/`GDC_list_projects` — it reported 2282 for TCGA-COAD when the real total is 461, understating KRAS mutation frequency in colorectal cancer as 11% instead of the correct ~54%). Now fetches true per-project totals from the `/projects` endpoint's `summary.case_count` field, the same field `GDC_list_projects` already uses, falling back to the old value only if that lookup fails.

- **Fix-R4C-1** (`dailymed_tool.py`) — SPL XML documents expose the same section content through multiple matching `<section>` elements (a Highlights summary plus the Full Prescribing Information), so `DailyMed_parse_drug_interactions` and `DailyMed_parse_dosing` returned exact duplicates (20 items/10 unique, and 103/79 respectively, for apixaban). Added a shared `_dedupe_items()` helper applied consistently across all 5 SPL parser operations (safe no-op where there's no duplication).

- **Fix-R4C-2** (`dailymed_tool.py`) — when `drug_name` was provided but the automatic setid lookup found no match (e.g. a misspelled drug name), the tool returned `"Missing required parameter: setid"` — telling the caller to do exactly what they'd already done. Now returns a distinct `"No DailyMed SPL found for drug_name=X"` message naming the actual problem, while the truly-missing-both-params case keeps its original message.

- **Fix-R4E-4** (`usda_plants_tool.py`) — `USDA_plants_search_by_name`'s `common_name` fallback (used when `CommonName` is empty) left raw HTML markup un-stripped, e.g. `"<i>Holcus sorghum</i> L."` — the sibling `scientific_name` field already stripped this via `_strip_html`, but the fallback didn't.

- **Fix-R4E-2** (`cli.py`) — `tu grep -i` errored with `"unrecognized arguments: -i"` since users reflexively type grep's `-i` flag. Accepted as a no-op flag (text mode, the default, is already case-insensitive).

- **R4E-1** — `tu`/pip package mismatch hit by a persona's fresh environment (stale pip install lacking `cli.py`) — local onboarding friction, not a code bug.
- **R4E-3** — `USDA_plants_search_by_name` expects `searchText` not `name`; the error message already names the correct field clearly, and renaming/aliasing risks the param-alias anti-pattern for a cosmetic naming mismatch.
- **R4E-5** — `GBIF_search_species` free-text ranking puts an unrelated virus above *Zea mays* for query "maize" — `GBIF_match_name` (scientific name) works correctly; likely upstream relevance ranking, not ToolUniverse-side.
- **R4B-1** — `MGnify_get_taxonomy` param naming (`analysis_id` vs a sibling tool's `study_accession`) — low severity, error message already actionable.

The MCP server in this development session holds a stale reference to a garbage-collected `uv` cache dir, causing TLS cert / "tool type not found" errors across `mcp__tooluniverse__*` calls. Every finding above was independently verified via the local CLI, which is unaffected. This is documented for consistency with prior round PRs (#295, #296, #297) — not something this PR changes.

## Test plan

- [x] 4 new mocked unit test files (`tests/unit/test_cbioportal_tool.py`, `test_gdc_mutation_frequency.py`, `test_usda_plants_tool.py`, `test_dailymed_dedup_and_errors.py`) — 11 new tests, all passing
- [x] 1 new CLI test (`test_grep_accepts_ignore_case_flag` in `tests/tools/test_tu_cli.py`) — passing, plus all 66 existing `grep`-related CLI tests still pass
- [x] Every fix re-verified live via `python3 -m tooluniverse.cli run` after the code change
- [x] Auto-regen stub drift (`uv.lock`, `_lazy_registry_static.py`) reverted before commit — diff contains only intentional changes

### DIFF
--- a/src/tooluniverse/cbioportal_tool.py
+++ b/src/tooluniverse/cbioportal_tool.py
@@ -151,6 +151,21 @@ class CBioPortalRESTTool(BaseTool):
 
     def run(self, arguments: Dict[str, Any]) -> Dict[str, Any]:
         try:
+            # Fix-R4A-001: _build_url only substitutes {placeholder} for keys
+            # actually present in `arguments`, so an omitted optional param
+            # (e.g. `limit`) left its literal "{limit}" placeholder unfilled
+            # in the endpoint template, sending a broken query string to the
+            # live API instead of falling back to the schema's declared
+            # default value.
+            schema_props = self.tool_config.get("parameter", {}).get("properties", {})
+            defaults = {
+                name: prop["default"]
+                for name, prop in schema_props.items()
+                if "default" in prop and name not in arguments
+            }
+            if defaults:
+                arguments = {**arguments, **defaults}
+
             if "query" in arguments and "keyword" not in arguments:
                 arguments = {**arguments, "keyword": arguments["query"]}
             if (
--- a/src/tooluniverse/cli.py
+++ b/src/tooluniverse/cli.py
@@ -1703,6 +1703,15 @@ def main() -> None:
     p.add_argument(
         "--categories", nargs="+", metavar="CAT", help="Filter by category names"
     )
+    # Fix-R4E-2: users reflexively type grep's -i flag; accept it instead of
+    # an unrecognized-arguments error. text mode (the default) already does
+    # case-insensitive matching, so this is a no-op rather than a real toggle.
+    p.add_argument(
+        "-i",
+        "--ignore-case",
+        action="store_true",
+        help="No-op: text mode (the default) is already case-insensitive",
+    )
     p.set_defaults(func=cmd_grep)
 
     # ── info ──────────────────────────────────────────────────────────────────
--- a/src/tooluniverse/dailymed_tool.py
+++ b/src/tooluniverse/dailymed_tool.py
@@ -1,5 +1,6 @@
 # dailymed_tool.py
 
+import json
 import requests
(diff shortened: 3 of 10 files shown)


gasvn:

PR #299, opened 2026-07-18
Test-harness round 5: ClinVar, CPIC, DailyMed domain-sweep fixes

gasvn:
## Summary

Round 5 of the ongoing test-harness sweep. Persona domains this round: neurology, dermatology, psychiatry, OB-GYN/reproductive health, veterinary/comparative oncology. Every issue below was CLI-verified (`python3 -m tooluniverse.cli run <Tool> '<args>'`) before fixing.

## Verified bugs fixed

- **Fix-R5D-1** (`clinvar_tool.py`) — `ClinVar_search_variants` already has an alias convention (`gene_symbol`→`gene`, `significance`→`clinical_significance`, `query`→`condition`), but any other unrecognized key (e.g. `gene_name`) was silently dropped, producing a generic "at least one search parameter is required" error that gave no hint the parameter itself was misnamed. Rather than adding yet another one-off alias, the error now names any unrecognized parameter(s) directly so a caller can fix the typo without guessing.

- **Fix-R5C-2** (`cpic_search_pairs_tool.py`) — `CPIC_get_recommendations` resolves a drug name to a guideline via CPIC's own `/drug` table, which only indexes generic names (confirmed via raw API query — no brand-name field exists). A brand name like "zoloft" failed to resolve and returned "Use CPIC_list_guidelines to find valid guideline IDs", which doesn't solve the actual problem (not knowing the generic name). The error now names the real constraint and suggests the generic name instead.

- **Fix-R5B-1** (`dailymed_tool.py`) — `_parse_dosing`, `_parse_drug_interactions`, and `_parse_clinical_pharmacology` each extracted all `<table>` elements first, then all `<paragraph>` elements, then all `<list><item>` elements, in three separate xpath passes — destroying document order. SPL sections routinely interleave a heading paragraph immediately before the table/list it introduces (e.g. "Juvenile Idiopathic Arthritis (2.3):" followed by its pediatric dosing table, confirmed via raw XML inspection), and the three-pass extraction grouped every table before every paragraph regardless of where each actually appeared in the source document. Added a shared `_extract_ordered_content()` helper that walks a `<text>` element's direct children in true document order, applied to the three affected methods. `_parse_adverse_reactions` and `_parse_contraindications` were deliberately left untouched — they have different (mutually-exclusive table-or-paragraph) branching logic that a mechanical order fix would silently alter beyond just reordering, and no ordering issue was reported/verified for them this round.

- **R5A-1** — `NCBIGene_get_summary` expects `id` not `gene_id`; error message already names the correct field clearly (naming-intuition mismatch, not a functional bug).
- **R5A-2** — `tu grep -i` unrecognized-argument error — already fixed in round 4 (Fix-R4E-2, PR #298, not yet merged when this round's worktree was branched off main).
- **R5A-3** — `OpenTargets_get_associated_drugs_by_disease_efoId` has two similarly-named clinical-stage fields (`drug.maximumClinicalStage` vs. top-level `maxClinicalStage`) with different values and no disambiguating documentation — likely a legitimate disease-specific vs. global-stage distinction passed through from OpenTargets' GraphQL schema rather than a ToolUniverse-side bug.
- **R5B-2** — `ClinicalTrials_search_studies`'s `query_intr` performs literal text matching rather than mechanism-of-action class expansion (a class-name query like "IL-17 inhibitor" underperforms a specific drug name) — tool description already models correct usage with a specific drug name example; arguably working as documented.
- **R5C-1** — `CPIC_get_gene_info` expects `symbol` not `gene_symbol`; error message already names the correct field clearly.
- **R5D-2** — `FDA_get_pregnancy_effects_info_by_drug_name` and `PubMed_Guidelines_Search` return raw passthrough envelopes without a `status` key, inconsistent with the `{status, data}` convention used by sibling tools — a systemic envelope-consistency issue better handled as a dedicated audit (see the prior issue #288 envelope batch, PRs #290-294) rather than a two-tool patch in this round.
- **R5E-1** — `ensembl_get_taxonomy`'s returned lineage excludes the queried taxon's own record — confirmed faithful to the real upstream Ensembl REST endpoint via direct curl, not a ToolUniverse-side bug (though the tool description doesn't warn about this, which is a documentation gap rather than a functional one).

## Note on file overlap with #298

This PR and round 4's PR (#298, not yet merged) both touch `dailymed_tool.py`, on different methods/fixes (round 4: dedup duplicated SPL sections; round 5: document-order preservation). Whichever merges second may need a small conflict resolution — both changes are independent and compose cleanly.

The MCP server in this development session holds a stale reference to a garbage-collected `uv` cache dir, causing TLS cert / "tool type not found" errors across `mcp__tooluniverse__*` calls. Every finding above was independently verified via the local CLI, which is unaffected. Documented for consistency with prior round PRs (#295–298) — not something this PR changes.

## Test plan

- [x] 3 new mocked unit test files (`test_dailymed_ordered_content.py`, `test_clinvar_unrecognized_param.py`, `test_cpic_brand_name_error.py`) — 8 new tests, all passing
- [x] Every fix re-verified live via `python3 -m tooluniverse.cli run` after the code change
- [x] Confirmed the 5 pre-existing failures in `tests/tools/test_dailymed_parser.py` (stale envelope-shape assumptions) pre-date this round's changes — not touched

### DIFF
--- a/src/tooluniverse/clinvar_tool.py
+++ b/src/tooluniverse/clinvar_tool.py
@@ -212,6 +212,34 @@ class ClinVarSearchVariants(ClinVarRESTTool):
             query_parts.append(f'"clinsig {clnsig}"[Filter]')
 
         if not query_parts:
+            # Fix-R5D-1: a caller-supplied param that doesn't match any
+            # recognized name/alias (e.g. "gene_name" instead of "gene" or
+            # "gene_symbol") was silently dropped, producing a generic
+            # "at least one search parameter is required" error that gave
+            # no hint the parameter itself was misnamed. Name the actual
+            # unrecognized keys so the caller can fix the typo directly.
+            recognized = {
+                "gene",
+                "gene_symbol",
+                "condition",
+                "query",
+                "variant_id",
+                "clinical_significance",
+                "significance",
+                "max_results",
+                "limit",
+            }
+            unrecognized = sorted(set(arguments) - recognized)
+            if unrecognized:
+                return {
+                    "status": "error",
+                    "error": (
+                        f"Unrecognized parameter(s): {', '.join(unrecognized)}. "
+                        "Valid search parameters: gene (or gene_symbol), "
+                        "condition (or query), variant_id, "
+                        "clinical_significance (or significance)."
+                    ),
+                }
             return {
                 "status": "error",
                 "error": "At least one search parameter is required",
--- a/src/tooluniverse/cpic_search_pairs_tool.py
+++ b/src/tooluniverse/cpic_search_pairs_tool.py
@@ -67,11 +67,19 @@ class CPICGetRecommendationsTool(BaseTool):
                 }
             result = _resolve_drug_to_guideline_id(drug)
             if result is None:
+                # Fix-R5C-2: CPIC's /drug table only has generic names (no
+                # brand-name field), so a brand name like "zoloft" never
+                # matches. The old message sent callers to browse guideline
+                # IDs, which doesn't actually solve "I don't know the
+                # generic name" -- name the real constraint instead.
                 return {
                     "status": "error",
                     "error": (
                         f"No CPIC guideline found for drug '{drug}'. "
-                        "Use CPIC_list_guidelines to find valid guideline IDs."
+                        "CPIC only indexes generic drug names (e.g. "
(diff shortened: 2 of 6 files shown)


gasvn:

PR #302, opened 2026-07-18
Test-harness consolidated: rounds 1-7 (26 verified fixes, 41 new tests)

gasvn:
## Summary

Consolidates 7 sequential test-harness rounds (#295, #296, #297, #298, #299, #300, #301) into one PR. Each round ran role-play researcher-persona agents against a fresh domain slice of the ~2000+ tool surface, then CLI-verified every reported issue (`python3 -m tooluniverse.cli run <Tool> '<args>'`) before fixing — consistently around half of raw persona reports turned out to be false positives across all 7 rounds.

**26 verified, root-caused bugs fixed** across 24 tool/CLI files, with **41 new/updated unit tests**, all passing. Full per-round breakdown with false positives and deferred findings below, organized by round for traceability back to the original PRs.

### A note on merging

Rounds 4, 5, 6, and 7 each independently touched `dailymed_tool.py` (and rounds 5/6 both touched `clinvar_tool.py`), since DailyMed's SPL parser surfaced new, different bugs almost every round. All four rounds' fixes compose correctly and are all present in this PR — verified by re-running every fix live via CLI after merging. Two conflict regions needed manual resolution (documented inline in the merge commits): round 6's `_cell_text()` helper coexisting with round 5/7's `_extract_ordered_content()` helper, and round 5's and round 7's independent (but code-identical) implementations of `_extract_ordered_content()` — the latter two were reconciled into one function with a combined docstring covering both motivating bugs.

---

## Round 1 (originally #295) — ChEMBL, ClinicalTrials, OpenTargets, ClinVar-annotation, Proteins API

Personas: pharmacologist, cancer genomicist, clinical microbiologist.

- **`OpenTargets_get_target_expression_by_ensemblID`** — completely broken (HTTP 400) since OpenTargets renamed the GraphQL `expressions` field to `baselineExpression` with a different row-based shape. Rebuilt against the live schema.
- **`ChEMBL_search_targets`** — `target_chembl_id` was silently dropped whenever combined with `fields`, returning the entire ~18,552-row target table instead of the one requested target.
- **`ChEMBL_search_mechanisms`** — schema advertised `target_chembl_id` as a working filter when the endpoint has always rejected it by design; updated the description to match real behavior.
- **`annotate_variant_multi_source`** — when a queried variant had no exact ClinVar match, the fallback gene-level context list was blindly summarized as `clinvar_classification`, silently misattributing an unrelated variant's classification.
- **`search_clinical_trials`** — silently required `AREA[HasResults]true` by default since first release (an untested placeholder), hiding real trials that haven't posted results yet. Also added `additionalProperties: false` so an unrecognized param no longer silently runs an unfiltered query.
- **`proteins_api_search`** — UniProt accessions (the tool's own documented example) always misrouted to a gene-name search and returned 0 results. Added accession-format detection before the gene-name heuristic.
- **Regression suite hygiene** — deleted a stale test file blocking pytest collection entirely (`test_hmdb_tool.py`, referencing a module moved to `broken_apis/`), replaced with `test_metabolite_tool.py`; fixed 9 more stale tests across 6 files left over from an earlier return-schema envelope migration.

False positive: `HPA_get_protein_interactions_by_gene` reported "dead" — already returns a clear, actionable error for a documented upstream deprecation.

## Round 2 (originally #296) — Structural biology/PDB, genomics/variant-calling, epidemiology, pharmacovigilance, imaging

Personas: structural biologist, clinical geneticist, epidemiologist, pharmacovigilance officer, radiologist.

- **`Foldseek_search_structure`** — the code's default `mode` (`"tmalign"`) didn't match the schema's documented default (`"3diaa"`), so any call omitting `mode` silently ran the wrong search mode and returned noise-level hits (e-value ≈0.3, 2-5% identity). Independently reproduced on two unrelated structures (SARS-CoV-2 Mpro, hemoglobin).
- **`PDBe_get_structure_ligands` / `PDBe_get_bound_molecules`** — reported "PDB entry not found" for real, well-known entries whenever the bound ligand happened to be polymer-typed rather than a discrete non-polymer ligand. Added a summary-endpoint check to disambiguate "doesn't exist" from "exists but no data at this endpoint."
- **`FAERS_search_reports_by_drug_and_reaction`** + 15 sibling tools — `medicinalproduct` only matched openFDA's raw free-text field, not the normalized generic-name field, so searching a drug's generic name (e.g. semaglutide) missed ~92% of real reports filed under brand names.
- **`annotate_variant_multi_source`** — sub-source calls (ClinVar, gnomAD, CIViC, UniProt) signal failure via a `{"status": "error"}` dict, not by raising, so a failed call was silently parsed as valid-but-empty data instead of being recorded in `sources_failed`.
- **Regression suite hygiene** — fixed 6 more stale tests in `test_faers_analytics_tool.py` (same envelope-migration debt pattern as round 1).

Deferred (real, lower-value/bigger-design-change): PDBe/RCSB free-text search relevance ranking, `RCSBData_get_nonpolymer_entity` polymer-typed 404s, NeuroVault missing keyword search, a stale TCIA example collection, a cosmetic empty-row bug, `get_cellpose_info` checking PyPI instead of local install.

## Round 3 (originally #297) — Toxicology, nutrition/food science, clinical genetics, environmental health, medical devices

Personas: toxicologist, nutrition scientist, clinical geneticist (substituted for an immunology persona that hit a policy-classifier false positive twice), environmental health scientist, medical device specialist.

- **`EPA_search_frs_facilities`** — `_summarize`'s field mapping read `state_name`/`parent_registry_id` (unreliable/almost-always-null columns) instead of `state_code`/`registry_id` (the correct, populated ones). Independently reproduced by two personas.
- **`OpenFDA_search_drug_enforcement` / `_device_enforcement` / `_food_enforcement`** — each tool's own `count` parameter description gave example field values that openFDA's backend always rejects with HTTP 500 (missing the required `.exact` keyword suffix). Independently reproduced by two personas copying the broken examples straight from the docs.
- **CLI `tu run` error rendering** — `BaseRESTTool`-backed tools put the actionable upstream error in a `detail` field on non-2xx responses, but the CLI's error renderer only ever surfaced the generic top-level `error` string, silently dropping `detail`. Now surfaced as a `Detail:` line.

False positives: `AOPWiki_list_aops` and `EPA_search_tri_facilities` reported as referencing nonexistent tools (both exist and work — the persona's own discovery call had failed, likely the same MCP staleness noted below); `Orphanet_search_by_name`'s misspelled-query "wrong top results" claim didn't reproduce (the `total` count field is a genuine but cosmetic upstream quirk; actual top-ranked results were correct for both spellings).

Deferred: `OpenFoodFacts_filter_products_by_tags` allergen-tag mismatch (root cause traced to OpenFoodFacts' own search API, needs client-side taxonomy validation — bigger scope than one round).

## Round 4 (originally #298) — Oncology, microbiome, cardiology, ophthalmology, agricultural science

- **`cbioportal_tool.py`** — `_build_url` only substitutes `{placeholder}` tokens for keys present in the call arguments, so an omitted optional param (e.g. `limit`, schema default=20) left its literal `{limit}` unfilled, sending a broken query and causing a live 400 on `cBioPortal_get_cancer_studies`.
- **`gdc_tool.py`** — `GDCMutationFreqByProjectTool` read `case_summary.doc_count` as a project's total case count, but that field is unreliable (reported 2282 for TCGA-COAD when the real total is 461, understating KRAS mutation frequency in colorectal cancer as 11% instead of ~54%). Now fetches true totals from the `/projects` endpoint.
- **`dailymed_tool.py`** — SPL documents expose the same section content through multiple matching `<section>` elements (Highlights summary + Full Prescribing Information), so `DailyMed_parse_drug_interactions`/`parse_dosing` returned exact duplicates (20/10 unique, 103/79 unique for apixaban). Added a shared `_dedupe_items()` helper applied across all 5 SPL parser operations. Also: when `drug_name` was provided but the automatic setid lookup found no match, the tool returned "Missing required parameter: setid" — telling the caller to do exactly what they'd already done; now names the real problem.
- **`usda_plants_tool.py`** — `USDA_plants_search_by_name`'s common_name fallback left raw HTML markup un-stripped (e.g. `"<i>Holcus sorghum</i> L."`).
- **`cli.py`** — `tu grep -i` errored with "unrecognized arguments" since users reflexively type grep's `-i` flag; accepted as a no-op (text mode is already case-insensitive).

## Round 5 (originally #299) — Neurology, dermatology, psychiatry, OB-GYN, veterinary/comparative oncology

- **`clinvar_tool.py`** — `ClinVar_search_variants` already has an alias convention (`gene_symbol`→`gene`, etc.), but any other unrecognized key (e.g. `gene_name`) was silently dropped, producing a generic "at least one search parameter is required" error with no hint the parameter itself was misnamed. Now names any unrecognized parameter(s) directly.
- **`cpic_search_pairs_tool.py`** — `CPIC_get_recommendations` resolves drug names via CPIC's own `/drug` table, which only indexes generic names — a brand name like "zoloft" failed with "Use CPIC_list_guidelines to find valid guideline IDs," which doesn't solve the actual problem. Now names the real constraint and suggests the generic name.
- **`dailymed_tool.py`** — `_parse_dosing`/`_parse_drug_interactions`/`_parse_clinical_pharmacology` each extracted all `<table>` elements first, then all `<paragraph>` elements, in separate passes, destroying document order (SPL sections routinely interleave a heading paragraph immediately before the table it introduces). Added `_extract_ordered_content()` to walk a `<text>` element's children in true document order.

## Round 6 (originally #300) — Rheumatology, endocrinology, hematology, nephrology, pulmonology

- **`clinvar_tool.py`** — `ClinVar_search_variants`'s `clinical_significance` filter only indexed single-word clinsig values via `"clinsig <value>"[Filter]`; confirmed live and via raw NCBI curl that `"risk factor"`/`"likely pathogenic"` silently returned 0 results despite matching variants existing (e.g. HFE C282Y). ClinVar separately indexes compound values via `clinsig_<value>[prop]`; the query now ORs both forms (verified no regression on the existing single-word case).
- **`dailymed_tool.py`** — three independent personas hit the same bug: `DailyMed_search_spls`'s pagination fields returned the literal JSON string `"null"` instead of a real null, passed through verbatim from DailyMed's API — a caller's truthy check would misfire. Also: `DailyMed_parse_adverse_reactions`'s table-header extraction joined multi-line SPL cells (e.g. `"TRIKAFTA<br/>N=202<br/>n (%)"`) with no separator, producing mangled keys like `"TRIKAFTAN=202n (%)"`.

## Round 7 (originally #301) — Geriatrics, pediatrics, sports medicine, medical genetics counseling, critical care

- **`dailymed_tool.py`** (HIGH severity) — `DailyMed_parse_adverse_reactions` had no handling for `<list>/<item>` elements at all — only `<table>` or (if no table) `<paragraph>` text. Confirmed live that warfarin's adverse-reactions section encodes its actual reaction list as `<list><item>` blocks, so every real item (hemorrhage, tissue necrosis, calciphylaxis, AKI, etc.) was silently dropped, taking the reaction count from 4 generic sentences to the full 18-item list.
- **`unified_guideline_tools.py`** — `PubMed_Guidelines_Search` extracted abstracts via regex on raw, unparsed efetch XML, so entity references were never resolved. Confirmed via raw curl that PubMed's true source XML contains standard single-escaped numeric character references like `&#x2265;`, which the sibling `PubMed_search_articles` correctly decodes (real XML parser) but this tool's regex-based extraction left as literal text. Added `html.unescape()`.

Deferred (rounds 4-7, real but out of scope): several naming-intuition param mismatches with already-actionable errors (anti-alias-sprawl policy); DailyMed tables with no `<thead>` at all falling back to unlabeled lists (heuristic fix too risky without a reported case); `PubMed_Guidelines_Search`'s missing status/data envelope (reported 3x across rounds — systemic issue better suited to a dedicated audit, see issue #288's prior envelope batch); DailyMed's drug-name→setid auto-lookup picking the most-recent SPL rather than best name match (resolved a "ibuprofen" query to a combination product); ClinVar's `protein_change` field returning multiple concatenated values (confirmed genuine upstream NCBI data, not a ToolUniverse bug); GBIF/ClinicalTrials free-text relevance ranking gaps (upstream, not ToolUniverse-side).

---

## Known session-local issue (not a code bug, referenced throughout)

The MCP server in the development session used across all 7 rounds held a stale reference to a garbage-collected `uv` cache directory, causing TLS cert / "tool type not found" errors across `mcp__tooluniverse__*` calls. Every finding in every round was independently verified via the local CLI (`python3 -m tooluniverse.cli run`), which is unaffected — not something this PR changes.

## Test plan

- [x] 41 new/updated unit tests across all 7 rounds, all passing on the consolidated branch
- [x] Every one of the 26 fixes re-verified live via CLI after merging, confirming the merge composed all overlapping `dailymed_tool.py`/`clinvar_tool.py` changes correctly (no fix lost or altered)
- [x] Full touched-file regression suite run clean (2 skips are pre-existing missing-dependency skips, 1 pre-existing unrelated flaky test confirmed present on a clean `origin/main` baseline via stash-comparison — semantic-search ranking in `test_tu_cli.py`, not caused by any round's changes)
- [x] Auto-regen stub drift (`uv.lock`, `_lazy_registry_static.py`) reverted before commit — diff contains only intentional changes


gasvn:

PR #303, opened 2026-07-18
Test-harness round 8: ClinVar, ClinicalTrials domain-sweep fixes

gasvn:
## Summary

Round 8 of the ongoing test-harness sweep. Persona domains this round: dentistry/oral health, urology, otolaryngology, palliative care, anesthesiology. Every issue below was CLI-verified (`python3 -m tooluniverse.cli run <Tool> '<args>'`) before fixing, including raw NCBI E-utils curl checks.

## Verified bugs fixed

- **Fix-R8B-1** (`clinvar_tool.py`) — `ClinVar_search_variants`'s gene-symbol query passed the caller's string straight through to ClinVar's `[gene]` index, which only matches canonical unhyphenated HGNC symbols. Older literature/informal usage still hyphenates common gene names ("BRCA-2", "HER-2"), and a hyphenated query silently returned 0 results despite 21,879 real matches existing under the unhyphenated form (confirmed live and via raw NCBI E-utils curl). Hyphens are now stripped before building the query.

- **Fix-R5D-1/R8C-1** (`clinvar_tool.py`) — reimplements and extends an unrecognized-parameter check. A caller-supplied param that doesn't match any recognized name/alias (e.g. `variant_name` — the real param is `variant_id`) was silently dropped. When it was the only param supplied this produced a generic "at least one search parameter is required" error with no hint the parameter itself was misnamed; when other valid params were also supplied (confirmed live: `{"gene": "GJB2", "variant_name": "35delG"}`), the query still ran but silently ignored the unrecognized filter, with `total_count` reflecting all 739 GJB2 variants rather than the intended narrower search. Both cases now name the unrecognized parameter(s) — as an error when no valid params exist, and as a new `ignored_parameters` note in the success response when the query still runs on other valid params.

- **Fix-R8C-2** (`clinicaltrials_gov_tools.json`) — `ClinicalTrials_search_studies` had no `additionalProperties` restriction, so an unrecognized parameter (e.g. `intervention` — the real key is `query_intr`) was silently ignored rather than rejected, and the query ran anyway using only the recognized params (confirmed live). Added `"additionalProperties": false`, matching the same fix already applied to the sibling `search_clinical_trials` tool in an earlier round.

- **Fix-R8E-1/R6C-2** (`clinvar_tool.py`) — `ClinVarGetVariantDetails` and `ClinVarGetClinicalSignificance` both returned the full raw esummary blob twice — once as the top-level `data` key (from the underlying `_make_request` envelope) and again inside `formatted_data["raw_data"]` — roughly tripling payload size for no informational gain and making the two tools' outputs nearly indistinguishable from each other despite their differently-named, purpose-built `formatted_data` blocks. Dropped the redundant top-level copy from both; raw access remains available via `formatted_data.raw_data`.

These will resolve once earlier rounds' fixes merge (see #302, the consolidated PR for rounds 1–7):
- **R8B-2/R8D-2** — DailyMed table-cell text containing raw XML whitespace artifacts — already fixed by the `_cell_text()` helper from round 6 (Fix-R6E-2), applied to both header and data cells.
- **R8D-1** — `DailyMed_parse_contraindications`/`parse_dosing` returning duplicated content from overlapping SPL sections — already fixed by the `_dedupe_items()` helper from round 4 (Fix-R4C-1).

- **R8A-1** — `DailyMed_parse_dosing` intermittent HTTP 500 with an empty error body — the error message already surfaces everything the upstream response provides (`f"HTTP {status}: {text[:200]}"`); a transient empty-bodied upstream 500 isn't something a better message can fix without adding retry logic, a bigger design change.
- **R8A-2** — `PubMed_Guidelines_Search` AND-combines all query terms literally with no relaxation/fallback, so a natural multi-concept clinical query underperforms a simpler one — a real UX gap, but fixing it needs query-relaxation logic, not a quick change.
- **R8C-3** — `ClinicalTrials_search_studies`'s condition+intervention filters don't strictly exclude an unrelated trial even with correct parameter names — likely an upstream ClinicalTrials.gov query-construction/matching looseness, consistent with previously-deferred relevance-ranking gaps in other search tools.

## A note on repeated rediscovery

3 of this round's ~11 raw findings turned out to be duplicates of bugs already fixed in rounds 4 and 6 (currently sitting in the still-open, not-yet-merged #302). This is the second round in a row where a meaningful fraction of persona findings re-discover already-fixed-but-unmerged bugs — worth prioritizing a merge of #302 to stop this pattern before round 9.

## Test plan

- [x] 3 new mocked unit test files (`test_clinvar_gene_hyphen_and_params.py`, `test_clinvar_no_duplicate_raw_data.py`, `test_clinicaltrials_search_studies_schema.py`) — 9 new tests, all passing
- [x] Existing `test_ctg_tool.py` and `test_round007_fixes.py` re-run to confirm no regression
- [x] Every fix re-verified live via `python3 -m tooluniverse.cli run` after the code change, plus raw NCBI E-utils curl checks confirming root cause

### DIFF
--- a/src/tooluniverse/clinvar_tool.py
+++ b/src/tooluniverse/clinvar_tool.py
@@ -188,7 +188,15 @@ class ClinVarSearchVariants(ClinVarRESTTool):
         query_parts = []
 
         if "gene" in arguments:
-            query_parts.append(f"{arguments['gene']}[gene]")
+            # Fix-R8B-1: HGNC gene symbols don't contain hyphens (modern
+            # nomenclature phased them out decades ago), but older
+            # literature/informal usage still writes them hyphenated (e.g.
+            # "BRCA-2", "HER-2") -- ClinVar's [gene] index only matches the
+            # canonical unhyphenated form, so a hyphenated query silently
+            # returned 0 results (confirmed live and via raw NCBI E-utils:
+            # "BRCA-2[gene]" -> 0 hits, "BRCA2[gene]" -> 21879 hits).
+            gene = arguments["gene"].strip().replace("-", "")
+            query_parts.append(f"{gene}[gene]")
 
         if "condition" in arguments:
             # Feature-70B-005: [disease/phenotype] is not a valid ClinVar eSearch field.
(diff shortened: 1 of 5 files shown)


gasvn:

PR #304, opened 2026-07-18
Test-harness round 9: PubMed, ClinicalTrials, DailyMed domain-sweep fixes

gasvn:
## Summary

Round 9 of the ongoing test-harness sweep. Persona domains this round: neonatology, gynecologic oncology, transplant medicine, radiation oncology, sleep medicine. Every issue below was CLI-verified (`python3 -m tooluniverse.cli run <Tool> '<args>'`) before fixing, including raw upstream API curl checks to confirm root cause.

## Verified bugs fixed

- **Fix-R9C-1** (`pubmed_tool.py`) — `PubMedRESTTool._build_params` passed a query's hyphens straight through to NCBI's esearch `term` param. NCBI's automatic term-mapping treats a hyphenated compound (e.g. "CYP3A5-guided") as a literal `[All Fields]` phrase rather than decomposing it into individual words, which silently zeroed out an otherwise-matching multi-keyword AND query — confirmed via raw E-utils curl: 0 results for "CYP3A5-guided tacrolimus dosing kidney transplant" vs. 38 with the hyphen replaced by a space, while "COVID-19" (a legitimately hyphenated term) returns the identical count either way, ruling out regression risk. Hyphens are now replaced with spaces before building the query.

- **Fix-R9D-1** (`ctg_tool.py`) — ClinicalTrials.gov API v2's Essie search engine treats an unquoted multi-word `query.cond`/`query.intr` value as an implicit-OR keyword match rather than a phrase, so "cervical cancer" ranked unrelated head/neck and esophageal trials above actual cervical cancer trials — confirmed independently via raw curl against the live API: 11215 unquoted matches (mostly off-topic) vs. 2519 phrase-quoted matches (correctly topped by cervical-cancer-specific trials); the ToolUniverse tool call reproduced this exactly (11215 vs. 2519 after the fix). Added a `_phrase_quote_if_plain()` helper that phrase-quotes a plain multi-word value, but leaves anything already using quotes/parens/boolean operators untouched, since `query_cond`'s own docs advertise Boolean-operator support (e.g. "breast cancer AND HER2") that a blind wrap would break — verified live that this exact query is unaffected.

- **Fix-R9D-2** (`clinicaltrials_gov_tools.json`) — `intervention` and `sponsor` were live, functional parameter aliases in `ClinicalTrialsTool`'s Python code (mapping to `query.intr`/`query.spons`) but undeclared in the tool's JSON schema, so `tu info` never showed them. This also matters for a separate, still-unmerged round-8 fix (`additionalProperties: false`) — once merged together, that restriction would have started silently rejecting these two previously-working aliases as unrecognized properties had they not been declared here.

- **Fix-R9E-1** (`unified_guideline_tools.py`) — `PubMedGuidelinesTool.run()` returned `_search_pubmed_guidelines`'s bare list of results directly on success, never wrapped in the standard `{"status": "success", "data": [...]}` envelope every sibling tool (e.g. `PubMed_search_articles`) uses. This exact inconsistency was independently reported by personas across 4 separate rounds (R5D-2, R7B-1, R7D-2, R9E-1) — enough independent confirmation to fix directly rather than defer again as a "systemic, needs a dedicated audit" issue.

- **Fix-R9A-2/R9D-3** (`dailymed_tool.py`) — `DailyMedSPLParserTool`'s `metadata.drug_name` was only ever populated from the caller's own `drug_name` argument, so it was always `null` for the common case of calling with `setid` directly (e.g. after a prior `search_spls` call) — independently reported by two personas this round, on two different drugs (Curosurf, cisplatin) — even though the SPL itself already names the product. Falls back to the SPL's own `manufacturedProduct` name, already available on the parsed XML at no extra network cost.

Will resolve once earlier rounds' PRs merge (see #302 for rounds 1–7, #303 for round 8):
- **R9A-1/R9B-1** — DailyMed table-header whitespace concatenation — already fixed by round 6's `_cell_text()` helper (Fix-R6E-2).
- **R9A-3** — `DailyMed_parse_dosing` duplicate content from overlapping SPL sections — already fixed by round 4's `_dedupe_items()` helper (Fix-R4C-1).
- **R9B-1** — `DailyMed_search_spls` pagination `"null"` strings — already fixed by round 6 (Fix-R6A-2/R6D-3/R6E-3).

- **R9C-2** — `FDA_get_drug_interactions_by_drug_name` has no `status` key in its response envelope — same systemic, already-repeatedly-deferred envelope-consistency issue (see prior issue #288 batch) affecting the broader FDA tool family, larger scope than one round.

## Test plan

- [x] 4 new mocked unit test files (`test_pubmed_hyphen_query.py`, `test_ctg_phrase_quoting.py`, `test_pubmed_guidelines_envelope.py`, `test_dailymed_metadata_drug_name.py`) — 18 new tests, all passing
- [x] Existing `test_ctg_tool.py` re-run to confirm no regression
- [x] Every fix re-verified live via `python3 -m tooluniverse.cli run` after the code change, plus raw upstream API curl checks confirming root cause and no regression on edge cases (Boolean-operator queries, already-quoted values, "COVID-19")

### DIFF
--- a/src/tooluniverse/ctg_tool.py
+++ b/src/tooluniverse/ctg_tool.py
@@ -1,7 +1,27 @@
+import re
 from copy import deepcopy
 from urllib.parse import urljoin
 from .restful_tool import RESTfulTool, execute_RESTful_query
 from .tool_registry import register_tool
+
+_ESSIE_OPERATOR_RE = re.compile(r'"|[()\[\]]|\b(?:AND|OR|NOT|AREA)\b', re.IGNORECASE)
+
+
+def _phrase_quote_if_plain(value: str) -> str:
+    """Fix-R9D-1: CTG API v2's Essie search engine treats an unquoted
+    multi-word query.cond/query.intr value as an implicit-OR keyword match
+    rather than a phrase, so "cervical cancer" ranks unrelated head/neck
+    and esophageal trials above actual cervical cancer trials (confirmed
+    live: 11215 unquoted matches vs. 2519 when phrase-quoted, with the
+    quoted results correctly topped by cervical-cancer-specific trials).
+    Quoting the value as an Essie phrase fixes this -- but only for a
+    plain multi-word value with no existing quotes/parens/boolean
+    operators, since query_cond's own docs advertise Boolean-operator
+    support (e.g. "breast cancer AND HER2") that a blind wrap would
+    break."""
+    if " " in value and not _ESSIE_OPERATOR_RE.search(value):
+        return f'"{value}"'
+    return value
 
 
 @register_tool("ClinicalTrialsTool")
@@ -190,7 +210,12 @@ class ClinicalTrialsTool(RESTfulTool):
                     study_type_clause = f"({study_type_clause})"
                 advanced_clauses.append(study_type_clause)
             elif key in _SEARCH_PARAM_MAP:
-                params[_SEARCH_PARAM_MAP[key]] = value
+                mapped_key = _SEARCH_PARAM_MAP[key]
+                if mapped_key in ("query.cond", "query.intr") and isinstance(
+                    value, str
+                ):
+                    value = _phrase_quote_if_plain(value)
+                params[mapped_key] = value
 
         if advanced_clauses:
             params["filter.advanced"] = " AND ".join(advanced_clauses)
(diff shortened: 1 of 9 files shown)


gasvn:

PR #305, opened 2026-07-18
Test-harness round 10: ClinVar, DrugBank/MeSH, PubMed guideline-search fixes

gasvn:
## Summary

Round 10 of the ongoing test-harness sweep. Persona domains this round: hepatology, gastroenterology, vascular surgery, neurosurgery, occupational medicine. Every issue below was CLI-verified (`python3 -m tooluniverse.cli run <Tool> '<args>'`) before fixing, including raw upstream API curl checks to confirm root cause.

## Verified bugs fixed

- **Fix-R10D-1** (`clinvar_tool.py`) — ClinVar's `[gene]` index only matches a bare HGNC symbol, but a natural clinical phrasing like "NF2 gene" (used verbatim in real research questions, e.g. "look up the NF2 gene") silently returned 0 results, since the literal trailing word "gene" became part of the queried string — confirmed live and via raw NCBI E-utils curl: `"Nf2 gene[gene]"` → 0 hits, `"NF2[gene]"` → 2612 hits. A trailing/leading "gene"/"protein" qualifier word is now stripped before querying (verified across 4 phrasing variants, all now returning the same 2612 matches).

- **Fix-R10C-2** (`xml_tools.json`) — 19 XMLTool-family tools (`drugbank_*`, `mesh_*`) declared `default` values for `case_sensitive`/`exact_match`/`limit` but ALSO listed those same params in `required`, making the declared defaults completely unusable — every caller had to always pass all 4 params explicitly, defeating the point of having defaults at all (confirmed live: `{"query": "clopidogrel"}` alone errored with `"'case_sensitive' is a required property"` before the fix, on the exact tool a vascular surgeon persona hit while investigating clopidogrel interactions). This was a uniform copy-paste pattern across the whole file (confirmed by scanning all tool configs), fixed uniformly across all 19 affected tools by removing the defaulted params from `required`; removing items from `required` can only make validation more permissive, so no previously-working call (passing all 4 params explicitly) regresses.

- **Fix-R10E-3** (`unified_guideline_tools.py`) — `PubMedGuidelinesTool`'s title/authors extraction only ever read `title`/`authors`, but NCBI Bookshelf-type records (`doctype="book"`, e.g. WHO monographs and guideline documents) store their title under `booktitle` and have no individual author list — confirmed via raw esummary curl for PMID 34787987 ("WHO guideline for clinical management of exposure to lead"), whose `title`/`authors` were both empty while `booktitle` and `publishername` ("World Health Organization") had the real values. Falls back to those fields; verified a second book-type record with genuinely no `publishername` anywhere in NCBI's own data correctly still shows an empty author string (nothing left to fall back to), and that regular journal-article records are unaffected.

Will resolve once earlier rounds' PRs merge (see #302 for rounds 1–7, #303 for round 8, #304 for round 9):
- **R10D-3/R10A-2** — `ClinVar_get_variant_details`/`get_clinical_significance` duplicate `raw_data` payload — already fixed in round 8 (Fix-R8E-1/R6C-2).
- **R10E-2/R10A-2** — `PubMed_Guidelines_Search` missing status/data envelope — already fixed in round 9 (Fix-R9E-1).
- **R10E-4** — `DailyMed_search_spls` pagination `"null"` strings — already fixed in round 6 (Fix-R6A-2/R6D-3/R6E-3).

- **R10D-2** — ClinVar `protein_change` field returning multiple concatenated stop-codon calls for a single variant — confirmed in round 7 (Fix-R7D-1 investigation) as genuine upstream NCBI multi-transcript data, not a ToolUniverse-side bug.
- **R10C-1** — `drugbank_get_drug_interactions_by_drug_name_or_id` expects `query` not `drug_name_or_id` despite the tool's own name — error message already names the correct field clearly, consistent with prior rounds' guidance against unbounded alias sprawl.
- **R10B-1/R10A-1** — `DailyMed_parse_adverse_reactions`/`parse_drug_interactions` table rows sometimes lack column headers or have inconsistent dict-vs-list shapes for sub-rows under merged/rowspan cells — same class of "no `<thead>` present in this specific source table" issue deliberately deferred in round 6 (R6D-1/R6D-2): a "assume the first row is a header" heuristic risks misinterpreting genuinely headerless data tables.
- **R10E-1** — `PubMed_search_articles` silently returns 0 results for a long (8-term) multi-concept AND query even though shorter subsets of the same terms return real results — same class of query-relaxation gap deferred in round 8 (R8A-2); needs fallback/relaxation logic, not a quick fix.

## Test plan

- [x] 3 new test files (`test_clinvar_gene_qualifier_words.py`, `test_xml_tools_required_defaults.py`, `test_pubmed_guidelines_book_records.py`) — 13 new tests, all passing
- [x] Existing `test_round007_fixes.py` re-run to confirm no regression
- [x] Every fix re-verified live via `python3 -m tooluniverse.cli run` after the code change, plus raw upstream API curl checks confirming root cause
- [x] The 19-tool `xml_tools.json` fix's diff was manually inspected to confirm it touches only `required` arrays and nothing else was reformatted/reordered

### DIFF
--- a/src/tooluniverse/clinvar_tool.py
+++ b/src/tooluniverse/clinvar_tool.py
@@ -5,6 +5,7 @@ This tool provides access to the ClinVar database for clinical variant informati
 disease associations, and clinical significance data.
 """
 
+import re
 import requests
 import time
 from typing import Dict, Any, Optional
@@ -188,7 +189,21 @@ class ClinVarSearchVariants(ClinVarRESTTool):
         query_parts = []
 
         if "gene" in arguments:
-            query_parts.append(f"{arguments['gene']}[gene]")
+            # Fix-R10D-1: ClinVar's [gene] index only matches a bare HGNC
+            # symbol, but a natural clinical phrasing like "NF2 gene" (used
+            # verbatim in real research questions -- e.g. "look up the NF2
+            # gene") silently returns 0 results, since the literal trailing
+            # word "gene" becomes part of the queried string (confirmed
+            # live and via raw NCBI E-utils curl: "Nf2 gene[gene]" -> 0
+            # hits, "NF2[gene]" -> 2612 hits). Strip a trailing/leading
+            # "gene"/"protein" qualifier word before querying.
+            gene = re.sub(
+                r"^\s*(?:gene|protein)\s+|\s+(?:gene|protein)\s*$",
+                "",
+                arguments["gene"],
+                flags=re.IGNORECASE,
+            ).strip()
+            query_parts.append(f"{gene}[gene]")
 
         if "condition" in arguments:
             # Feature-70B-005: [disease/phenotype] is not a valid ClinVar eSearch field.
--- a/src/tooluniverse/data/xml_tools.json
+++ b/src/tooluniverse/data/xml_tools.json
@@ -54,10 +54,7 @@
         }
       },
       "required": [
-        "query",
-        "case_sensitive",
-        "exact_match",
-        "limit"
+        "query"
       ]
     },
     "return_schema": {
@@ -229,10 +226,7 @@
         }
       },
       "required": [
-        "query",
-        "case_sensitive",
-        "exact_match",
-        "limit"
+        "query"
       ]
     },
     "return_schema": {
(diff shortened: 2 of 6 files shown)


gasvn:

PR #307, opened 2026-07-18
Round 12: PRIDE search, WHO GHO param mapping, Semantic Scholar bulk-sort, NCBI lineage names

gasvn:
## Round 12: bibliometrics, biodiversity/taxonomy, proteomics, public health, metabolomics

Fifth round of the ongoing test-harness sweep (continues #302 → #303 → #304 → #305 → #306). Five role-play researcher personas exercised tool families not covered by earlier rounds — bibliometrics (Semantic Scholar / OpenAlex / Crossref), biodiversity/taxonomy (GBIF / NCBI Taxonomy / IUCN), proteomics (PRIDE / UniProt PTM), public health surveillance (WHO GHO), and metabolomics (KEGG / Reactome / Metabolite DB) — chosen specifically to avoid re-discovering bugs already fixed but not yet merged in earlier rounds. Every finding below was CLI-verified and root-caused against the live upstream API (raw `curl`) before any fix.

### Fixed

**Fix-R12C-1 — `PRIDE_search_proteomics` ignored `query` and `page_size` entirely (HIGH)**
The tool's config pointed at `GET /projects?query={query}`. Raw curl confirmed PRIDE's `/projects` endpoint ignores `query` completely — a real query (`TP53`) and a nonsense string (`xyzzyzzynonexistentquery12345`) both returned the identical unfiltered 100 projects. The actual keyword-search endpoint is `/search/projects?keyword={query}&pageSize=N`, confirmed live to filter correctly (0 results for the nonsense query, TP53-relevant hits for `TP53`). `page_size` was also dead — the static `fields.params: {"pageSize": "20"}` in the JSON config was never read by any code path. Fixed by pointing the config at the real endpoint and wiring `page_size` into `_build_url` (falls back to the schema's advertised default of 20; confirmed live PRIDE otherwise defaults to 100/page).
`src/tooluniverse/pride_tool.py`, `src/tooluniverse/data/pride_tools.json`

**Fix-R12C-3 — `PRIDE_get_projects_for_protein` gave an unactionable error for wrong identifier type (MEDIUM)**
Passing a gene symbol (`TP53`) instead of a UniProt accession (`P04637`) returned a bare "PRIDE API error" — confirmed PRIDE returns an empty-body HTTP 404 for any unrecognized identifier, so there's no upstream detail to surface. Added a 404-specific message hinting the identifier format is the likely problem, without guessing a specific resource type (shared code path across all PRIDE tools).
`src/tooluniverse/pride_tool.py`

**Fix-R12B-1 — `NCBIDatasets_get_taxonomy`'s `lineage` was unusable ancestor IDs with no names (MEDIUM)**
`lineage` returned 30 bare numeric tax_ids with zero names/ranks, despite the tool's description promising "full lineage" — confirmed this is exactly what NCBI's own Datasets v2 API returns unmodified, not something dropped by the tool. NCBI's endpoint accepts a comma-joined batch of tax_ids in one request (confirmed live), so added a `lineage_names` field that resolves the whole lineage's `organism_name`/`rank` in a single extra batched call rather than one call per ancestor. The original `lineage` field is untouched (still bare ints) for backward compatibility; enrichment degrades gracefully to `[]` on failure rather than failing the whole call.
`src/tooluniverse/ncbi_datasets_tool.py`, `src/tooluniverse/data/ncbi_datasets_tools.json`

**Fix-R12D-1 — `WHOGHO_get_indicator_data`'s `filter`/`top` params were silently dropped (HIGH)**
Requesting `{"filter":"SpatialDim eq 'USA'","top":3}` returned the same unfiltered 20-row page every time (verified with 3 different filter/top combinations, and cross-checked that the raw WHO GHO OData API honors `$filter`/`$top` fine). Root cause: unlike its sibling `WHOGHO_search_indicators`, this tool's config had no `fields.param_mapping` block, so the generic `BaseRESTTool` never translated `filter`/`top` into the OData `$filter`/`$top` query params. Silent — no error, just wrong (unfiltered) data, which is the worst failure mode for a "compare across countries" workflow. Pure JSON config fix: added the missing `param_mapping` block, identical to the working sibling tool's.
`src/tooluniverse/data/who_gho_tools.json`

**Fix-R12A-1 — `SemanticScholar_search_papers` with `sort` failed with a deterministic 400 (HIGH)**
Any `sort` value (e.g. `citationCount:desc`) switches the tool to Semantic Scholar's `/paper/search/bulk` endpoint (needed because the plain `/paper/search` endpoint silently ignores `sort`). Raw curl isolated the exact cause: the bulk endpoint rejects the `tldr` field with `400 {"error": "Unrecognized or unsupported fields: [tldr]"}`, even though `/paper/search` supports it fine. The bulk endpoint also has no `limit` concept (only token-based pagination over its full result set), so `limit` was being silently ignored server-side too. Fixed by dropping `tldr` from the fields list only on the bulk path (kept on the normal path), and slicing results to the requested `limit` client-side when sorting.
`src/tooluniverse/semantic_scholar_tool.py`

**Fix-R12A-2 — `Crossref_search_works` + `sort` silently discards query relevance (MEDIUM, docs-only)**
`{"query":"CRISPR Cas9 genome editing","sort":"is-referenced-by-count","order":"desc"}` returned globally top-cited works (DESeq2, SAMtools, GSEA...) with zero relevance to the query. Confirmed via identical raw curl to `api.crossref.org` that this is genuine upstream Crossref behavior — combining `sort` with `query` weakens/discards the BM25 relevance matching — not a ToolUniverse bug. Added a caveat to the tool description warning about this known combination pitfall.
`src/tooluniverse/data/crossref_tools.json`

- **R12C-2** — `PRIDE_get_projects_for_protein` rejects `protein` as a param name with a generic jsonschema "'accession' is a required property" error rather than naming the unrecognized key. Error is already actionable (states the correct key); this is a naming-intuition mismatch, not a broken tool — consistent with prior rounds' treatment of similar cases.
- **R12D-2** — `WHOGHO_get_indicator_data` on an invalid indicator code returns a generic `"{api_name} API error"` (no status code/detail surfaced by the CLI's default rendering). Confirmed WHO GHO returns an empty-body 404, so there's genuinely nothing more to surface for this case; the message itself comes from `base_rest_tool.py`, shared by many unrelated tool families — broadening it needs a dedicated audit (same precedent as the prior envelope-consistency batch, issue #288), not a one-off tweak here.
- **R12D-3** — `WHS2_161` (a cardiovascular indicator surfaced by `WHOGHO_search_indicators`) returns zero data rows. Confirmed via raw curl this is a genuine upstream WHO GHO data gap (stale/retired indicator still listed in the catalog), not a tool bug.
- **R11 carryover items** (UniProt bare-envelope, GWAS error/empty-success inconsistency, STRING type mismatch, CxGDisc relevance ranking, ChEMBL nested output, ZINC22 timeout) — unchanged from round 11, still deferred for the same reasons documented there.

### Testing

18 new unit tests across 5 files (all mocked, no network calls):
- `tests/unit/test_pride_search_proteomics.py` (6)
- `tests/unit/test_ncbi_datasets_lineage_names.py` (4)
- `tests/unit/test_who_gho_indicator_data_param_mapping.py` (3)
- `tests/unit/test_semantic_scholar_bulk_sort.py` (4)
- `tests/unit/test_crossref_sort_caveat.py` (1)

Every fix was also verified live against the real upstream API via the CLI before and after the change (see fix descriptions above for the specific before/after evidence).

### DIFF
--- a/src/tooluniverse/data/crossref_tools.json
+++ b/src/tooluniverse/data/crossref_tools.json
@@ -2,7 +2,7 @@
   {
     "type": "CrossrefRESTTool",
     "name": "Crossref_search_works",
-    "description": "Search Crossref Works API for scholarly articles, publications, and research outputs by keyword. Returns metadata including title, abstract, journal, authors, publication year, DOI, URL, citations, and publisher. Supports filtering by publication type (journal article, conference paper, etc.) and date range. Use this to discover academic literature, verify citations, or build bibliographies. Find DOIs to use with Crossref_get_work for complete metadata.",
+    "description": "Search Crossref Works API for scholarly articles, publications, and research outputs by keyword. Returns metadata including title, abstract, journal, authors, publication year, DOI, URL, citations, and publisher. Supports filtering by publication type (journal article, conference paper, etc.) and date range. Use this to discover academic literature, verify citations, or build bibliographies. Find DOIs to use with Crossref_get_work for complete metadata. Note: Crossref's own API weakens keyword relevance-matching whenever a non-default 'sort' query param is combined with 'query' (e.g. sort=is-referenced-by-count), returning globally top-cited works largely unrelated to the search term -- omit 'sort' (the default relevance ranking) unless you have separately verified the combination for your query.",
     "parameter": {
       "type": "object",
       "properties": {
--- a/src/tooluniverse/data/ncbi_datasets_tools.json
+++ b/src/tooluniverse/data/ncbi_datasets_tools.json
@@ -474,7 +474,7 @@
     },
     {
         "name": "NCBIDatasets_get_taxonomy",
-        "description": "Get detailed taxonomy information from NCBI by taxonomy ID. Returns organism name, rank, full lineage, counts of genes/assemblies/RNA types, and child taxa. Covers all organisms in NCBI Taxonomy including bacteria, archaea, eukaryotes, and viruses. Example: tax_id 9606 returns Homo sapiens with 20,625 protein-coding genes.",
+        "description": "Get detailed taxonomy information from NCBI by taxonomy ID. Returns organism name, rank, full lineage (as bare ancestor tax_ids in `lineage`, and as resolved organism_name/rank pairs in `lineage_names`), counts of genes/assemblies/RNA types, and child taxa. Covers all organisms in NCBI Taxonomy including bacteria, archaea, eukaryotes, and viruses. Example: tax_id 9606 returns Homo sapiens with 20,625 protein-coding genes.",
         "parameter": {
             "type": "object",
             "properties": {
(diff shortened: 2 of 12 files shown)


gasvn:

PR #308, opened 2026-07-18
Round 13: WormBase gene-ID resolution, Wikidata schema defaults, gated-tool discovery

gasvn:
## Round 13: toxicology, developmental biology, immunology, patent search, agricultural science

Sixth round of the ongoing test-harness sweep (continues #302 → #303 → #304 → #305 → #306 → #307). Five role-play researcher personas exercised entirely fresh tool families — toxicology (CTD), developmental/model-organism biology (WormBase, Xenbase, VEuPathDB), immunology (VDJdb, IEDB), patent/IP analysis (USPTO), and agricultural science (USDA Plants, Wikidata) — continuing the domain-diversification strategy from rounds 11-12 to avoid rediscovering bugs already fixed in earlier, still-unmerged rounds. Immunology (VDJdb/IEDB) came back completely clean: 7 tool calls, cross-database concordance, zero issues.

### Fixed

**Fix-R13B-1 / Fix-R13B-2 — `WormBase_get_gene` 500'd on a "WB:"-prefixed ID or a bare gene symbol (HIGH)**
`_resolve_wbgene_id` had two independent bugs, both confirmed live:
- A "WB:"-prefixed WBGene ID (`WB:WBGene00006746` — exactly what the sibling tool `Alliance_search_genes` returns for the same gene) wasn't recognized as already-resolved (`startswith("WBGENE")` doesn't match `"WB:WBGENE..."`), so the raw colon-containing string was interpolated straight into the WormBase REST URL, which 500'd.
- A bare gene symbol (`unc-6`) fell through to an Alliance Genome Resources lookup using `category=gene`/`species=...` query params on `/api/search` — confirmed via raw curl these params are no longer honored and always return zero results. The working endpoint (already used correctly by the `alliance_genome_tool.py` sibling) is `/api/search_autocomplete` with client-side filtering by category and species curie prefix.
Both now resolve correctly; verified live for both `WB:WBGene00006746` and `unc-6`.
`src/tooluniverse/wormbase_tool.py`

**Fix-R13E-1 — `Wikidata_search_entities` disguised a MediaWiki failure as success (HIGH)**
The `language` param was documented as defaulting to `'en'` but had no `default` in the JSON schema, so omitting it sent no `language` param at all. Confirmed live that MediaWiki's `wbsearchentities` then returns HTTP 200 with `{"error": {"code": "missingparam", ...}}` embedded in the body — ToolUniverse reported this as `status: success`, requiring the caller to inspect payload contents to discover the call actually failed. Added the schema default so the documented behavior is real.
`src/tooluniverse/data/wikidata_entity_tools.json`

**Fix-R13D-1 — `tu grep`/`tu info` made real, gated tools look nonexistent (HIGH)**
Tools with an unmet `required_api_keys` (here: 6 USPTO patent-search tools needing `USPTO_API_KEY`) are correctly excluded from the active tool registry — but this made `tu grep uspto` return 0 matches and `tu info get_patent_overview_by_text_query` return "Tool not found", both indistinguishable from the tool genuinely not existing. `tu run` on the exact same tool name already gives the correct, actionable "requires API key(s) not set: USPTO_API_KEY" message via an existing `_excluded_api_key_tools` tracking dict — it just wasn't wired into the discovery commands. Fixed by threading that same lookup into `GetToolInfoTool` and `GrepToolsTool` (`tool_discovery_tools.py`), and fixing `cli.py`'s `_render_info` (which previously hardcoded every error to a generic "not found", discarding any more specific message) and `_render_grep` (renders the new `gated_matches` list) to surface it. `tu grep get_patent_overview` and `tu info get_patent_overview_by_text_query` now both correctly report the missing API key instead of "doesn't exist."
`src/tooluniverse/tool_discovery_tools.py`, `src/tooluniverse/cli.py`

**Fix-R13E-3 — `Wikidata_SPARQL_query`'s `max_results` was required despite being documented and implemented as optional**
The schema listed `max_results` in `required`, but its own description says "Optional result limit override... If not specified, uses the LIMIT clause," and the Python implementation already does `arguments.get("max_results")` with graceful fallback. Every call omitting it failed validation before ever reaching that correctly-written handling. Removed from `required`.
`src/tooluniverse/data/wikidata_sparql_tools.json`

**Feature-R13A-1 — `CTD_get_chemical_diseases` description over-promised evidence-type data (docs-only)**
The description promised "direct evidence type (therapeutic, marker/mechanism)" for every chemical-disease association. Confirmed via raw curl to RENCI's Automat CTD mirror that the actual edge properties for chemical-disease pairs are `{agent_type, knowledge_level, primary_knowledge_source, publications}` — no `predicate`/`qualified_predicate` key at all, unlike the sibling chemical-gene-interaction edges which do carry one. Genuine upstream data gap, not a code bug (the shared `_format_edge` extraction is correct); added a caveat to the description instead of a code change.
`src/tooluniverse/data/ctd_tools.json`

- **R13A-2** — `PubChemTox_get_toxicity_summary` rejects a guessed `chemical_name` param (correct name: `compound_name`). Error is already actionable; naming-intuition mismatch, not a broken tool.
- **R13B-3** — `VEuPathDB_search_genes_by_organism` returns a raw WDK validation error for an out-of-scope organism (C. elegans isn't covered by VEuPathDB, which is pathogen/vector-focused) — the underlying negative result is actually correct, just opaquely worded.
- **R13D-2** — `tu status` has no way to enumerate all 87 gated tools by name/required-key; only a bare count. The core discoverability blocker (a specific gated tool being indistinguishable from nonexistent) is fixed by R13D-1 above; full enumeration is a separate, larger UX enhancement.
- **R13D-3** — `PubMed_search_articles` for "CRISPR Cas9 patent" returned only tangentially relevant results (CAR-T papers mentioning CRISPR in passing). Confirmed a PubMed relevance-ranking characteristic, not a tool defect — PubMed doesn't index patents and "patent" as a free-text term dilutes relevance.
- **R13E-2** — `USDA_plants_get_profile`'s `fips_distribution` field is always null. Confirmed via raw curl this is genuinely null on the `PlantProfile` endpoint itself (even though a `HasDistributionData: true` flag suggests the data exists elsewhere) — probing several plausible distribution-endpoint URL patterns (`PlantDistribution/{id}`, `PlantProfile/{id}/Distribution`, etc.) all 404'd. The real endpoint likely requires live network-tab inspection of the USDA Plants website to find; deferring rather than guessing further.

### Testing

19 new unit tests across 4 files (all mocked, no network calls):
- `tests/unit/test_wormbase_gene_id_resolution.py` (5)
- `tests/unit/test_wikidata_schema_fixes.py` (4)
- `tests/unit/test_gated_tool_discovery.py` (9)
- `tests/unit/test_ctd_disease_predicate_caveat.py` (1)

Also re-ran the full pre-existing `tests/tools/test_tu_cli.py` suite (369 tests) since the `tu grep`/`tu info` changes touch shared CLI rendering code — all green except one already-documented, pre-existing flaky test (`test_find_tool_name_with_underscores_works`, a semantic-search ranking issue unrelated to this round, confirmed in earlier rounds to fail identically against a clean `origin/main` baseline).

Every fix was also verified live against the real upstream API via the CLI before and after the change (see fix descriptions above for the specific before/after evidence).

### DIFF
--- a/src/tooluniverse/cli.py
+++ b/src/tooluniverse/cli.py
@@ -258,6 +258,21 @@ def _render_grep(d: dict) -> str:
             if d.get("limit") == 0:
                 return f"0 of {total} matches (limit=0, no results shown)"
             return f"0 of {total} matches (offset past end — use --offset < {total})"
+        # Fix-R13D-1: surface tools that exist but are hidden because a
+        # required API key is unset -- without this, `tu grep uspto` looked
+        # identical to "no such tools exist" (confirmed live), when in fact
+        # the tools are real and just need a key set.
+        gated = d.get("gated_matches")
+        if gated:
+            lines = [
+                f"0 loaded matches, but {len(gated)} gated tool(s) matched by name:"
+            ]
+            for g in gated:
+                lines.append(
+                    f"  {g['name']}  (requires: {', '.join(g['missing_api_keys'])})"
+                )
+            lines.append("  Set the API key(s) as environment variables, then retry.")
+            return "\n".join(lines)
         # Feature-R13A-01 / R21B-03: context-sensitive hints for 0 name-field matches.
         if d.get("field") == "name":
             pattern = d.get("pattern", "")
@@ -377,7 +392,16 @@ def _render_info(d: dict) -> str:
     """Render get_tool_info result as human-readable tool card."""
     if "error" in d:
         name = d.get("name", "")
+        error_msg = d.get("error", "")
         suggestions = d.get("suggestions", [])
+        # Fix-R13D-1: this used to hardcode "not found" for every error here,
+        # discarding a more specific message like "requires API key(s) not
+        # set: X" (confirmed live this contradicted `tu run`'s own error for
+        # the exact same tool name, which does surface the real reason).
+        # Only fall back to the generic not-found/"did you mean" framing when
+        # the tool is genuinely absent from the registry.
+        if name and error_msg and "not found" not in error_msg:
+            return f"Error: Tool '{name}' {error_msg}"
         # R22B-04: append "Did you mean?" hint when suggestions are available.
         if name:
             hint = ""
@@ -981,10 +1005,13 @@ def cmd_info(args: argparse.Namespace) -> None:
         )
     # R22B-04: inject "did you mean" suggestions into each not-found error entry so
     # _render_info can display them without needing direct access to `tu`.
+    # Fix-R13D-1: skip this for a gated tool (error != "not found") -- the
+    # tool name is already exact, so suggesting near-miss alternatives is
+    # noise, not help.
     if isinstance(result, dict) and "tools" in result:
         all_names = list(tu.all_tool_dict.keys())
         for tool in result["tools"]:
-            if isinstance(tool, dict) and "error" in tool:
+            if isinstance(tool, dict) and tool.get("error") == "not found":
                 name = tool.get("name", "")
                 if name:
                     # Feature-23B-04: raised cutoff from 0.5 → 0.62 to avoid spurious
--- a/src/tooluniverse/data/ctd_tools.json
+++ b/src/tooluniverse/data/ctd_tools.json
@@ -176,7 +176,7 @@
   },
   {
     "name": "CTD_get_chemical_diseases",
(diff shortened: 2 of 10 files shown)


gasvn:

PR #309, opened 2026-07-18
Round 14: dbSNP not-found detection, BRENDA SABIO-RK failure visibility, ArrayExpress sample extraction

gasvn:
## Round 14: neuroscience, structural biology, functional genomics, enzymology, clinical genomics

Seventh round of the ongoing test-harness sweep (continues #302 → ... → #308). Five role-play researcher personas exercised entirely fresh tool families — neuroscience (Allen Brain Atlas, Allen Cell Types), structural biology (CATH, ClusPro/docking coverage), functional genomics (ArrayExpress, BioStudies), enzymology (BRENDA), and clinical genomics (dbSNP, CADD) — continuing the domain-diversification strategy that's kept every round since 11 free of duplicate rediscoveries.

### Fixed

**Fix-R14E-1 — `dbsnp_get_variant_by_rsid`/`dbsnp_get_frequencies` silently "succeeded" on a nonexistent rsID (HIGH)**
NCBI's `esummary.fcgi` doesn't 404 for a bad rsID — it returns a per-ID entry shaped `{"uid": "...", "error": "cannot get document summary"}` (confirmed live via raw curl). Both tools only checked whether the rsid was a key in the response dict, not whether that entry itself carried an error marker, so a nonexistent variant silently returned `status: "success"` with every field null/empty — a caller checking only `status` would proceed with a garbage record. Both methods now check for the embedded error marker and return an actionable "not found" error instead.
`src/tooluniverse/dbsnp_tool.py`

**Fix-R14D-1 — `BRENDA_get_enzyme_kinetics` silently dropped its headline capability on any SABIO-RK failure (HIGH)**
The entire SABIO-RK section (kinetic parameter fetching — the tool's own advertised primary purpose) was wrapped in one bare `except Exception: pass`. A connection timeout to `sabiork.h-its.org` (confirmed live, reproducible from this network) caused `kinetic_parameters`/`sabiork_total_entries` to be omitted from the response entirely, with `status: "success"` and zero indication that kinetics data was even attempted. Split into a `Timeout`-specific branch and a generic-exception branch, both of which now populate empty defaults for those keys and record a `sabiork_error` message in `metadata` — the caller can now tell "checked SABIO-RK, zero real entries" apart from "SABIO-RK fetch failed."
`src/tooluniverse/brenda_tool.py`

**Fix-R14C-2 — `arrayexpress_get_experiment_samples` returned the wrong section entirely (MEDIUM)**
`_extract_samples_from_section` matched on section `type in ["samples", "sample"]`, which matched the "Samples" *wrapper* section — aggregate summary attributes only (`"Sample count"`, plus one `"Experimental Factors"` attribute per factor that a dict-build silently collapsed to just the last one on a name collision). The real per-sample-group records live one level deeper, as `"Characteristics Table"` entries under a `"Source Characteristics"` subsection (confirmed via raw BioStudies API for `E-GEOD-33447`). Fixed the match to target the real per-record level; verified the fix generalizes correctly against a second, structurally-different accession (`E-GEOD-48162`, a ChIP-seq study with a completely different characteristics schema).
`src/tooluniverse/arrayexpress_tool.py`

**Feature-R14A-2 — `AllenCellTypes_search_specimens` silently ignored an unrecognized `gene` param (MEDIUM)**
This tool wraps electrophysiology/morphology specimen data, not transcriptomics — but passing `gene` (a natural first guess) was silently dropped instead of rejected, returning unrelated specimens as if the request had been fully satisfied. Verified `AllenCellTypesSpecimensTool.run()` only ever reads its three declared params, so added `"additionalProperties": false` to the schema — unknown params are now rejected with a clear jsonschema error.
`src/tooluniverse/data/allen_cell_types_tools.json`

**Feature-R14B-1 / Feature-R14C-1 — two description caveats for confirmed upstream data gaps (docs-only)**
- `CATH_get_superfamily`'s description unconditionally promised a classification name. Confirmed via raw curl to `cathdb.info` that `classification_name`/`classification_description` are genuinely `null` in CATH's own API for some superfamilies (e.g. `1.10.530.10`, the lysozyme-like fold) — the same extraction code correctly returns the name for other superfamilies (e.g. `2.40.50.140`), so this is upstream curation, not a ToolUniverse bug.
- `biostudies_search`'s description didn't warn that common biomedical phrases are dominated by `S-EPMC*` (EuropePMC-collection literature) hits rather than actual datasets. Confirmed live for `"breast cancer transcriptomics"`. Added a caveat pointing to `biostudies_search_by_collection` for dataset-only searches.
`src/tooluniverse/data/cath_tools.json`, `src/tooluniverse/data/biostudies_tools.json`

- **R14A-1** — `AllenBrain_search_genes`'s description says "by gene symbol" but the actual param is `gene_acronym`. Error message already names the correct param; wording drift, not a broken tool.
- **R14A-3** — No ToolUniverse tool wraps Allen Institute's cell-type *transcriptomic* API (only the electrophysiology/morphology one exists) — a genuine coverage gap, not a bug in an existing tool.
- **R14B-2** — `CATH_get_domain_summary` returns a raw HTTP 400 for an invalid domain suffix (e.g. `1lyzA01` when only `1lyzA00` exists) with no hint about CATH's domain-numbering convention. Correct rejection, just an unhelpful message on a fairly rare edge case.
- **R14B-3** — No ToolUniverse tool wraps ClusPro or any protein-protein docking service (only protein-*ligand* docking exists). Coverage gap, confirmed via 5 independent `tu grep` queries before concluding absence.
- **R14D-2** — `UniProt_get_function_by_accession` returns a bare `{"result": [...]}` envelope inconsistent with the `status`/`data`/`metadata` convention used elsewhere. Matches the long-documented, systemic envelope-consistency issue (same class as the prior issue #288 batch) — needs a dedicated audit, not a one-off fix here.

### Testing

14 new unit tests across 5 files (all mocked, no network calls):
- `tests/unit/test_dbsnp_nonexistent_rsid.py` (4)
- `tests/unit/test_brenda_sabiork_failure_visibility.py` (3)
- `tests/unit/test_arrayexpress_samples_extraction.py` (2)
- `tests/unit/test_allen_cell_types_schema.py` (3)
- `tests/unit/test_round14_docs_caveats.py` (2)

Every fix was also verified live against the real upstream API via the CLI before and after the change (see fix descriptions above for the specific before/after evidence, including cross-checking the ArrayExpress fix against a second, structurally-different real accession).

### DIFF
--- a/src/tooluniverse/arrayexpress_tool.py
+++ b/src/tooluniverse/arrayexpress_tool.py
@@ -235,9 +235,19 @@ class ArrayExpressRESTTool(BaseTool):
         """Extract sample information from a BioStudies section"""
         samples = []
 
-        # Check if current section is a sample
+        # Fix-R14C-2: the literal "Samples" section is just a wrapper
+        # carrying aggregate summary attributes (e.g. "Sample count", plus
+        # one "Experimental Factors" attribute per factor -- which the dict
+        # build below collapses to just the last one on name collision) --
+        # not per-sample data. Confirmed live via the raw BioStudies API
+        # (E-GEOD-33447) that the actual per-sample-group records live one
+        # level deeper, under a "Source Characteristics" subsection, as
+        # "Characteristics Table" entries (one per distinct combination of
+        # characteristics, e.g. age/tissue). Match those instead of the
+        # aggregate wrapper; "sample" (singular) is kept for any submission
+        # that uses that type directly on an individual record.
         section_type = section.get("type", "")
-        if section_type.lower() in ["samples", "sample"]:
+        if section_type.lower() in ["sample", "characteristics table"]:
             # Extract sample attributes
             if "attributes" in section and isinstance(section["attributes"], list):
                 sample_data = {}
--- a/src/tooluniverse/brenda_tool.py
+++ b/src/tooluniverse/brenda_tool.py
@@ -466,6 +466,7 @@ class BRENDATool(BaseTool):
 
         sources_used = []
         result: Dict[str, Any] = {"ec_number": ec_number}
+        sabiork_error: Optional[str] = None
 
         # 1. ExPASy ENZYME: enzyme identity and catalytic activity
         try:
@@ -480,6 +481,13 @@ class BRENDATool(BaseTool):
             pass  # Non-fatal, continue with kinetics
 
         # 2. SABIO-RK: kinetic parameters (Km, kcat, Ki, Vmax)
+        # Fix-R14D-1: this whole block used to be one bare `except Exception:
+        # pass`, so any failure here (confirmed live: sabiork.h-its.org
+        # connection timeouts) silently dropped "kinetic_parameters" from
+        # the response entirely -- the tool's own headline capability --
+        # while still returning status:"success" with no indication that
+        # SABIO-RK was even attempted, let alone that it failed. Record the
+        # failure instead of erasing all trace of it.
         try:
             sabio = self._fetch_sabiork_kinetics(ec_number, organism, limit)
             result["kinetic_parameters"] = sabio.get("kinetic_laws", [])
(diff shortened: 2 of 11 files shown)


gasvn:

PR #311, opened 2026-07-18
Round 16: PubMed date-range filter, NHANES dataset resolution, HPO namespace leakage

gasvn:
## Round 16: clinical genetics, epidemiology, structural bioinformatics, pharmacology, immunogenetics

Ninth round of the ongoing test-harness sweep (continues #302 → ... → #310). Five role-play researcher personas exercised fresh tool families — clinical genetics (HPO, Monarch), epidemiology (NHANES), structural bioinformatics (MobiDB, IUPred3), pharmacology (Guide to Pharmacology), and immunogenetics (IPD-IMGT/HLA, HLA Ligand Atlas). This was an unusually high-value round: four confirmed HIGH-severity bugs, one of them (PubMed date filtering) in a tool used constantly across the entire campaign.

### Fixed

**Fix-R16E-1 — `PubMed_search_articles`'s `mindate`/`maxdate` silently had zero effect when used alone (HIGH — largest blast radius of any fix in this campaign)**
Confirmed via multiple raw curl tests against NCBI's `esearch.fcgi`: an open-ended date range (`mindate` supplied without `maxdate`, or vice versa — the natural way a caller would ask "articles from 2024 onward") returns the **exact same unfiltered result count** as no date filter at all. Only when both bounds are present together does NCBI actually apply the filter. Separately confirmed that omitting `datetype` (even with both dates set) silently shifts the filter basis from publication date to Entrez date. `_build_params` now backfills whichever bound is missing with an out-of-PubMed's-real-range placeholder (`"1000"`/`"3000"`) and defaults `datetype` to `"pdat"` whenever either date bound is supplied. Verified live: a query that previously always included a 2017 article regardless of `mindate:"2024/01/01"` now correctly excludes it, and the symmetric `maxdate`-only case works too.
`src/tooluniverse/pubmed_tool.py`

**Fix-R16B-1 / Fix-R16B-2 — `NHANES_download_and_parse` had two bugs breaking the discover→fetch workflow (HIGH)**
- Requesting the `Examination` component without `dataset_name` silently downloaded the wrong file (`BPX_J`, blood pressure) instead of the one implied by the requested `variables` (`BMXBMI`/`BMXWT`, body measures) — returning `status: "success"` with the requested variables silently missing. `Examination` and `Questionnaire` each span many distinct files, exactly like the already-guarded `Laboratory` component; extended that same "requires `dataset_name`" guard to both.
- Passing `dataset_name` with the cycle suffix already included (exactly what the sibling tool `nhanes_search_datasets` returns, e.g. `"BMX_J"`) caused the suffix to be appended a second time, producing a 404 downloading `BMX_J_J.XPT`. Fixed with an `endswith(suffix)` check before appending.
`src/tooluniverse/nhanes_tool.py`

**Fix-R16A-2 — `get_HPO_ID_by_phenotype` could return a non-HPO term as its top hit (HIGH)**
Monarch's search API has no server-side namespace filter (confirmed live: a `prefix` query param is silently ignored), and its `category` filter (`biolink:PhenotypicFeature`) matches equivalent terms across multiple ontologies — for `"seizures"`, the top-ranked result was `MP:0002064` (Mouse Phenotype ontology), not an HPO term, despite the tool's name promising HPO IDs. Added an opt-in `result_id_prefix` config key to the shared `MonarchTool` class, which client-side filters results by ID prefix — opt-in specifically so it doesn't affect other Monarch-type tools (e.g. gene/disease search) that don't declare it. Also over-fetches before filtering so the returned count still matches what the caller requested.
`src/tooluniverse/restful_tool.py`, `src/tooluniverse/data/monarch_tools.json`

**Fix-R16C-1 — `MobiDB_get_protein`/`MobiDB_get_consensus` leaked a raw JSON-decode exception (MEDIUM)**
MobiDB returns HTTP 200 with a completely empty body for an unrecognized identifier (confirmed live, e.g. a UniProt entry name like `"SNCA_HUMAN"` instead of an accession) rather than a 404 — so `raise_for_status()` never fired, and the old code's direct `response.json()` call let a raw `json.JSONDecodeError` ("Expecting value: line 1 column 1") leak through as an opaque "Unexpected error." Added a shared fetch helper that detects the empty body and returns a clean, actionable error instead.
`src/tooluniverse/mobidb_tool.py`

- **R16A-1 / R16A-3 / R16A-4 / R16A-5** — Several HPO/Monarch tools reject a natural-first-guess param name (`phenotype`→`query`, `hpo_id`→`term_id`, `gene_symbol`→`query`/`subject`) with an already-actionable jsonschema error. Naming-intuition mismatch, matches the established defer pattern from many earlier rounds.
- **R16D-1** — `GtoPdb_search_targets` doesn't match the standard HGNC gene symbol (`"ADRB2"`) or a multi-word phrase, and the first-failure hint misleadingly suggests trying the ligands tool instead. Worth a future look, but out of scope for this round's budget.
- **R16D-2** — `GtoPdb_get_disease_associations` requires `disease_id`, not the `targetId` a sibling tool (`get_interactions`) accepts — error is already actionable (names the missing param), just doesn't cross-reference the sibling tool.
- **R16E-2** — `HLALigandAtlas_get_benign_peptides`'s metadata reports `rows_scanned` (the full unfiltered table size) but not the actual filtered-match count, making it hard to gauge result completeness. Metadata-completeness UX gap, not incorrect data.
- **R16B-3 / R16B-4** — `nhanes_get_dataset_info` returns a generic category landing-page URL rather than a per-file listing (the sibling `nhanes_search_datasets` already does this better); `nhanes_search_datasets` with a plain-English term like `"obesity"` returns zero hits since it only matches literal CDC variable-name/label text. Both are real UX rough edges but not incorrect-data bugs.

### Testing

17 new unit tests across 4 files (all mocked, no network calls):
- `tests/unit/test_pubmed_date_range_filter.py` (4)
- `tests/unit/test_nhanes_dataset_resolution.py` (6)
- `tests/unit/test_monarch_result_id_prefix.py` (4)
- `tests/unit/test_mobidb_empty_response.py` (3)

Also re-checked existing PubMed test coverage (`tests/tools/test_pubmed_search_include_abstract.py`) since the date-filter fix touches shared, heavily-used code — 2 failures there are pre-existing and unrelated (confirmed via `git stash` to fail identically against a clean `origin/main` baseline before this round's changes).

Every fix was also verified live against the real upstream API via the CLI before and after the change (see fix descriptions above for the specific before/after evidence, including multiple isolating raw-curl comparisons for the PubMed date-range bug).

### DIFF
--- a/src/tooluniverse/data/monarch_tools.json
+++ b/src/tooluniverse/data/monarch_tools.json
@@ -105,7 +105,7 @@
   },
   {
     "name": "get_HPO_ID_by_phenotype",
-    "description": "Retrieve the HPO ID of a phenotype or symptom.",
+    "description": "Retrieve the HPO ID of a phenotype or symptom. Results are restricted to the HP: namespace (Monarch's underlying search spans multiple ontologies -- e.g. MP: mouse phenotype, UPHENO: cross-species upper ontology -- for the same query term).",
     "parameter": {
       "type": "object",
       "properties": {
@@ -134,6 +134,7 @@
       "limit": 20,
       "offset": 0
     },
+    "result_id_prefix": "HP:",
     "label": [
       "Monarch",
       "Phenotype",
--- a/src/tooluniverse/mobidb_tool.py
+++ b/src/tooluniverse/mobidb_tool.py
@@ -12,7 +12,7 @@ No authentication required.
 """
 
 import requests
-from typing import Dict, Any
+from typing import Dict, Any, Optional
 from .base_tool import BaseTool
 from .tool_registry import register_tool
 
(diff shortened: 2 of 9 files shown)


gasvn:

PR #312, opened 2026-07-18
Fix Round 17: TheraSAbDab target aliasing, SAbDab SPA-migration detection, Rfam operation fallback

gasvn:
## Summary

Round 17 of the ongoing test-harness sweep: 5 role-play researcher persona agents targeted antibody-structure, ncRNA, and comparative-genomics domains not recently covered. Every reported issue was CLI-verified against live upstream APIs before fixing; only root-caused, reproducible bugs are fixed here (naming-intuition mismatches, ranking-quality gaps, and other upstream-only issues were reviewed and deliberately deferred).

- **Fix-R17D-1** — `TheraSAbDab_search_by_target` matched targets by plain substring containment against Rfam-style `"/"`- and `";"`-joined alias strings (e.g. `"PDCD1/CD279/PD1"`). Confirmed live that a search for `"PD-1"` matched `"pd1"` inside `"entpd1"` (ENTPD1/CD39, an unrelated target), silently polluting PD-1 results with CD39 antibodies. Now splits the stored target string into individual alias tokens and requires an exact match against one of them.
- **Fix-R17D-2/3** — SAbDab has migrated to a new React SPA ("SAbDab2"); its old structure-download and summary URLs now return an HTTP-200 HTML shell instead of real PDB/TSV data (confirmed via `curl -sI`: `/api/pdb/{id}/` 307-redirects to an internal-only, publicly-unreachable `sabdab-backend:8000` host). Before this fix, `SAbDab_get_structure` silently returned the HTML as if it were a successful PDB download, and `SAbDab_get_structure_summary` blamed the HTML on "structure may not be an antibody complex" — both misleading. Both now detect the SPA shell and report the real upstream cause honestly.
- **Fix-R17C-2** — Every `Rfam_*` tool config constrains `operation` to a single-value enum (each tool name maps to exactly one Rfam operation), yet `operation` was still marked `required`, forcing callers to look up and pass back a value the config already pins. `RfamTool.run()` now falls back to the sole enum value when `operation` is omitted, and the 9 Rfam tool configs no longer list it in `required`.

## Test plan

- [x] 11 new mocked unit tests (`tests/unit/test_therasabdab_target_matching.py`, `tests/unit/test_sabdab_spa_migration_detection.py`, `tests/unit/test_rfam_operation_fallback.py`) covering the fixed behavior plus adjacent non-regression cases (e.g. CD39/HER2 searches still work, real PDB/TSV content still parses, multi-value enums correctly still require `operation`).
- [x] Pre-existing `tests/tools/test_sabdab_tool.py` and `tests/tools/test_rfam_tool.py` pass unchanged.
- [x] `code-simplifier` pass applied (deduped alias-set lookup in TheraSAbDab, extracted shared `_looks_like_html` helper in SAbDab).
- [x] All three fixes independently verified live against the real upstream APIs before and after the change.

### DIFF
--- a/src/tooluniverse/data/rfam_tools.json
+++ b/src/tooluniverse/data/rfam_tools.json
@@ -22,7 +22,7 @@
           "description": "Output format"
         }
       },
-      "required": ["operation", "family_id"]
+      "required": ["family_id"]
     },
     "return_schema": {
       "type": "object",
@@ -70,7 +70,7 @@
           "description": "Compress output with gzip"
         }
       },
-      "required": ["operation", "family_id"]
+      "required": ["family_id"]
     },
     "return_schema": {
       "type": "object",
@@ -108,7 +108,7 @@
           "description": "Required: Rfam accession or family name"
         }
       },
-      "required": ["operation", "family_id"]
+      "required": ["family_id"]
     },
     "return_schema": {
       "type": "object",
@@ -151,7 +151,7 @@
           "description": "Maximum time to wait for results (default: 120s). Job continues running if timeout reached."
         }
       },
-      "required": ["operation", "sequence"]
+      "required": ["sequence"]
     },
     "return_schema": {
       "type": "object",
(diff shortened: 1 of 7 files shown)


gasvn:

PR #314, opened 2026-07-19
Fix Round 18: PDBe/EMDB, Reactome, CTD, IEDB, and KEGG bugs

gasvn:
## Summary

Round 18 of the ongoing test-harness sweep: 5 role-play researcher personas exercised structural biology (PDBe/EMDB), pathway/network biology (Reactome), toxicology (CTD), immunology (IEDB), and pharmacogenomics (KEGG) — domains not previously covered. 27 findings were CLI-verified against live upstream APIs; 12 confirmed root-caused and fixed here (the rest were deliberately deferred: genuine coverage gaps, upstream data-model nuances, or STRING/IEDB behavior inherited from those APIs — not ToolUniverse bugs).

## Fixed

- **Fix-R18A-1** — `EMDB_search_structures`'s `rows` parameter had no matching `{placeholder}` in its endpoint template and was silently dropped (confirmed live: `rows=2` and `rows=10` both returned the same 10 results). Unmatched args are now passed through as query params.
- **Fix-R18A-2 / Fix-R18C-6** — `tu run`'s human-readable error renderer matched a bare `"not found"` substring to decide whether to show "check tool name spelling" tips, so a tool's own HTTP-404-shaped error (e.g. a bad PDB ID, or CTD's `"'cadmium' was not found in the RENCI CTD mirror"`) produced misleading tool-discovery guidance for what was actually a bad parameter *value*. Switched to the structured `error_details.type == "ToolUnavailableError"` signal, which reliably discriminates a genuine unknown tool name.
- **Fix-R18A-3** — PDBe SIFTS 404s (e.g. `PDBeSIFTS_get_scop_mapping`) now clarify that the entry likely just lacks data of that specific mapping type, instead of a bare, ambiguous status code.
- **Fix-R18B-1** — Reactome's `containedEvents` endpoint mixes full event dicts with bare integer DB IDs for some sub-pathways (confirmed live for R-HSA-2219528: 2 of 3 real sub-pathways came back as bare ints). These were silently dropped instead of resolved, both losing real hierarchy data and making `total_events` disagree with `pathway_count + reaction_count`. Bare IDs are now batch-resolved via one extra call to Reactome's `/data/query/ids`.
- **Fix-R18B-2** — `ReactomeInteractors_get_protein_interactors` sent `page=-1` (a Reactome-specific "ignore pagination, return everything" signal), making its own documented `page_size` parameter have zero effect (confirmed live: returned all 153 interactors regardless of `page_size=10`).
- **Fix-R18C-1 / Fix-R18D-3** — two more actionable fields silently dropped by the CLI's error renderer: a top-level `"suggestion"` string some tools use for a single redirect hint (e.g. `CTD_get_gene_diseases` → OpenTargets), and a top-level `"detail"` dict some `BaseRESTTool`-backed tools use for PostgREST-style upstream error detail (e.g. IEDB's shared query tool).
- **Fix-R18C-2/4** — CTD's RENCI backend only indexes CURIE-prefixed IDs (`MESH:`, `CAS:`, `NCBIGene:`) in `equivalent_identifiers`, but the tool's own parameter docs promise bare CAS RN / MeSH ID / NCBI Gene ID work (confirmed live: `"D000082"`, `"80-05-7"`, `"7157"` all failed bare, succeeded prefixed). Recognizable bare ID formats now try the correctly-prefixed CURIE first, falling back to the original string so name-based and already-prefixed lookups are unaffected.
- **Fix-R18D-1/2** — IEDB's MHC-II prediction endpoint returns a plain-text validation error (not TSV) with HTTP 200 for invalid input (e.g. a too-short sequence), which was silently misparsed by `csv.DictReader` as a fake successful prediction with a fabricated rank. Separately, even genuine successful responses were affected: the real TSV column is named `"rank"`, not `"percentile_rank"` (unlike the sibling MHC-I endpoint, which really does use that name), so every real MHC-II result silently reported a hardcoded rank of 100 regardless of the true binding rank.
- **Fix-R18E-1/2** — `KEGG_get_drug`'s `PATHWAY`/`DISEASE` fields are nested 2 spaces under `TARGET`/`EFFICACY` in KEGG's flat-file format (not top-level like every other field this parser recognized), and `KEGG_get_disease` uses the field name `DIS_PATHWAY` (not `PATHWAY`) — both silently unrecognized, dropping all pathway/disease associations (confirmed live: 3 real pathways and 7 real diseases lost for imatinib/D01441; 1 real pathway lost for CML/H00004).
- **Fix-R18E-4** — `KEGG_get_brite_hierarchy` 404s uninformatively for KEGG's "table-type" BRITE files, which have no JSON tree representation upstream even though they're real, listed hierarchies (confirmed live for `br08341`, Pharmacogenomic biomarkers).
- **Fix-R18E-5** — `KEGG_get_variant` silently dropped `COSF` (COSMIC Fusion) and `OmimVar` cross-reference lines — the parser only recognized `ClinVar`/`dbSNP`/`COSM` keys.
- **Fix-R18E-6/7** — `KEGG_get_network` 404s on a bare `"nt######"`-style network ID unless a `"network:"` prefix is prepended (confirmed live), and its child elements are listed under `MEMBER` (not `ELEMENT`) for that same ID family — both silently broke the natural chaining path from sibling tools that surface `nt######` IDs.

Genuine coverage gaps or upstream-API-inherited behavior, not ToolUniverse code bugs: Reactome's enrichment-vs-found-entities count mismatch (AKT1's dual UniProt mapping, a real upstream data-model nuance) and STRING's silent resolution of an ambiguous gene symbol like "AKT"; CTD's lack of a chemical-pathway tool and one disease query returning genuinely sparse results in the June-2024 snapshot; IEDB's organism-name search not matching common pathogen abbreviations (correct `ilike` behavior, just a documentation gap).

## Test plan

- [x] 35 new mocked unit tests across 8 files covering the fixed behavior plus adjacent non-regression cases (e.g. name-based CTD lookups still work, MHC-I prediction unaffected by the shared TSV-parsing change, `N#####`-style KEGG network IDs aren't double-prefixed).
- [x] Pre-existing `test_ctd_tool.py` and `test_iedb_bcell_prediction.py` pass unchanged.
- [x] Full `tests/unit` suite (the standard CI scope) run clean — 0 failures.
- [x] `code-simplifier` pass applied (deduped KEGG variant-record construction into a shared helper, collapsed a 5-branch cross-ref elif-chain into a table-driven loop, and refactored CTD's CURIE-candidate generation into a small rules table).
- [x] All 12 fixes independently verified live against the real upstream APIs before and after the change.

### DIFF
--- a/src/tooluniverse/cli.py
+++ b/src/tooluniverse/cli.py
@@ -479,7 +479,15 @@ def _render_run(d: dict) -> str:
 
     # Feature-25B-02: for "tool not found" errors, replace generic network tips
     # with tool-discovery tips and include fuzzy suggestions when available.
-    is_not_found = "not found" in short_err.lower()
+    # Fix-R18A-2/R18C-6: a plain "not found" substring match also fires on a
+    # tool's own HTTP-404-shaped error message (e.g. "PDBe API error: 404
+    # Client Error: Not Found for url: ..." or CTD's "'cadmium' was not found
+    # in the RENCI CTD mirror") -- confirmed live these produced misleading
+    # "check tool name spelling" tips for a perfectly valid tool call with a
+    # bad parameter VALUE, not a bad tool name. A genuine unknown-tool-name
+    # error is reliably tagged error_details.type == "ToolUnavailableError"
+    # (confirmed live); use that structured signal instead of the message text.
+    is_not_found = details.get("type") == "ToolUnavailableError"
     is_api_key_error = "requires api key" in short_err.lower()
     suggestions = d.get("suggestions") or details.get("suggestions") or []
     if is_api_key_error:
@@ -507,6 +515,24 @@ def _render_run(d: dict) -> str:
         )
         if hint:
             cli_steps = [*cli_steps, hint]
+        # Fix-R18C-1: some tools put a single actionable redirect at the
+        # top-level "suggestion" key (e.g. CTD_get_gene_diseases pointing
+        # callers at OpenTargets) -- confirmed live this was silently
+        # dropped since only the plural "suggestions" (fuzzy tool-name
+        # matches) and error_details.next_steps were ever surfaced.
+        suggestion = d.get("suggestion")
+        if suggestion:
+            cli_steps = [*cli_steps, suggestion]
+        # Fix-R18D-3: BaseRESTTool-backed tools put the actionable upstream
+        # error in a top-level "detail" dict (PostgREST-style hint/message,
+        # e.g. IEDB's shared query tool), distinct from error_details --
+        # confirmed live this was silently dropped, leaving only the generic
+        # "Error: HTTP request failed" with no indication of the real cause.
+        raw_detail = d.get("detail")
+        if isinstance(raw_detail, dict):
+            detail_hint = raw_detail.get("hint") or raw_detail.get("message")
+            if detail_hint:
+                cli_steps = [*cli_steps, detail_hint]
         if cli_steps:
             lines.append("Tips:")
             for step in cli_steps:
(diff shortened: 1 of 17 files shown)


gasvn:

PR #315, opened 2026-07-19
Fix Round 19: DepMap, ClinGen, Ensembl, HPA, STITCH bugs

gasvn:
## Summary

Round 19 of the ongoing test-harness sweep: 5 role-play researcher personas exercised comparative/functional genomics (Ensembl), clinical variant curation (ClinGen), cancer genomics (DepMap/COSMIC/cBioPortal/Cellosaurus), tissue expression (GTEx/HPA), and drug-target pharmacology (DGIdb/DrugCentral/STITCH) — domains not previously covered. 27 findings were CLI-verified against live upstream APIs; 14 confirmed root-caused and fixed here (the rest deferred: genuine coverage gaps needing bigger redesigns, e.g. cBioPortal's clinical-attribute filter requiring a fundamentally different POST-based multi-step request pattern, or Cellosaurus's `fields` parameter needing a full short-code-to-JSON-key mapping table).

## Fixed

- **Fix-R19C-1/3** — `DepMap_search_genes`/`DepMap_search_cell_lines` scanned only the first 5 pages of a ~45,751-gene / ~2,266-model catalog (the Sanger Cell Model Passports API's `filter[gene]`/`filter[model]` params silently don't filter — confirmed live), so any gene or cell line not alphabetically very early (KRAS, EGFR, TP53, A549) was always reported "not found". Both now binary-search the full, sorted catalog. `DepMap_get_cell_line` read field names (`tissue`, `cancer_type`, `gender`, `ethnicity`, `msi_status`) that don't exist on the model resource at all — the real data lives on related sample/tissue/cancer_type/patient resources, now resolved in one request via the JSON:API `include` param. `DepMap_get_cell_lines` fixes `model_name` (cheap, no extra request) and now warns explicitly when a tissue/cancer_type filter has no effect server-side, instead of silently returning unfiltered results as if they'd been filtered.
- **Fix-R19B-1/2** — ClinGen's clinical actionability API returns a JSON table (`{"columns": [...], "rows": [[...]]}`), not a list of dicts. The gene filter's `isinstance(curations, list)` guard was always `False` against this shape — silently skipping the filter entirely in `ClinGen_get_actionability_adult`/`_pediatric` (returning all 254 unfiltered rows for BRCA1) and silently zeroing out results in `ClinGen_search_actionability` (returning nothing, even though BRCA1 has 6 real curated Adult entries).
- **Fix-R19B-3** — `ClinGen_get_variant_classifications` downloaded the entire, unpaginated ClinGen Evidence Repository (confirmed live: 28+MB, still streaming after 200s) and filtered client-side. Now requires a `gene` argument and uses a fast (~1s), already gene-filtered JSON endpoint instead.
- **Fix-R19A-1** — `Ensembl_get_cross_references` silently truncated `xrefs` to 100 while `database_summary` was computed from the full, untruncated list (confirmed live for TP53: 157 real xrefs), so the per-database counts didn't match what was actually returned. Cap raised to 500 (effectively never triggers) with an explicit `truncated` flag.
- **Fix-R19A-2** — `EnsemblArchive_get_id_history` blanket-mapped every upstream HTTP 400 to "Invalid ID format", discarding the real message — confirmed live that Ensembl returns 400 (not 404) for a syntactically valid but nonexistent ID too, with a real, more useful error in the response body.
- **Fix-R19D-1** — `HPA_search_genes_by_query` has no working server-side result limit (confirmed live: a `limit` query param has no effect) — a short, valid gene symbol like "INS" returned 8,441 loosely-matched genes. Exact gene-symbol matches now rank first and the response is capped client-side (default 50, overridable via a new `max_results` argument).
- **Fix-R19E-3** — `STITCHTool` pointed at `string-db.org`, STRING's own protein-only network API — confirmed live it silently returned real but entirely protein-protein data for chemical queries (e.g. resolving "aspirin" returned the gene SLC17A4). Repointed at STITCH's real current domain (`stitch-db.org`, reached via a 301 redirect from the historical `stitch.embl.de`), which correctly resolves chemical identifiers (aspirin → `CID100002244`). The network/interactions sub-endpoints still 404 on the new domain even for a correctly-resolved identifier (confirmed live); both now give an honest error instead of implying the identifier itself is the problem.
- **Docs fix** — two GTEx tool descriptions inaccurately claimed a "GTEx V11 (January 2026)" dataset; corrected to reflect the real default (V8, 2020) and the actual queryable `dataset_id` values (v7/v8/v10/kids_first_harmonization — no V11 exists via this API).

Genuine coverage gaps or bigger-scope redesigns, not quick fixes: cBioPortal's `clinical_attribute_id` filter requires a fundamentally different POST-based `/clinical-data/fetch` request (needs a sample-ID list first, not a simple GET filter — confirmed live); Cellosaurus's `fields` sub-parameter needs a full short-code-to-JSON-key mapping table across ~60 field tags that I couldn't fully validate; Cellosaurus's `query_converter` garbling natural-language queries into malformed Solr syntax (NLP-translation bug, bigger scope); DGIdb's `get_gene_info` returning output identical to `get_gene_druggability` (upstream GraphQL query is missing fields, low-risk one-line fix candidate for a future round); a few LOW-severity documentation/consistency gaps (HPA envelope-shape inconsistency, matches the long-documented systemic issue from issue #288; ClinGen sibling-tool field-null inconsistency; COSMIC gene-level mutation ordering, not independently confirmed as a code bug).

## Test plan

- [x] 26 new mocked unit tests across 5 files covering the fixed behavior plus adjacent non-regression cases (e.g. `N#####`-style DepMap network IDs aren't affected by the binary search changes, MHC-I-equivalent regression checks aren't needed here but sibling-tool non-regression is verified for each fix, name-based lookups that already worked keep working).
- [x] Pre-existing GTEx/HPA test suites (`test_gtex_v2_variant_only.py`, `test_gtex_v2_tool.py`, `test_hpa_tool.py`) pass unchanged.
- [x] Full `tests/unit` suite (the standard CI scope) run clean — 0 failures.
- [x] `code-simplifier` pass applied — extracted a shared binary-search helper (`_find_catalog_page`) used by both `DepMap_search_genes` and `DepMap_search_cell_lines`, a shared 404-error builder for STITCH's two broken endpoints, and a shared gene-filter helper for ClinGen's two actionability methods.
- [x] All 14 fixes independently verified live against the real upstream APIs before and after the change.

### DIFF
--- a/src/tooluniverse/clingen_tool.py
+++ b/src/tooluniverse/clingen_tool.py
@@ -27,6 +27,38 @@ ACTIONABILITY_PEDIATRIC_URL = (
 EREPO_BASE_URL = "https://erepo.clinicalgenome.org/evrepo/api"
 
 
+def _actionability_rows_to_dicts(data: Any) -> List[Dict[str, Any]]:
+    """Fix-R19B-1/2: the actionability API's `?flavor=flat` response is a
+    JSON table -- {"columns": [...26 names...], "rows": [[...26 values...],
+    ...]} -- confirmed live, not a list of per-curation dicts. The gene
+    filter's `isinstance(curations, list)` guard was always False against
+    this dict, silently skipping the filter entirely in
+    _get_actionability_adult/_pediatric (returning everything, unfiltered)
+    and silently zeroing out `matches` in _search_actionability (returning
+    nothing). Zip columns with each row into real dicts so both a
+    downstream isinstance(list) check and dict-style field access work.
+    """
+    if isinstance(data, list):
+        return data
+    columns = data.get("columns")
+    rows = data.get("rows")
+    if not isinstance(columns, list) or not isinstance(rows, list):
+        return []
+    return [dict(zip(columns, row)) for row in rows]
+
+
+def _filter_curations_by_gene(
+    curations: List[Dict[str, Any]], gene: str
+) -> List[Dict[str, Any]]:
+    """Keep only actionability curations whose `geneOrVariant` field contains
+    `gene` (case-insensitive substring; the field holds comma-separated gene
+    lists, e.g. "BRCA1,BRCA2")."""
+    gene_upper = gene.upper()
+    return [
+        c for c in curations if gene_upper in str(c.get("geneOrVariant", "")).upper()
+    ]
+
+
 @register_tool("ClinGenTool")
 class ClinGenTool(BaseTool):
     """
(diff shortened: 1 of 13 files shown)


gasvn:

PR #316, opened 2026-07-19
Fix Round 20: Rhea reaction participants, UniProt/Orphanet/Pharos data-completeness bugs

gasvn:
## Summary

Round 20 of the ongoing test-harness sweep: 5 role-play researcher personas exercised protein biology (UniProt), structural biology (AlphaFold/SASBDB), proteomics (MassIVE/Expression Atlas/PeptideAtlas/BioSamples), rare disease genetics (Orphanet), and signal transduction/pharmacology (Pharos/SIGNOR/Rhea) — domains not previously covered. 18 findings were CLI-verified against live upstream APIs; 12 confirmed root-caused and fixed here.

## Fixed

- **Fix-R20E-1** — `Rhea_get_reaction`/`Rhea_get_reaction_participants` silently dropped generic/polymer ChEBI "residue" participants (e.g. the protein-tyrosine substrate of a kinase reaction, confirmed live for RHEA:10596) because Rhea identifies them with a `rhea-comp:` id in the htmlequation instead of `chebi:`, which the parser's regex only matched for the latter. Now captured with `is_generic`/`rhea_comp_id` fields instead of vanishing entirely.
- **Fix-R20A-1** — `UniProtIDMap_convert_ids`/`_to_pdb`/`gene_to_uniprot` silently truncated large result sets to 25 records (confirmed live: EGFR→PDB has 354 real xrefs, not 25) — the polling loop treated a status-endpoint redirect's small default-page-size preview as the final answer instead of always fetching the full, paginated result set via `size=500` + Link-header pagination.
- **Fix-R20D-1** — 5 of 7 code-based Orphanet operations (`get_genes`, `get_epidemiology`, `get_natural_history`, `get_classification`, `get_icd_mapping`) silently returned `status:success` with empty data for a nonexistent ORPHA code, instead of erroring like their siblings `get_disease`/`get_phenotypes` already did — a materially misleading result in a clinical-genetics context.
- **Fix-R20D-2** — `Orphanet_get_classification`'s own description promises "parent and child disease categories," but the implementation only ever returned a flat list of named classification *systems* the disease is filed under (membership, not hierarchy). Now fetches the real parent/children tree via the RDcode API's `PreferentialParent`/`PreferentialChildren` endpoints (confirmed live: Cystic fibrosis's parent is "Rare respiratory disease," which itself has 82 real children), keeping the systems-membership list too (renamed `member_of_classifications`).
- **Fix-R20E-2** — `Pharos_search_targets`'s `tdl` filter was accepted by the schema but never read by the code — confirmed live that `tdl:"Tdark"` and `tdl:"Tclin"` returned byte-identical results. Pharos' GraphQL `IFilter` has no direct `tdl` field; fixed by expressing it through Pharos' generic `facets` mechanism (confirmed via GraphQL introspection).
- **Fix-R20E-3** — `Pharos_get_tdl_summary` returned only static TDL labels/descriptions plus a note telling the caller to look elsewhere, despite its own description promising real counts. Now returns real whole-proteome TDL counts (Tclin: 704, Tchem: 1904, Tbio: 12303, Tdark: 5501) from a single minimal `facets` query.
- **Fix-R20A-3** — UniProt's JSONPath-extraction tools (`get_organism_by_accession`, `get_sequence_by_accession`, `get_function_by_accession`, `get_subcellular_location_by_accession`) surfaced a cryptic raw `"No data found for JSONPath: ..."` error for a real but inactive/deleted accession (confirmed live for Q9ZZZ9) instead of explaining why — unlike `get_entry_by_accession`, which already handles the same case gracefully.
- **Fix-R20B-1** — `alphafold_get_summary`'s `metadata.count` was hardcoded to 1 whenever the API returned a dict rather than a top-level list — silently wrong for `/uniprot/summary`, which wraps its real payload as `{"uniprot_entry": ..., "structures": [...]}` (confirmed live: P69905 has 16 real structures, reported as 1).
- **Fix-R20B-2** — CLI default (non-`--json`) error output dropped a tool's `detail` field whenever it was a JSON-encoded string rather than a dict with `hint`/`message` keys (confirmed live for SASBDB 404s and AlphaFold 400s), hiding the actually-useful upstream message behind `--json`.
- **Fix-R20C-1/2** — `ExpressionAtlas`'s `gene` parameter across `get_baseline_expression`/`search_differential_experiments`/`search_experiments` had no real filtering effect (confirmed live: a real gene and a nonsense gene string returned byte-identical results) — it relies on a literal-text match against EBI Search's description index, which almost never contains individual gene symbols. Rather than a full re-architecture (would need per-gene expression-value lookups the API doesn't cheaply expose), every gene-querying response now surfaces an explicit `warning` instead of silently implying the unfiltered catalog is gene-specific.
- **Fix-R20C-4** — MassIVE's ProXI error handler discarded the upstream JSON error body's actual reason (confirmed live: `{"message":"No data found for the specified parameters."}`), showing only a generic `"404 Client Error: ..."` string.
- **Fix-R20B-3** — `SASBDB_search`'s free-text fallback (correctly self-documents that SASBDB has no full-text search) returned all ~5432 published entries unbounded (~350KB) with no way to cap it. Added a `max_results` argument (default 50) with `total_entries`/`truncated` fields.

- UniProt's two sibling proteome-count tools (`UniProt_get_proteome` vs `UniProtRef_get_proteome`) disagree by 14 proteins (147506 vs 147520) — likely a snapshot-timing or counting-method difference between two independently-maintained endpoints, not a client-side bug.
- `ExpressionAtlas_get_experiment`'s promised `factors`/`technology`/`contrasts`/`assay_count`/`pubmed_ids` fields are always empty (confirmed live for the tool's own documented test example) because the upstream GXA `/json/experiments/{accession}` endpoint's `experiment` sub-object genuinely doesn't carry those keys — no richer endpoint found after probing several candidates. Tool description corrected to stop overpromising.
- `Pharos_get_target`'s `status:"success"` + `data:null` combination on a no-match is a minor envelope-consistency nit (arguably valid REST semantics: request succeeded, no match found), not incorrect behavior — left as-is.
- Feature-20D-003 (misleading CLI "check tool name spelling" tips on an Orphanet not-found error) is a duplicate of the already-fixed-but-not-yet-merged Fix-R18A-2/R18C-6 from round 18's PR — not re-fixed here to avoid redundant work.

## Test plan

- [x] 26 new mocked unit tests across 10 files covering every fix plus adjacent non-regression cases (e.g. small/ordinary ChEBI participants still parse correctly, active UniProt accessions still extract fine, condition-only Expression Atlas queries are unaffected by the gene-filter warning, network-error resilience in the new Orphanet existence-check helper).
- [x] 2 existing test files strengthened: `tests/tools/test_pharos_tool.py` (fixed a vacuous assertion checking a nonexistent `idgTDL` key instead of the real `tdl` key, which let the tdl-filter bug ship unnoticed; also strengthened the tdl_summary test to check real counts), `tests/tools/test_tu_cli.py` (added coverage for the new detail-string CLI surfacing).
- [x] Full non-live-API suite (2568 tests) passes clean; the only failures are pre-existing live-API flakiness in files this PR never touches (BindingDB, ComplexPortal, DailyMed) and one pre-existing embedding-search relevance test, both independently confirmed to fail identically on unmodified `main`.
- [x] `code-simplifier` pass applied — extracted a shared `_orpha_not_found_error` helper deduplicating 4 repeated existence-check call sites in `orphanet_tool.py`; everything else in the diff was already clean.
- [x] All 12 fixes independently verified live against the real upstream APIs before and after the change.

### DIFF
--- a/src/tooluniverse/alphafold_tool.py
+++ b/src/tooluniverse/alphafold_tool.py
@@ -123,6 +123,25 @@ class AlphaFoldRESTTool(BaseTool):
 
         return {"response": resp}
 
+    @staticmethod
+    def _response_count(data: Any) -> int:
+        """Best-effort result count across this tool's many differently-
+        shaped AlphaFold endpoints. Several (e.g. /uniprot/summary, which
+        wraps its real payload as {"uniprot_entry": ..., "structures":
+        [...]}, or /annotations, which wraps it as {..., "annotation":
+        [...]}) return a dict whose only list-valued field is the actual
+        result list -- report that list's length instead of a hardcoded 1,
+        which was always wrong for these shapes (confirmed live: P69905's
+        /uniprot/summary carries 16 structures, not the 1 the old code
+        reported)."""
+        if isinstance(data, list):
+            return len(data)
+        if isinstance(data, dict):
+            for value in data.values():
+                if isinstance(value, list):
+                    return len(value)
+        return 1
+
     def run(self, arguments: Dict[str, Any]):
         """Execute the tool with provided arguments."""
         # Normalize uniprot_id / uniprot_accession → qualifier
@@ -172,7 +191,7 @@ class AlphaFoldRESTTool(BaseTool):
                     "status": "success",
                     "data": data,
                     "metadata": {
-                        "count": len(data) if isinstance(data, list) else 1,
+                        "count": self._response_count(data),
                         "source": "AlphaFold Protein Structure DB",
                         "endpoint": url,
                         "query": arguments,
--- a/src/tooluniverse/cli.py
+++ b/src/tooluniverse/cli.py
@@ -453,6 +453,38 @@ def _render_info(d: dict) -> str:
     return "\n".join(lines)
 
 
+def _extract_detail_hint(raw_detail: Any) -> str | None:
+    """Pull a human-readable hint out of an error envelope's `detail` field.
+
+    Different tools populate `detail` inconsistently: a dict with hint/
+    message keys, or (very commonly, since many tools just pass through
+    `resp.text`) a raw response-body string that is often itself a
+    JSON-encoded object, e.g. '{"code": "404", "status": "The Uniprot code
+    p00698 does not exist in the SASBDB"}'. The default (non---json) CLI
+    output previously dropped this entirely, showing only the generic
+    top-level error string and leaving the actually-useful upstream
+    message discoverable only via --json (confirmed live for SASBDB 404s
+    and AlphaFold 400s).
+    """
+    if isinstance(raw_detail, dict):
+        return raw_detail.get("hint") or raw_detail.get("message")
+    if isinstance(raw_detail, str) and raw_detail.strip():
+        try:
+            parsed = json.loads(raw_detail)
+        except (json.JSONDecodeError, TypeError):
+            parsed = None
+        if isinstance(parsed, dict):
+            for key in ("message", "error", "status", "detail", "reason"):
+                val = parsed.get(key)
+                if isinstance(val, str) and val.strip():
+                    return val
+        # Not JSON, or no usable field inside it -- fall back to the raw
+        # string, capped so a huge HTML error page doesn't flood the
+        # terminal.
+        return raw_detail if len(raw_detail) <= 300 else raw_detail[:297] + "..."
+    return None
+
+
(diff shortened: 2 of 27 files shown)


gasvn:

PR #317, opened 2026-07-19
Fix Round 21: HTTP-status truthiness bug, BVBRC/IPD/VDJDB/GWAS data-completeness bugs

gasvn:
## Summary

Round 21 of the ongoing test-harness sweep: 5 role-play researcher personas exercised toxicology/pharmacovigilance (PubChemTox/FAERS), immunogenetics (IPD-IMGT/HLA, VDJDB, HLA Ligand Atlas, MHC Motif Atlas), microbiology (BV-BRC, Pathoplexus), neuroscience (NeuroMorpho, Allen Brain Atlas, NeuroVault), and population genetics (gnomAD, PGS Catalog, GWAS Catalog/SumStats) — domains not previously covered. 14 findings were CLI-verified against live upstream APIs; 9 confirmed root-caused and fixed here.

## Fixed

- **Fix-R21D-1/2** — `NeuroMorphoTool` and `AllenBrainTool`'s HTTPError handlers always reported "unknown" instead of the real HTTP status code, because `e.response else "unknown"` relies on `requests.Response.__bool__` (which is `False` for every 4xx/5xx response) instead of an explicit `is not None` check — the exact opposite of the intended fallback. Confirmed live for `NeuroMorpho_get_neuron` on a nonexistent id (real status: 404, reported: "unknown"). Fixed both; NeuroMorpho's handler also now surfaces the upstream JSON body's `message` field (e.g. "Requested neuron not found").
- **Fix-R21C-1** — `BVBRC_search_pathways`'s `select_fields` list omitted `gene`/`feature_id`/`product`/`annotation`, so genuinely distinct pathway records (different genes, or the same EC/pathway independently annotated by RefSeq vs. PATRIC) collapsed into what looked like exact duplicate rows (confirmed live: 10/10 identical rows for *M. tuberculosis* "Fatty acid metabolism"; direct API introspection showed the underlying records actually differ by gene: fadD15, fadD19, fadD35, etc.).
- **Fix-R21B-3** — `MHCMotifAtlas_get_allele_ligands` matched allele codes via exact, case-sensitive string equality against the atlas's all-uppercase TSV files, so a well-formed but lowercase allele (e.g. `"a0201"`) silently matched zero peptides.
- **Fix-R21A-2** — `PubChemToxTool` conflated two very different failure modes into the identical `"This heading may not exist for this compound"` message: a compound name that never resolved to a CID at all, and a real, resolved compound simply missing one specific toxicity heading. Confirmed live that PubChem's own name→CID endpoint itself 404s for unresolvable names (not just an empty CID list) — now distinguished with a clear, distinct error for each case.
- **Fix-R21B-1** — `IPD_search_hla_alleles` silently returned zero results for an `"HLA-"`-prefixed allele name — exactly the format sibling tool VDJDB itself returns (`mhc.a: "HLA-A*02:01"`) — since the underlying `name` field doesn't carry that locus prefix. Now strips it before searching.
- **Fix-R21B-2** — `VDJDBTool`'s `species` filter is an exact-match filter against exactly 3 CamelCase values (`HomoSapiens`/`MusMusculus`/`MacacaMulatta`) but was typed as free text, so a plausible alias like `"human"` silently matched zero of 20,139 real records for the same epitope (confirmed live). Added alias normalization shared by both `search_cdr3` and `get_antigen_specificity`.
- **Fix-R21C-2** — Pathoplexus's 4 HTTPError handlers said "check the organism slug" without ever naming the actual valid slugs, forcing a trip back to the tool description. Now lists them directly in the error message.
- **Fix-R21E-2** — `GWASSumStats_get_trait_studies`'s 404 for an unrecognized `trait_id` gave no hint that the API only accepts EFO ontology ids — a natural trap since a sibling database (PGS Catalog) returns MONDO ids for the same diseases (confirmed live: `MONDO_0004975` 404s identically to a genuinely nonexistent EFO id, while `EFO_0000249` for the same disease works). Now names the likely cause when `trait_id` isn't EFO-shaped.
- **Docs fix** — `NeuroVault_list_collections`'s description promised the ability to "find specific studies," but the public API has no working keyword/search/filter parameter at all (confirmed live against 5 candidate param names — `name`, `name__icontains`, `q`, `search`, `keywords` — all silent no-ops returning the full unfiltered 17,588-collection catalog). Corrected to stop overpromising.

- FAERS/FDA free-text drug-name search pulls in combination products (e.g. searching "acetaminophen" surfaces opioid-acetaminophen combos), which skews aggregate adverse-event counts — a data-modeling tradeoff inherent to openFDA's free-text matching, not a clear client-side bug; fixing it well would need a fundamentally different exact-ingredient-match query path.
- FAERS silently returning an empty result for both "drug not recognized" and "drug recognized, zero events" is a low-value fix needing an extra existence-check request per call, with uncertain benefit given FAERS' free-text nature.
- A GWAS Summary Statistics region query returning duplicate records is confirmed (via direct curl against the raw EBI API) to originate from the upstream API itself, not from this package.
- Feature-21E-001 (CLI's generic "check tool name" tips firing on data-level "not found" errors, e.g. a mistyped gene symbol) is a duplicate of the already-fixed-but-not-yet-merged Fix-R18A-2/R18C-6 from round 18's PR (#314) — not re-fixed here to avoid redundant/conflicting work.

## Test plan

- [x] 24 new mocked unit tests across 7 new files, plus 2 tests added to the existing IPD-IMGT/HLA test file covering the new HLA-prefix-stripping behavior.
- [x] Full `tests/unit` suite (the standard, fully-mocked CI scope) passes clean — 0 failures.
- [x] `code-simplifier` pass applied — consolidated Pathoplexus's 4 near-identical HTTPError handlers into one shared `_http_error_response()` helper (carefully preserving the module-level `_ORGS` string constant, required since this project targets Python ≥3.10 and nested identical quotes inside an f-string only became legal in 3.12/PEP 701); fixed one line-length formatting issue. Everything else in the diff was already clean.
- [x] All 9 fixes independently verified live against the real upstream APIs before and after the change.

### DIFF
--- a/src/tooluniverse/allen_brain_tool.py
+++ b/src/tooluniverse/allen_brain_tool.py
@@ -55,7 +55,12 @@ class AllenBrainTool(BaseTool):
                 "error": "Failed to connect to Allen Brain Atlas API.",
             }
         except requests.exceptions.HTTPError as e:
-            status = e.response.status_code if e.response else "unknown"
+            # `bool(e.response)` is `requests.Response.ok`, which is False for
+            # every 4xx/5xx response -- so `e.response else "unknown"` always
+            # discarded the real status code on an actual HTTP error (same bug
+            # confirmed live in the sibling NeuroMorpho tool). Use an explicit
+            # None check instead.
+            status = e.response.status_code if e.response is not None else "unknown"
             return {
                 "status": "error",
                 "error": f"Allen Brain Atlas API HTTP error: {status}",
--- a/src/tooluniverse/bvbrc_tool.py
+++ b/src/tooluniverse/bvbrc_tool.py
@@ -663,6 +663,18 @@ class BVBRCTool(BaseTool):
             "ec_number",
             "ec_description",
             "taxon_id",
+            # BV-BRC's pathway index has one row per gene/annotation-source
+            # per pathway/EC pair (e.g. the same EC number annotated once
+            # by RefSeq and once by PATRIC, or matched by multiple distinct
+            # genes). Without these fields selected, those genuinely
+            # distinct records collapse into what looks like exact
+            # duplicate rows (confirmed live for M. tuberculosis
+            # "Fatty acid metabolism": 10/10 rows identical without these
+            # fields, vs. fadD15/fadD19/fadD35 etc. once selected).
+            "gene",
+            "feature_id",
+            "product",
+            "annotation",
         ]
 
         query = self._build_query_string(
--- a/src/tooluniverse/data/bvbrc_tools.json
+++ b/src/tooluniverse/data/bvbrc_tools.json
@@ -1717,6 +1717,34 @@
                       "integer",
                       "null"
                     ]
+                  },
+                  "gene": {
+                    "type": [
+                      "string",
+                      "null"
+                    ],
+                    "description": "Gene symbol for this pathway/EC record, e.g. \"fadD15\". Distinguishes genuinely different genes sharing the same EC number/pathway from true duplicate rows."
+                  },
+                  "feature_id": {
+                    "type": [
+                      "string",
+                      "null"
+                    ],
+                    "description": "BV-BRC feature id for this record, e.g. \"RefSeq.83332.12.NC_000962.CDS.2448160.2449962.fwd\"."
+                  },
+                  "product": {
+                    "type": [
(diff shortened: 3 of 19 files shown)


gasvn:

PR #318, opened 2026-07-19
Fix Round 22: variant/lipid/bibliometrics/synbio tool fixes

gasvn:
## Summary

Round 22 of the ongoing test-harness campaign. 5 role-play research personas (structural biologist, metabolomics researcher, synthetic biologist, clinical variant interpreter, bibliometrics researcher) exercised ~90 tool calls across 6 domains; every finding was CLI-verified against live upstream APIs before fixing.

## Fixes

- **ThreeDBeacons_get_annotations**: missing required `type` param now errors clearly instead of a raw upstream 422
- **GeneBe_classify_variant**: `hgvs_c`/`hgvs_p`/`effects` now extracted from the correct nested `acmg_by_gene[0]` location instead of being silently dropped
- **SynBioHub_***: detect synbiohub.org's new login wall (401/HTML redirect) with actionable errors instead of an opaque XML parse crash; wired up `SYNBIOHUB_API_TOKEN` support
- **MetabolomicsWorkbench_***: parse TSV-formatted responses (some endpoints ignore the requested JSON format) into structured rows instead of one raw string
- **SAbDab_get_structure**: detect SAbDab's SPA-migration HTML shell instead of silently reporting success with mislabeled non-PDB content; clarified `get_structure_summary`'s error message
- **SemanticScholar_get_paper_citations/get_paper_references**: fixed a wrong example paper ID in the docs (it resolved to an unrelated paper)
- **VEP_predict_pathogenicity**: added `variant_allele` so rsID-mode predictions at multiallelic sites can be attributed to the correct alt allele
- **OpenCRAVAT_list_annotators**: `category` filter now also matches name/title, not just type (every module currently shares `type="annotator"`, making the filter previously unable to find anything by name)
- **ORCID_search_researchers**: switched to the `expanded-search` endpoint so results carry names/institutions instead of bare IDs needing a separate `get_profile` call per candidate
- **DNA_primer_design**: short-template failures now hint at the tool's documented 200bp minimum instead of a bare range-mismatch message
- **SwissLipids_search**: exact-name matches now ranked to the front instead of being buried under derivative entries and lost to truncation
- **SwissLipids entity names**: stripped raw `<greek>`/`<stereo>` markup
- **Crossref_search_members / NEB_Tm_calculate**: doc-only caveats for multi-word query gaps and cross-tool Tm divergence

- NPAtlas underscore corruption in compound names — confirmed at the byte level to be genuine upstream data, not a client-side encoding bug
- MARRVEL API (`api.marrvel.org`) is down at the network level (confirmed via direct curl, DNS resolves fine but requests hang to timeout) — doesn't yet meet the project's 3-independent-attempt `broken_apis` retry threshold, flagging for future-round retry
- Two findings (GeneBe HTTP-500 detail-field CLI rendering, SemanticScholar_get_paper's generic error message) are duplicates of round 20's already-fixed `_extract_detail_hint` CLI change, pending merge in an earlier PR

## Test plan

- 8 new mocked unit test files + 2 extended existing test files, all passing


gasvn:

PR #319, opened 2026-07-19
Fix Round 23: immunology/microbiome/cardiology/genomics tool fixes

gasvn:
## Summary

Round 23 of the ongoing test-harness campaign. 5 role-play research personas (neuroscience/CNS drug discovery, immunology/vaccine design, environmental toxicology, microbiome/metagenomics, cardiology/precision-medicine) exercised ~100 tool calls across 6+ domains; every finding was CLI-verified against live upstream APIs before fixing.

## Fixes

- **IEDB_predict_mhcii_binding**: reads the `mhcii/` endpoint's real `rank` column instead of the nonexistent `percentile_rank` column, which silently defaulted every result to 100.0 regardless of true binding strength
- **IEDB_predict_mhci/mhcii_binding, predict_bcell_epitopes, predict_antigen_processing**: detect IEDB's plain-text error responses (e.g. an invalid HLA allele name) before parsing them as TSV, which previously produced bogus single-column rows keyed on the error message itself with a fabricated `status:"success"`
- **kegg_list_organisms**: switched off KEGG's retired `/list/organism` endpoint (HTTP 400) to `/list/genome`, with matching line-format parsing
- **GTDB_get_species**: normalize species-name capitalization before querying (the endpoint is case-sensitive, unlike its sibling search endpoint)
- **GTDB_search_genomes**: boost exact/prefix organism-name matches to the front, since the endpoint matches on whole taxonomic clades and can bury the actually-requested organism under unrelated family members
- **OpenTargets_get_target_expression_by_ensemblID**: updated the GraphQL query off the removed `expressions` field to the current `baselineExpression` biosample/datasource model
- **AOPWiki_get_aop / get_key_event**: detect AOP-Wiki's string `"unprocessable_entity"` not-found status (the existing check only matched numeric HTTP-style codes), which previously returned a silent `status:"success"` with every field null for an invalid id
- **IMPC_search_genes**: boost an exact marker-symbol match in the Solr query so it doesn't get buried outside the default result window
- **IMPC_get_gene_summary**: cross-check the genotype-phenotype core for `has_phenotype_data`, since the gene core's own phenotype-summary fields are inconsistently populated and can silently under-report real data
- **AHA_ACC_search_guidelines**: broadened the PubMed query to also match guideline titles, since recent joint-committee guidelines (e.g. the 2024 hypertrophic cardiomyopathy guideline) carry no Corporate Author field
- **EBIProteins_get_epitopes**: extract `iedb_ids` from the `xrefs` array instead of `evidences.source`, where the IEDB epitope ID actually lives
- **PGSCatalog_search_traits**: doc-only caveat that multi-word trait phrases can silently zero out upstream even when a near-identical trait exists

- MCP `find_tools`/`grep_tools` returning truncated/incomplete tool names, and `mcp__tooluniverse__execute_tool` failing session-wide — MCP-interface-layer issues, not tool code bugs
- OpenTargets known-drugs' null `disease` entries — confirmed genuine raw upstream GraphQL data (verified via direct API query), not a ToolUniverse transformation bug
- ADMET-AI's raw-torch-error message on invalid SMILES — needs the optional `[ml]` extra installed to safely verify a fix; current behavior is caught (not an unhandled crash), just a technical error message
- Several lower-severity findings needing further investigation before a confident fix: CTD name-resolution brittleness (BPA/hyphenation), ChEMBL structural-alert filter parameter apparently ignored server-side, CTD missing organism/PMID fields, NCBIDatasets taxonomy lineage returning unresolved numeric IDs, `grep_tools` regex-mode case sensitivity, NCBIDatasets silent-empty results for strain-qualified organism names, BVBRC fragmented synonym names (unverified against raw API)

## Test plan

- 9 new mocked unit test files, all passing

### DIFF
--- a/src/tooluniverse/aopwiki_tool.py
+++ b/src/tooluniverse/aopwiki_tool.py
@@ -20,6 +20,19 @@ def _get_json(url: str, timeout: int = 30) -> Any:
         return json.loads(resp.read().decode("utf-8", errors="ignore"))
 
 
+def _aopwiki_not_found(result: Any) -> bool:
+    """AOP-Wiki returns HTTP 200 with a body like
+    {"message": "This AOP does not exist.", "status": "unprocessable_entity"}
+    for an invalid id (confirmed live for both /aops and /events) -- there is
+    no HTTP error to catch. Detect that string status, not a numeric one.
+    """
+    return isinstance(result, dict) and result.get("status") in (
+        404,
+        500,
+        "unprocessable_entity",
+    )
+
+
 @register_tool(
     "AOPWikiListTool",
     config={
@@ -131,7 +144,7 @@ class AOPWikiDetailTool:
         except Exception as e:
             return {"status": "error", "error": f"AOPWiki API error: {e}"}
 
-        if not isinstance(result, dict) or result.get("status") in (404, 500):
+        if not isinstance(result, dict) or _aopwiki_not_found(result):
             return {
                 "status": "error",
                 "error": f"Key Event {event_id} not found or server error",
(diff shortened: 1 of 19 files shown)


gasvn:

PR #320, opened 2026-07-19
Fix Round 24: structural biology/epidemiology/plant/single-cell tool fixes

gasvn:
## Summary

Round 24 of the ongoing test-harness campaign. 5 role-play research personas (structural biology/crystallography, epidemiology/public health, plant/agricultural genomics, pharmacogenomics/drug safety, single-cell genomics) exercised ~130 tool calls across 6+ domains; every finding was CLI-verified against live upstream APIs before fixing.

## Fixes

- **CPIC_get_alleles**: the `limit` URL template placeholder was never substituted when omitted (no schema default), sending a literal `limit={limit}` to the API and returning all alleles unbounded instead of the documented default of 50
- **MarineRegions_search_by_name**: same missing-default pattern for `offset`
- **DiseaseSH_get_covid_historical / DiseaseSh_get_historical**: same pattern for `country`, which broke the documented "omit for global data" behavior with a 404
- **DiseaseSH_get_vaccine_coverage**: `country` is now required, since the upstream vaccine-coverage-by-country endpoint has no `all` global variant (confirmed live it 404s even with `country=all`, unlike the historical-cases endpoint)
- **hca_search_projects**: fixed the `disease` filter facet name (Azul rejects `disease` with a 400; the real facet is `donorDisease`) and fixed organ/donorDisease extraction, which read nonexistent top-level dict keys instead of the real nested `specimens`/`donorOrganisms` array fields
- **WHOGHO_get_indicator_data**: added the missing `$filter`/`$top` OData param mapping (the sibling tool `WHOGHO_search_indicators` already had it) so a country/year filter and row limit actually reach the API instead of being silently dropped
- **ExpressionAtlas_get_baseline**: removed the broken `gene_mentioned`/`gene_specific_count` fields, which were computed by full-text-searching experiment descriptions for a bare gene symbol — confirmed live this never matches real data, always reporting false. Now lists species-level baseline experiments honestly and points to `GxA_get_experiment_expression` for actual per-gene checks
- **Foldseek_search_structure**: the Python code's inline default (`tmalign`) disagreed with the documented JSON schema default (`3diaa`), so an omitted `mode` silently ran the slower, less specific TM-align search — confirmed live this explains implausible top hits for well-characterized folds (e.g. no EGFR-family structures in the top 10 for an EGFR kinase domain query); also fixed a copy-paste typo in the mode description
- **UniProtIDMap_to_pdb** (and its siblings via shared `_submit_and_poll`): jobs that complete almost instantly get results embedded directly in the status-check response instead of a `FINISHED` status, and the code returned that embedded copy immediately instead of falling through to the properly paginated results fetch — silently truncating to UniProt's unpaginated default page size (25) even when hundreds of matches exist (confirmed live: P00533→PDB has 354 matches, not 25)

Several findings need further investigation or are architecturally larger in scope: `CxGDisc_search_datasets`'s strict Cell-Ontology-label-only substring matching (no synonym/plural support), `PlantReactome_search_pathways`'s species filter 404ing for all but one species, `USDA_plants_search_by_name`'s broken single-word common-name ranking, and `EpiGraphDB_get_drugs_for_trait`'s possibly-noisy drug-repurposing results for some traits (may be inherent to the underlying loose-threshold GWAS gene-mapping data rather than a ToolUniverse bug).

One reported finding (`GxA_get_experiment_expression`'s `gene_id` filter) was re-verified live and found to work correctly — the original report tested genes genuinely absent from that specific experiment's row subset, a false positive.

## Test plan

- 5 new mocked unit test files, all passing

### DIFF
--- a/src/tooluniverse/data/cpic_tools.json
+++ b/src/tooluniverse/data/cpic_tools.json
@@ -1078,7 +1078,8 @@
             "integer",
             "null"
           ],
-          "description": "Maximum number of alleles to return (default 50)"
+          "description": "Maximum number of alleles to return (default 50)",
+          "default": 50
         }
       },
       "required": [
--- a/src/tooluniverse/data/disease_sh_ext_tools.json
+++ b/src/tooluniverse/data/disease_sh_ext_tools.json
@@ -10,7 +10,8 @@
             "string",
             "null"
           ],
-          "description": "Country name or ISO code. Use 'all' for global data. Examples: 'usa', 'china', 'india', 'uk', 'germany', 'all'"
+          "description": "Country name or ISO code. Use 'all' for global data. Examples: 'usa', 'china', 'india', 'uk', 'germany', 'all'",
+          "default": "all"
         },
         "lastdays": {
           "type": [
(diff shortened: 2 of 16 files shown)


gasvn:

PR #321, opened 2026-07-19
Fix Round 25: genetics/cheminformatics/proteomics/omics tool fixes

gasvn:
## Summary

Round 25 of the ongoing test-harness campaign. 5 role-play research personas (rare disease genetics/genetic counseling, cheminformatics/medicinal chemistry, proteomics/mass spectrometry, virology/infectious disease genomics, clinical trials/regulatory affairs) exercised ~145 tool calls across 6+ domains; every finding was CLI-verified against live upstream APIs before fixing.

## Fixes

- **MedlinePlus_get_genetics_condition_by_name / MedlinePlus_get_genetics_gene_by_name**: three compounding bugs fixed together. MedlinePlus's genetics download endpoints are case-sensitive on the identifier segment — an uppercase gene symbol like `FBN1.json` silently serves XML instead of the requested JSON despite the `.json` extension, forcing gene lookups down a much buggier fallback path. Lowercasing routes both condition and gene lookups through the reliable real-JSON response shape. Separately, the list-parsing helper only handled one of the two possible list shapes (real JSON vs. XML-parsed), so `genes` was always empty for conditions and `health_conditions` was duplicated for genes; and the gene handler read a `gene-summary` wrapper key that doesn't exist in real JSON responses, silently emptying every field. Also added the `inheritance` field, present in the raw upstream data but never extracted at all.
- **visualize_molecule_2d**: the parameter schema required `smiles`, `inchi`, AND `molecule_name` together, even though the tool's own implementation explicitly supports providing just one (with its own clear error when none are given) — the sibling tool `visualize_molecule_3d` correctly has no such requirement for the same three-way-alternative param set.
- **OmicsDI_search_datasets**: `omics_type`/`organism`/`tissue` filters were sent as separate query params, which the upstream API silently ignores — confirmed live that a Proteomics-filtered query still returned Transcriptomics-dominated results identical to an unfiltered search. OmicsDI only respects these as Lucene clauses folded into the query string itself; added a small custom tool class to build that composite query before delegating to the shared REST-tool request logic.
- **DrugCentral_search / DrugCentral_get_drug**: the `inchikey` field silently fell back to MyChem.info's internal record ID whenever the real InChIKey was absent — confirmed live for pembrolizumab (a monoclonal antibody with no small-molecule structure, hence legitimately no InChIKey upstream), which returned two different non-InChIKey identifiers as `inchikey` for the same drug. Fixed to report `None` instead of mislabeling an unrelated ID, and to deduplicate search results on the internal ID directly rather than on the now-frequently-None inchikey value.

A persona-D report on virology tools triggered an unrelated safety review; its concrete tool-bug findings (BV-BRC `host_group` enum documentation, ChEMBL target-filter schema mismatch, an MCP TLS-cert environment issue already seen in prior rounds) are left for a future round's more careful pass, and its "no systematic viral drug-resistance-mutation database" coverage-gap finding is deliberately not pursued.

Several other lower-confidence findings need further investigation before a confident fix: `Enamine_search_smiles`'s description overstating real API integration, `Metabolite_get_info`'s silent autocomplete substitution to an isotopologue compound, EBIProteins 404 handling losing upstream diagnostic detail, `UniProt_get_ptm_processing_by_accession`'s narrow PTM feature-type filter, and the MCP `find_tools`/`grep_tools` truncated-tool-name discovery-index issue seen recurringly across rounds.

## Test plan

- 4 new mocked unit test files, all passing

### DIFF
--- a/src/tooluniverse/_lazy_registry_static.py
+++ b/src/tooluniverse/_lazy_registry_static.py
@@ -504,6 +504,7 @@ STATIC_LAZY_REGISTRY = {
     "OMIMTool": "omim_tool",
     "ORCIDTool": "orcid_tool",
     "OSFPreprintsTool": "osf_preprints_tool",
+    "OmicsDITool": "omicsdi_tool",
     "OmniPathTool": "omnipath_tool",
     "OncoKBTool": "oncokb_tool",
     "OncoTreeBaseTool": "oncotree_tool",
--- a/src/tooluniverse/data/medlineplus_tools.json
+++ b/src/tooluniverse/data/medlineplus_tools.json
@@ -185,7 +185,7 @@
   },
   {
     "name": "MedlinePlus_get_genetics_condition_by_name",
-    "description": "Get detailed information from MedlinePlus Genetics corresponding to genetic condition name. Returns XML data parsed into a dictionary with condition description, related genes, synonyms, and other details.",
+    "description": "Get detailed information from MedlinePlus Genetics corresponding to genetic condition name. Returns condition description, related genes, inheritance pattern, synonyms, and other details.",
     "parameter": {
       "type": "object",
       "properties": {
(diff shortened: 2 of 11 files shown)


gasvn:

PR #322, opened 2026-07-19
Fix Round 26: ProtParam MW, OpenGenes diseases, FooDB redirects, MedlinePlus parsing, PanelApp search, EnsemblCompara trees, OrthoDB, OpenTargets GWAS/pagination

gasvn:
## Summary

Round 26 of the ongoing test-harness campaign. 5 role-play researcher personas (nutrition/food science, dermatology/skin genetics, reproductive biology/fertility, aging/geroscience, bioinformatics/sequence analysis) surfaced 39 findings; every finding below was CLI-verified against the live upstream API before and after the fix.

## Fixes

- **ProtParam_calculate** (`protparam_tool.py`): `_calc_mw()` added a water molecule's mass twice — once before the residue-summation loop, once after — overshooting every molecular weight by 18.02 Da regardless of sequence length. Verified: `ACDEFGHIK` now correctly returns 1019.14 Da (was 1037.15); 148-aa lysozyme now returns 16536.95 Da (was 16554.97), matching UniProt P61626's reference mass.
- **OpenGenes_get_gene** (`open_genes_tool.py`): `disease_categories` was always empty — `diseaseCategories` entries key their label as `icdCategoryName`, not `name` (confirmed via raw API curl to `open-genes.com/api/gene/LMNA`). Also surfaces a new `diseases` field (specific named disease associations like "Progeria") that was present in the raw API but never read at all.
- **FooDB_get_compound** (`foodb_tool.py`): retired/renumbered FDB ids 302-redirect to a completely unrelated compound's current page instead of 404ing (confirmed live: `FDB012199` → `FDB004133`, an unrelated compound). The tool silently followed the redirect via `requests`' default behavior and returned the wrong compound's data labeled with the originally-requested id. Now detects the redirect and reports "not a current id" instead.
- **MedlinePlusRESTTool._extract_text_content** (`medlineplus_tool.py`): a `<html:p>` paragraph containing a nested inline tag (e.g. `<html:i>FMR1</html:i>`) lost the tag's own text entirely ("The  gene provides instructions..." — double space), and a paragraph with *no* nested tag parsed as a bare string rather than a dict and was silently dropped by the old `isinstance(p, dict)` filter (lost the entire middle paragraph of FMR1's function description). The real-JSON html-string code path also only stripped `<p>`/`</p>`, leaving other inline tags like `<i>` literal in the output text.
- **PanelApp_search_panels** (new `panelapp_tool.py`): PanelApp's `search`/`q`/`name__icontains` query params are all silent no-ops server-side (confirmed via introspection — only an exact full-string `name=` match filters anything; the endpoint's own OpenAPI schema documents no search param at all). Every query previously returned the same unfiltered 434-panel list. New `PanelAppSearchTool` paginates the full panel list and filters client-side by name/disease_group/disease_sub_group.
- **Sequence_stats** (`sequence_analyze_tool.py`): `estimated_mw_da` was unlabeled monoisotopic mass, silently mismatched against the average mass most tools (ExPASy ProtParam, UniProt) report by default (off by ~11-18 Da). Renamed to `estimated_mw_monoisotopic_da`; description now points to `ProtParam_calculate` for average mass.
- **EnsemblCompara_get_gene_tree** (`ensembl_compara_tool.py`): four compounding bugs, confirmed live for TP53 (the tool's own documented example) and LYZ — `tree_id` was read from a nonexistent nested `tree.id` field instead of the real top-level `id`; leaf-member detection required a `species` key that leaf nodes never carry (species info lives under `taxonomy`), so `members` was always empty; `newick` was never populated since Ensembl's JSON response has no `newick` field in any mode (now fetched via a second request with a `text/x-nh` Content-Type header); and the Ensembl-gene-ID path form 404d because it was missing a required `{species}` path segment.
- **OrthoDB_search_groups** (`orthodb_tool.py`): only the first of up to 50 returned groups was enriched with `name`/`level_taxid` via the `/tab` endpoint (confirmed live: 9 of 10 results for "lysozyme" were bare `{group_id}` objects). Now enriches every returned group.
- **graphql_tool.remove_none_and_empty_values** (`graphql_tool.py`): a list entry like `{"disease": null}` recursed to `{}` after its null key was stripped, but the emptiness filter only checked the item's *pre-recursion* value, so the resulting `{}` survived. Confirmed live in `OpenTargets_get_associated_drugs_by_target_ensemblID`'s `diseases` array (bare `{}` interleaved with real entries for BRODALUMAB). Now filters on the recursed result.
- **OpenTargets_search_gwas_studies_by_disease** (`graphql_tool.py`): a legacy `EFO_` disease id silently returns `{"count": 0}` (a normal empty list, not a null entity), so the existing EFO→MONDO not-found detection — which only fires on a null entity — never caught it (confirmed live: `EFO_0000676` for psoriasis silently returns 0 studies while the current `MONDO_0005083` id returns 79). Now flags legacy EFO ids specifically when the studies count is 0.
- **OpenTargets_get_diseases_phenotypes_by_target_ensembl** (`opentarget_tools.json`): no `page` argument on `associatedDiseases`, so Open Targets silently applied its default page size regardless of the true count (confirmed live: SIRT1 has `count: 3432` but only 25 rows were ever returned, with no way to retrieve the rest). Added the same `page: Pagination` support already used by sibling OpenTarget tools (e.g. `OpenTargets_get_target_interactions_by_ensemblID`).

- MCP `execute_tool` TLS-cert-path and tool-registry-lookup failures — environment/deployment issue outside this repo, recurring across rounds 22-26.
- OpenTargets `Target.expressions` → `baselineExpression` GraphQL schema drift — duplicate of round 23's fix (unmerged).
- MedlinePlus genes-always-empty / health_conditions-duplicated — duplicate of round 25's fix (unmerged).
- `IMPC_get_gene_summary` `has_phenotype_data` and `IMPC_search_genes` relevance ranking — duplicates of round 23's fixes (unmerged).
- ChEMBL/GtoPdb gene-symbol-aware target search, ClinGen `gene_symbol` alias, MetaboAnalyst evidence-count fields, `disease_target_score` timeout-vs-pageSize semantics, OpenGenes `expression_change` legend — lower-severity or larger-scope items left for a future round.

## Test plan

- [x] 12 new/updated mocked unit test files covering every fix above
- [x] `uv run pytest tests/unit -q --no-cov` — 2201 passed, 11 skipped (all environmental: scanpy not installed, no-tools-loaded fixtures, one resource-exhaustion deselect)
- [x] code-simplifier pass applied (deduplicated a near-identical helper in `open_genes_tool.py`)

### DIFF
--- a/src/tooluniverse/_lazy_registry_static.py
+++ b/src/tooluniverse/_lazy_registry_static.py
@@ -533,6 +533,7 @@ STATIC_LAZY_REGISTRY = {
     "OrthoDBTool": "orthodb_tool",
     "OxOTool": "oxo_tool",
     "PANTHERTool": "panther_tool",
+    "PanelAppSearchTool": "panelapp_tool",
     "PDBECompoundTool": "pdbe_compound_tool",
     "PDBeAPIRESTTool": "pdbe_api_tool",
     "PDBeLigandsTool": "pdbe_ligands_tool",
--- a/src/tooluniverse/data/metaboanalyst_tools.json
+++ b/src/tooluniverse/data/metaboanalyst_tools.json
@@ -53,7 +53,7 @@
   {
     "name": "MetaboAnalyst_name_to_id",
     "type": "MetaboAnalystTool",
-    "description": "Map metabolite common names to database identifiers (KEGG, HMDB, PubChem, ChEBI). Resolves metabolite names against the KEGG compound database and retrieves cross-references to HMDB, PubChem, and ChEBI. Returns matched name, molecular formula, and exact mass for each metabolite. Use for: converting metabolite lists to standardized IDs before pathway or enrichment analysis.",
+    "description": "Map metabolite common names to database identifiers (KEGG, PubChem, ChEBI). Resolves metabolite names against the KEGG compound database and retrieves cross-references to PubChem and ChEBI. Returns matched name, molecular formula, and exact mass for each metabolite. Note: hmdb_id is almost always null -- KEGG discontinued HMDB cross-references in its DBLINKS data years ago, so this field is populated only for the rare legacy entry that still carries one; use a dedicated HMDB lookup tool to resolve HMDB IDs. Use for: converting metabolite lists to standardized IDs before pathway or enrichment analysis.",
     "parameter": {
       "type": "object",
       "properties": {
--- a/src/tooluniverse/data/opentarget_tools.json
+++ b/src/tooluniverse/data/opentarget_tools.json
@@ -83,14 +83,15 @@
     },
     {
         "name": "OpenTargets_get_diseases_phenotypes_by_target_ensembl",
-        "description": "Find diseases or phenotypes associated with a specific target using ensemblId.",
+        "description": "Find diseases or phenotypes associated with a specific target using ensemblId. `associatedDiseases.count` is the total number of associations; `rows` returns only one page of them (25 by default) -- pass `page` to retrieve more (e.g. page 2 of a target with thousands of associations like SIRT1's 3432).",
         "label": [
             "Target",
             "Disease",
             "Phenotype",
             "Association",
             "OpenTarget",
-            "GraphQL"
+            "GraphQL",
+            "Pagination"
         ],
         "parameter": {
             "type": "object",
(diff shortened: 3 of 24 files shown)


gasvn:

PR #323, opened 2026-07-19
Fix Round 27: DepMap cell-line lookup, IMPC crash, Metabolite/CTD pubmed_ids, JASPAR validation, MGnify biome, ClinVar summary safety

gasvn:
## Summary

Round 27 of the ongoing test-harness campaign. 5 role-play researcher personas (oncology/precision cancer medicine, neuroscience/neurodegeneration, endocrinology/metabolic disease, microbiome/gut-health, RNA biology/epigenetics) surfaced 30 findings; every finding below was CLI-verified against the live upstream API before and after the fix.

## Fixes

- **DepMap_get_cell_line / DepMap_search_cell_lines / DepMap_get_cell_lines** (`depmap_tool.py`): the Cell Model Passports API silently ignores `filter[model]=model_name:X` on `/models` (confirmed live: A549, HeLa, and a nonsense string all returned the identical unfiltered first page, `model_id SIDM01774` every time). Switched to the real `/search/models?q=...` endpoint. Separately, even with a correctly resolved `model_id` most fields came back null: `tissue`/`cancer_type`/`sample_site`/`gender`/`ethnicity`/`age_at_sampling`/`tissue_status`/`msi_status` are not attributes on the model resource at all — they live on related sample/patient/model_msi_status resources — and `mutational_burden`/`model_name` were read from nonexistent attribute keys (real keys are `mutations_per_mb` and `names`, a list). Fixed by requesting `include=sample,sample.tissue,sample.cancer_type,sample.patient,model_msi_status` and resolving the relationship chain (including a live id-type mismatch — tissue/cancer_type ids come back as `int` while relationship pointers use `str`). The plural list endpoint had the same field-mapping bugs plus a nonexistent `meta.total` key (real key is `meta.count`); since the API has no working tissue/cancer_type filter under any syntax tried (5 variants tested), callers now get an explicit note instead of silently-ignored results.
- **IMPC_get_phenotypes_by_gene** (`impc_tool.py`): crashed with `"list index out of range"` whenever a valid `mgi_id` query matched zero records in the genotype-phenotype core — confirmed live for Lrrk2 (MGI:1913975) and Brca1 (MGI:104537), both real, resolvable genes that simply have no significant phenotype calls in this core. One line was unguarded for an empty `docs` list, unlike the adjacent line for `mgi_id`.
- **Metabolite_get_diseases / HMDB_get_diseases** (`metabolite_tool.py`): the RENCI Automat CTD-mirror fallback (used since CTD's native `batchQuery.go` became CAPTCHA-blocked) hardcoded `pubmed_ids` to an empty list for every result, even though the mirror's edge data does carry `publications`. `direct_evidence` is genuinely unavailable from this source (confirmed via live API introspection — no predicate field exists at all in the mirror's edge schema), now documented honestly in the tool description instead of silently returning null.
- **jaspar_search_matrices** (`jaspar_tools.json`): the parameter schema lacked `additionalProperties: false`, so an unrecognized param like `tf_name` (a very natural guess, matching the sibling tool `UniBind_search_datasets`'s naming convention) was silently forwarded to and ignored by JASPAR's API, returning the entire 5935-matrix unfiltered database with no indication anything was wrong.
- **MGnify_get_samples** (`mgnify_expanded_tool.py`): `environment_biome` was always null in list results because `attributes["environment-biome"]` is null for most samples upstream; the real per-sample biome path lives in the sample's own `biome` relationship instead, which the list endpoint already includes per item.
- **annotate_variant_multi_source** (`compound_variant_tool.py`): the summary block derived `clinvar_classification` from ClinVar's deliberate gene-context fallback list (returned when no exact match exists) without checking `exact_match` first — confirmed live that EGFR L858R, a ClinVar Oncogenic/Tier I-Strong variant, was being summarized as "Uncertain significance" from an unrelated p.Leu828Ser record. This was a real clinical-safety-relevant bug. Now only derives the classification from a real match.

- MCP `execute_tool` TLS-cert-path and tool-registry-lookup failures, and OpenTargets abbreviated-name/full-name mismatches via MCP discovery — environment/deployment issues outside this repo, recurring across rounds 22-27.
- OncoKB demo-mode `geneExist:false` messaging, CIViC/ClinVar variant-name search ergonomics (3-letter codes, protein-change search), ClinGen/GenCC alias resolution (PARK8), PathwayCommons `limit`-vs-depth param naming, `MGI_get_gene` `cross_references` duplication, `NCBIDatasets_get_taxonomy`'s unresolved numeric lineage, `MyDisease_get_disease`'s undelivered `disgenet` field, `ChIPAtlas_enrichment_analysis` returning a submission URL instead of computed results, APPRIS SSL flakiness, and several OpenTargets camelCase-vs-snake_case parameter mismatches — lower-severity, larger-scope, or environment-dependent items left for a future round.

## Test plan

- [x] 4 new mocked unit test files + 2 updated existing test files covering every fix above
- [x] `uv run pytest tests/unit -q --no-cov` — 2189 passed, 11 skipped (all environmental: scanpy not installed, no-tools-loaded fixtures, one resource-exhaustion deselect)
- [x] code-simplifier pass applied (deduplicated the model-list parsing logic shared by `_get_cell_lines`/`_search_cell_lines` in `depmap_tool.py`; removed a redundant default in `metabolite_tool.py`)

### DIFF
--- a/src/tooluniverse/compound_variant_tool.py
+++ b/src/tooluniverse/compound_variant_tool.py
@@ -286,7 +286,15 @@ class CompoundVariantAnnotationTool(BaseTool):
                 summary["sources_with_data"].append(source)
 
         clinvar = annotations.get("clinvar", {})
-        if clinvar.get("variants"):
+        # _parse_clinvar's "variants" list falls back to unrelated top-10
+        # gene-level context variants when the specific change isn't found
+        # (exact_match: False) rather than being empty -- confirmed live
+        # this previously let clinvar_classification silently report an
+        # unrelated variant's classification (e.g. EGFR L858R -- a
+        # ClinVar-Oncogenic, Tier I-Strong variant -- summarized as
+        # "Uncertain significance" from an unrelated p.Leu828Ser record)
+        # whenever exact_match was False. Only derive it from a real match.
+        if clinvar.get("exact_match") and clinvar.get("variants"):
             classifications = [
                 v["classification"]
                 for v in clinvar["variants"]
--- a/src/tooluniverse/data/jaspar_tools.json
+++ b/src/tooluniverse/data/jaspar_tools.json
@@ -609,7 +609,8 @@
           "default": 20,
           "description": "Results per page."
         }
-      }
+      },
+      "additionalProperties": false
     },
     "fields": {
       "endpoint": "https://jaspar.elixir.no/api/v1/matrix/",
--- a/src/tooluniverse/data/metabolite_tools.json
+++ b/src/tooluniverse/data/metabolite_tools.json
@@ -134,7 +134,7 @@
   },
   {
     "name": "Metabolite_get_diseases",
-    "description": "Get curated disease associations for a metabolite using CTD (Comparative Toxicogenomics Database). Accepts HMDB ID, compound name, or PubChem CID. Resolves the identifier to PubChem to get the canonical compound name, then queries CTD for chemical-disease associations with direct evidence. Returns disease name, disease ID (MeSH), disease categories, direct evidence type (marker/mechanism), and supporting PubMed IDs.",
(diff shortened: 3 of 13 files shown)


gasvn:

PR #325, opened 2026-07-19
Fix Round 29: ClinGen variant-classification multi-minute hang, GtoPdb/GMrepo silent param drops

gasvn:
## Summary

Round 29 of the ongoing test-harness campaign. 5 role-play researcher personas (psychiatric genetics, nephrology, musculoskeletal genetics, gastroenterology/liver disease, autoimmune/immunodeficiency) surfaced 41 findings; every finding below was CLI-verified against the live upstream API before and after the fix.

## Fixes

- **ClinGen_get_variant_classifications** (`clingen_tool.py`): fetched the entire ClinGen Evidence Repository (`classifications/all`, a ~22MB/~10k-row TSV) with no server-side filter and filtered client-side. This routinely took 2-5+ minutes and sometimes hung past a 300s client timeout — confirmed independently by **5 separate persona sessions across rounds 28 and 29** (PAH, BRCA1, DMD, CACNA1C, PTPN22, PKD1). Live API archaeology found a genuinely different, fast, server-side-filtered JSON endpoint on the same API (`classifications?gene=X` or `?variant=X`, no `/all`) responding in ~1-2s. Rewritten to use it, mapping the new JSON shape back onto the tool's existing declared 7-field schema so there's no downstream shape change. The upstream `variant=` param was confirmed to resolve to the variant's *whole gene* rather than just that variant (e.g. `variant=CA114360` returned 25 PAH rows, not 1) — added a client-side exact-match narrowing pass for that.
- **GtoPdb_search_diseases** (`gtopdb_tools.json`): the parameter schema lacked `additionalProperties: false`, so an unrecognized param like `disease_name` was silently forwarded and ignored by the upstream API, returning 50 completely unrelated diseases with `status: success` and no indication the filter had no effect.
- **GMrepo_get_phenotypes** (`gmrepo_tools.json`): same `additionalProperties` gap, plus the tool's own description text ("Optionally filter by keyword...") primed users toward a nonexistent `keyword` parameter instead of the real `query` parameter — confirmed live this exact mistake silently returned the unfiltered default phenotype list. Description now names the real parameter.

## Deferred (documented, not fixed this round)

- `ClinGen_get_actionability_adult`/`pediatric`'s columnar-table parsing — duplicate of round 28's fix (PR #324, unmerged).
- OpenTargets `Target.expressions` → `baselineExpression` GraphQL schema drift — duplicate of round 23's fix (PR #319, unmerged).
[...]

PR #324, opened 2026-07-19
Fix Round 28: CTD CAS/MeSH resolution, STITCH wrong database, UniProt protein names, PubChem image crash, ClinGen actionability parsing

gasvn:
## Summary

Round 28 of the ongoing test-harness campaign. 5 role-play researcher personas (cardiovascular genetics, infectious disease/antimicrobial resistance, ophthalmology/vision genetics, hematology, toxicology/environmental health) surfaced 48 findings; every finding below was CLI-verified against the live upstream API before and after the fix.

## Fixes

- **CTD_get_chemical_diseases / CTD_get_gene_chemicals** (`ctd_tool.py`): the RENCI graph mirror stores CAS RNs and MeSH ids as CURIEs (`CAS:335-67-1`, `MESH:C006780`), not bare — confirmed live that bare `335-67-1` and `C006780` (the tool's own docstring example!) both failed to resolve while the CURIE-prefixed forms succeeded. The tool's own description explicitly promises CAS RN and MeSH ID support as plain input. Now retries with the appropriate prefix when the bare term doesn't resolve and the input matches a CAS/MeSH-SCR pattern.
- **STITCH_get_chemical_protein_interactions / get_interaction_partners / resolve_identifier** (`stitch_tool.py`): pointed at `string-db.org` based on a stale comment claiming `stitch.embl.de` "moved" there. STRING and STITCH are genuinely different sister databases sharing the same API software — confirmed live that string-db.org's own `/resolve` fuzzy-matched the chemical name "aspirin" to an unrelated protein (SLC17A4), and its `/interaction_partners` rejected the real STITCH chemical id as "not found". Every chemical-protein query was silently returning wrong, mislabeled protein-protein data. Now points to `stitch-db.org` (the current live STITCH host; `stitch.embl.de` now redirects there per the shared API docs' own database table).
- **UniProt_search / uniprot get_by_accession-family tools** (`uniprot_tool.py`): `protein_name` was always empty for unreviewed (TrEMBL) UniProtKB entries — over 99% of the database. Reviewed (Swiss-Prot) entries have a `recommendedName`; unreviewed entries have no such field at all, only `submissionNames` (a list). Confirmed live for Q53707 (mecA/PBP2' from *Staphylococcus aureus*). New shared `_protein_name()` helper falls back to `submissionNames[0]`, used by both call sites that previously duplicated (and incompletely handled) this extraction.
[...]

PR #313, opened 2026-07-18
Test-harness consolidated: rounds 1-17 (64 verified fixes, 179 new tests)

gasvn:
## Summary

Consolidates 11 sequential test-harness PRs (#302 → #303 → #304 → #305 → #306 → #307 → #308 → #309 → #310 → #311 → #312, covering rounds 1–17) into one mergeable PR. Each round ran 2–5 role-play researcher-persona agents against a fresh domain slice of the ~2700-tool surface, then CLI-verified every reported issue (`python3 -m tooluniverse.cli run <Tool> '<args>'`, usually with raw upstream-API `curl` to confirm root cause) before fixing — roughly half of raw persona reports across all 17 rounds turned out to be false positives or naming-intuition mismatches with already-actionable errors, and were deliberately left unfixed (see each round's own PR for the full false-positive/deferred lists — not repeated here for length).

**64 verified, root-caused bugs fixed** across ~60 tool/CLI/config files, with **179 new/updated unit tests**, all passing. 122 files changed overall (5537 insertions, 519 deletions).

### A note on merging

8 files were independently touched by more than one round: `clinvar_tool.py` (rounds 1–7, 8, 10), `clinicaltrials_gov_tools.json` (rounds 1–7, 8, 9), `ctg_tool.py` / `dailymed_tool.py` / `unified_guideline_tools.py` (rounds 1–7 and 9, plus 10 for the guideline tools), `cli.py` (rounds 1–7 and 13), `epa_envirofacts_tool.py` (rounds 1–7 and 15), and `pubmed_tool.py` (rounds 9 and 16). All eight compose correctly and are all present in this PR. Two kinds of conflict came up during merging:

- **Textual conflicts** (git couldn't auto-merge): `clinvar_tool.py` twice — once where round 8 hoisted an unrecognized-parameter check that round 1–7 still had duplicated inline (kept round 8's hoisted version), and once where round 8's hyphen-stripping gene-name fix and round 10's "strip a trailing gene/protein qualifier word" fix both touched the same `if "gene" in arguments:` block (combined into one transformation: strip the qualifier word, then strip hyphens).
[...]

PR #326, opened 2026-07-20
Consolidate test-harness rounds 18-29 (12 PRs) into one

gasvn:
## Summary

Consolidates 12 individual round PRs (#314-325) plus the earlier consolidated PR #313 (rounds 1-17) into a single mergeable branch, following the same process used for the #302-312 → #313 consolidation. Merged all 13 branches in ascending PR-number order, resolving real conflicts and deduplicating overlapping fixes along the way. Replaces #313-325, which will be closed in favor of this PR.

Net: 297 files changed, ~15.7k insertions / ~1.3k deletions across ~120 tool/config files and ~180 test files.

## Conflict resolution and dedup

Most of the 13 branches merged cleanly (no conflicts). Where two rounds touched the same file, each case was resolved on its merits, generally favoring the more complete/root-caused fix and dropping the redundant one — not a blind "later wins":

- **cli.py** (#314×#316): kept #316's more general `_extract_detail_hint` (handles JSON-string `detail` payloads too), preserving #314's separate, non-overlapping top-level `suggestion` handling.
- **sabdab_tool.py** (#313×#318): both fixed the same SPA-migration-detection bug; kept #318's positive PDB-record-prefix check (more robust than #313's HTML-only negative check), updated the superseded test's assertion.
- **opentarget_tools.json** (#313×#319): duplicate `expressions`→`baselineExpression` schema fix; kept #313's broader field set.
- **gtdb_tool.py** (#315×#319): NOT a duplicate — different bugs (prefix-stripping vs. case-normalization) in the same function; kept both.
- **iedb_prediction_tool.py** (#314×#319): duplicate rank-field/non-tabular-response fix; kept #319's more defensive fallback chain and its `_iedb_error_response` helper (fires before `_parse_tsv`'s check), updated the superseded test.
- **expression_atlas_tool.py**/`.json` (#316×#320): genuine design fork (keep gene-mention tagging with a caveat vs. remove it entirely) — kept #320's removal after confirming live that the "does it mention the gene" signal is false for every real gene, not just weak.
- **foldseek_tool.py**, **uniprot_idmapping_tool.py** (#313×#320): pure comment/documentation duplicates of the same already-applied fix.
- **compound_variant_tool.py** (#313×#323): kept #313's version (adds an explicit `clinvar_note` diagnostic), updated #323's test assertion to match; kept #323's new tests elsewhere.
- **depmap_tool.py** (#315×#323, flagged as high-risk): live-verified both approaches. #323's real `/search/models?q=` endpoint is correct and fast, but confirmed live it does **not** return exact matches first (`q=HCC38` returns "HCC38-BL" before the exact "HCC38" match) — which would silently break `DepMap_get_cell_line`'s by-name lookup (relies on `cell_lines[0]`). Combined #323's endpoint with #315's exact-match-first client-side sort; kept #323's more robust JSON:API relationship-graph resolver for `_get_cell_line`; removed the now-dead `_find_catalog_page`/binary-search path for cell lines (still used for gene search). Added a new regression test for the ordering fix.
- **clingen_tool.py** — actionability (#315×#324): kept #324's version (retains `gene`/`Gene`/`hgncId` filter fallbacks that #315 drops); removed #315's now-dead module-level `_actionability_rows_to_dicts`/`_filter_curations_by_gene` functions in favor of #324's static-method version.
- **clingen_tool.py** — variant classifications (#315×#325): kept #325's version (#315's breaks the tool's declared 7-field `return_schema`). **Found and fixed a new gap while merging**: #325's rewrite dropped the "require a filter" guard entirely — live-verified that calling `/classifications` with neither `gene` nor `variant` is just as unbounded as the original `/classifications/all` bug this fix was meant to solve (30s+ hang, no response). Restored the guard (now requiring gene *or* variant) and added a regression test that asserts the API is never called with no filter.
- **ctd_tool.py** (#314×#324): kept #314's version (also handles bare NCBI Gene IDs, which #324 misses); simplified `_resolve_curie` to drop its now-fully-redundant internal CAS/MeSH retry, since #314's `_candidate_curies` already generates every candidate the loop tries. Updated two tests whose mocked call-count/ordering assumed the old (superseded) retry sequence.
- **stitch_tool.py** (#315×#324): combined both — kept #315's `_endpoint_unavailable_error()` helper (its message correctly says the *endpoint* is unavailable even for a correctly-resolved identifier; #324's message misleadingly implied the identifier itself might be wrong) with #324's fuller module docstring/host-migration rationale.

## Other findings surfaced during consolidation (not conflicts, spot-checked per reviewer request)

- **PR #317's `NeuroMorphoTool`/`AllenBrainTool` HTTP-status-truthiness fix had 2 vacuous regression tests** — verified by reverting the fix and rerunning: both tests still passed, because the tests mocked `e.response` as a bare `MagicMock()`, whose `__bool__` defaults to `True` regardless of `status_code`, so the tests never actually exercised the `bool(e.response)` vs `e.response is not None` distinction the fix is about. Fixed the mocks to use a real `requests.Response` (whose `__bool__` genuinely reflects `.ok`), confirmed the tests now fail against the reverted code and pass against the real fix.
- **PR #313's PubMed hyphen-to-space query transform has a real, broad blast radius**: live-verified `cancer statistics 2020-2023 review` returns 22 results with the hyphen vs. 577 with it replaced by a space — a 26x difference for any query containing what looks like a numeric range, not just the narrow compound-word case (`CYP3A5-guided`) it was written for. A naive "don't touch hyphens between digits" fix doesn't work either, since the original motivating case (`CYP3A5-guided`) also has a hyphen preceded by a digit. Left as-is (already shipped, not part of this consolidation's scope) but flagging for a dedicated follow-up round.
- **PR #313's PubMed date-range placeholder backfill** (`mindate`/`maxdate` defaulting to `"1000"`/`"3000"` when only one bound is supplied) was spot-checked live and confirmed correct — open-ended ranges filter as expected in both directions.

## Test plan

- [x] `uv run pytest tests/unit -q --no-cov` — full suite green after every individual merge step and again after the final `uv.lock` cleanup (2189+ tests, only the usual 11 environmental skips: scanpy not installed, no-tools-loaded fixtures, one resource-exhaustion deselect)
- [x] Live-verified `ClinGen_get_variant_classifications` — fast response for `{"gene": "PAH"}`, and the restored guard fails fast with a clear error for `{}` instead of hanging
- [x] New/updated regression tests for every dedup decision above where behavior actually changed (DepMap exact-match ordering, ClinGen empty-filter guard, CTD retry-order, IEDB/SAbDab message-text updates)
- [x] `git diff origin/main..HEAD -- uv.lock pyproject.toml` confirmed none of the 13 source branches legitimately touch dependencies; local `uv.lock` drift from running tests during the merge was reverted before pushing


gasvn:

PR #327, opened 2026-07-20
Fix Round 30: Orphanet slash-encoding, gene-disease tool wrong OpenTargets call, CPIC case-sensitivity, OpenGenes 404 handling

gasvn:
## Note on base branch

This PR targets `test-harness/consolidated-rounds-1-29` (PR #326), not `main`, since #326 is still open/unmerged and this round was branched from its tip to avoid rediscovering/duplicate-fixing bugs already resolved there. Once #326 merges into `main`, this PR should be retargeted to `main` (or merged after #326 lands, whichever is cleaner) -- the diff below is round 30's changes only.

## Summary

Round 30 of the test-harness sweep: 5 role-play researcher personas exercised dermatology/skin genetics, reproductive medicine/fertility genetics, sleep medicine/circadian biology, geriatrics/aging biology, and forensic/population genetics -- domains not previously covered. All 5 hit a session-wide MCP-server environment issue (stale uv cache breaking network/native-library tool calls, consistent with prior rounds' recurring finding) and worked around it via the CLI fallback, surfacing several genuine, CLI-verified issues.

## Fixes

- **Orphanet_get_gene_diseases**: gene names containing a literal "/" (e.g. "lamin A/C" for LMNA, which has 21 curated Orphanet disease associations including Hutchinson-Gilford progeria syndrome) silently returned 0 diseases. Confirmed live: the correctly percent-encoded slash (`%2F`) 404s on Orphadata's router, while replacing "/" with "-" before encoding succeeds. Added a shared `_encode_orphadata_gene_name()` helper.
- **gather_gene_disease_associations** (`CompoundGeneDiseaseAssociationTool`): for a gene-only query, the OpenTargets sub-call passed the gene *symbol* into `OpenTargets_get_disease_ids_by_name` (a disease-name search), returning 0-1 spurious substring-match hits instead of the gene's real disease associations (confirmed live: LMNA has 6084 in OpenTargets). This meant the tool's core concordance-scoring feature could never show OpenTargets agreeing with any other source on a gene-only query -- directly undermining its purpose. Fixed by resolving the gene symbol to its Ensembl target ID first, then querying its real associated-diseases list.
- **CPIC_get_gene_info / CPIC_get_alleles**: CPIC's gene/allele tables store symbols uppercase, so a lowercase or mixed-case input silently returned `status: success` with 0 rows (confirmed live: `"cyp2d6"` → 0 rows, `"CYP2D6"` → 1). Added an `uppercase_params` config mechanism to `BaseRESTTool`, mirroring the existing `lowercase_params` one already used for CPIC's drug-name table (which stores lowercase -- the opposite direction).
- **OpenGenesGeneTool**: an unknown gene symbol signals "not found" via a genuine HTTP 404 (confirmed live), which the tool's generic exception handler previously turned into a hard error -- a different response envelope than the tool's own existing graceful handling for the same conceptual outcome (a 200 response with a malformed/empty body). Added a `not_found_ok` option to the shared `_fetch_json()` helper so both paths route through the same graceful "not in Open Genes" response.

- `ChEMBL_search_mechanisms`'s `target_chembl_id` param being declared in the schema despite being rejected at runtime -- confirmed this is deliberate, self-documenting design (the schema description itself says "NOT SUPPORTED", and the runtime rejection message correctly redirects callers to the right alternative tools), not a bug.
- `GTEx_get_median_gene_expression`'s description mismatch with "V11" -- already fixed in a prior round; both tool descriptions in this worktree already correctly state V8 is the default and V11 doesn't exist via this API. The persona's finding was likely against the stale MCP server.
- `EVA_get_variants_by_gene`'s `cohortStats.ALL` showing zeroed ref/alt allele counts while a real minor-allele-frequency value (0.0059) sits correctly in the adjacent `maf` field -- this is legitimate (if unusually shaped) upstream data for a summary-only submission lacking individual genotype breakdowns, not a ToolUniverse defect.
- `dbsnp_search_by_gene`'s lack of relevance/significance ordering, ChEMBL's literal substring target-name matching, ARNTL/BMAL1 Ensembl symbol alias resolution (confirmed upstream Ensembl behavior, not ours to fix), and DGIdb's duplicate interaction-type entries -- lower-severity or larger-scope items left for a future round.
- Recurring MCP `execute_tool` TLS-cert-path and tool-type-registry-desync failures, and `find_tools`/`grep_tools` surfacing renamed/nonexistent tool names (e.g. `OpenTargets_get_asso_targ_by_dise_efoI`) -- environment/deployment issues outside this repo, recurring across every round.

## Test plan

- [x] 4 new/updated mocked unit test files covering every fix above
- [x] Full `tests/unit` suite green (only the usual environmental skips)
- [x] code-simplifier pass applied (only reflow needed, logic already clean)

### DIFF
--- a/src/tooluniverse/base_rest_tool.py
+++ b/src/tooluniverse/base_rest_tool.py
@@ -275,6 +275,22 @@ class BaseRESTTool(BaseTool):
                     if isinstance(value, str):
                         arguments[key] = value.lower()
 
+            # Fix-R30E-3: the mirror-image case -- CPIC's gene/allele tables
+            # store symbols uppercase (e.g. "CYP2D6"), so a lowercase or
+            # mixed-case input to `symbol=eq.{symbol}`/`genesymbol=eq.{...}`
+            # silently returns an empty success (confirmed live: "cyp2d6" ->
+            # 0 rows, "CYP2D6" -> 1). `fields.uppercase_params` lists params
+            # to upcase.
+            uppercase_params = (
+                self.tool_config.get("fields", {}).get("uppercase_params") or []
+            )
+            if uppercase_params:
+                arguments = dict(arguments)
+                for key in uppercase_params:
+                    value = arguments.get(key)
+                    if isinstance(value, str):
+                        arguments[key] = value.upper()
+
             url = self._build_url(arguments)
             params = self._build_params(arguments)
 
--- a/src/tooluniverse/compound_gene_disease_tool.py
+++ b/src/tooluniverse/compound_gene_disease_tool.py
@@ -66,11 +66,22 @@ class CompoundGeneDiseaseAssociationTool(BaseTool):
         results_by_source["OMIM"] = self._extract_genes_or_diseases(r, "OMIM")
 
         # 3. OpenTargets
-        r = _try_tool(
-            "OpenTargets",
-            "OpenTargets_get_disease_ids_by_name",
-            {"name": disease or gene},
-        )
+        # Fix-R30D-6: for a gene-only query, passing the gene symbol into
+        # OpenTargets_get_disease_ids_by_name (a disease-name search) doesn't
+        # look up the gene's associated diseases at all -- confirmed live it
+        # returns 0-1 spurious hits that merely happen to contain the gene
+        # symbol as a substring (e.g. "congenital muscular dystrophy due to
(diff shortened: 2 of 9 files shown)


gasvn:

PR #328, opened 2026-07-20
Fix Round 31: annotate_variant_multi_source ClinVar/UniProt/rsid bugs, OpenTargets datasource+note bugs, CPIC alias/note fixes

gasvn:
## Note on base branch

This PR targets `test-harness/round-30-domain-sweep` (PR #327), not `main`, since PR #326 (rounds 1-29 consolidated) and PR #327 (round 30) are both still open/unmerged and this round was branched from #327's tip to avoid rediscovering/duplicate-fixing bugs already resolved there. Once #326 and #327 merge into `main`, this PR should be retargeted to `main`. The diff below is round 31's changes only.

## Summary

Round 31 of the test-harness sweep: 5 role-play researcher personas exercised otolaryngology/hearing genetics, nutritional genomics, addiction medicine pharmacogenomics, urology/GU oncology genetics, and vaccine immunology -- domains not previously covered. All 5 hit the same recurring session-wide MCP-server environment issue (stale uv cache) and worked around it via the CLI fallback. Notably, 4 different personas independently hit distinct bugs concentrated in one shared tool.

## Fixes

- **annotate_variant_multi_source** (`CompoundVariantAnnotationTool`), three sub-fixes found by 3 different personas:
  - ClinVar sub-call used `query` (documented as an alias for `condition`, a disease-name search), not the dedicated `gene` parameter -- confirmed live this only *coincidentally* matched conditions whose name happened to contain the gene symbol (40 rows for HOXB13 vs the real 1943 via the correct `gene` param). Also added rsid→gene resolution via dbSNP, since none of ClinVar/CIViC/gnomAD/UniProt are rsid-keyed (confirmed live both `query` and `variant_id` set to a bare rsid return 0 rows) -- an rsid-only query no longer silently skips every source.
  - ClinVar result rows always had `"condition": ""` -- confirmed live `ClinVar_search_variants` never returns a condition field at all (it's a search filter on that tool, not a returned field). Removed the misleading always-empty key.
  - UniProt parsing expected `data` to itself be a list; the real shape is a dict with a `results` list -- confirmed live this always fell through to a truncated Python `repr()` string instead of parsed fields.
- **OpenTargets_get_evidence_by_datasource**: the `chembl` datasourceId is retired upstream (confirmed live + via GraphQL evidence introspection: `clinical_precedence` is the current id covering the same drug-evidence category) -- updated the description's valid-sources list. Also fixed a shared bug in `graphql_tool.py`: the IntOGen-specific "0 evidence rows" note fired for *any* `OpentargetTool`-based tool with a zero count, not just the actual IntOGen tool -- confirmed live it misattributed a zero count from this tool's own `chembl` query to IntOGen (never queried), even telling the caller to "use OpenTargets_get_evidence_by_datasource instead" while that IS the tool being called.
- **CPIC_get_gene_info**: added `gene`/`gene_symbol` aliases for `symbol`, matching the sibling tool `CPIC_search_gene_drug_pairs`' existing alias pattern. While implementing this, found and fixed a real bug in the shared `BaseRESTTool` `path_aliases` mechanism it depends on: an alias key wasn't removed from `args` after being mapped to its canonical name, so it leaked into the query string as an unrecognized param -- confirmed live this isn't harmless, PostgREST-backed APIs (CPIC) try to parse every query key as a filter and 400 with "failed to parse filter" on it, breaking the whole request.
- **CPIC_get_recommendations**: the "no recommendations found" note always claimed "this guideline uses a dosing algorithm rather than a recommendation table" whenever the drug-filtered result was empty -- confirmed live this is wrong for a multi-drug guideline filtered to a drug it doesn't cover (e.g. CYP2D6/opioids guideline 100416 has 66 real recommendation rows, just none for methadone/buprenorphine/naltrexone, only codeine/tramadol/hydrocodone). Now distinguishes "no table at all" from "table exists, not for this drug".

- `EVA_get_variants_by_gene`'s response size (per-sample genotype dumps making a 30-variant response ~474k lines) -- real usability issue but a larger-scope fix (would need a summary/compact mode).
- Orphanet has no disease-name-to-ORPHA-code resolution tool, so `Orphanet_get_genes` silently returns empty for a disease name instead of a numeric code -- a genuine coverage gap, not a code bug in the existing tool.
- `search_clinical_trials` intermittently returning a JSON-decode error disguised as "no studies found" -- plausible but not reliably reproduced this round.
- `OpenTargets_get_associated_targets_by_disease_efoId`'s lack of pagination despite the API supporting it, `dbsnp_search_by_gene`'s lack of relevance ordering, HPA nTPM fields returning null even for well-characterized genes (not independently verified against the live HPA UI), `dbsnp_get_variant_by_rsid`'s always-"N" allele field, and `CPIC_get_drug_info`'s silent empty result for brand names -- lower-severity or larger-scope items.
- Recurring MCP `execute_tool` TLS-cert-path and tool-type-registry-desync failures, and `find_tools`/`grep_tools` surfacing renamed/nonexistent tool names -- environment/deployment issues outside this repo, recurring across every round.
- **Two false positives caught via live re-verification**: (1) a persona reported `IEDB_predict_mhcii_binding`'s `percentile_rank` always 100.0 -- this exact bug was already fixed in an earlier round and the current code correctly maps `rank`→`percentile_rank` (confirmed live for the exact peptide cited in the report). (2) a persona reported `ClinGen_get_variant_classifications` taking ~230s for GJB2 -- the current code responds in ~3.7s, matching round 29's fix; likely a transient environment issue during that persona's session, not a regression.

## Test plan

- [x] 12 new/updated mocked unit test files covering every fix above
- [x] Full `tests/unit` suite green (only the usual environmental skips)
- [x] code-simplifier pass applied (one redundant nested-if simplified, rest already clean)

### DIFF
--- a/src/tooluniverse/base_rest_tool.py
+++ b/src/tooluniverse/base_rest_tool.py
@@ -66,11 +66,18 @@ class BaseRESTTool(BaseTool):
         """
         url = self.tool_config["fields"]["endpoint"]
 
-        # Apply path_aliases: map alias → canonical name before substitution
+        # Apply path_aliases: map alias → canonical name before substitution.
+        # Fix-R31B-3: an alias key left in `args` after this point leaked into
+        # _build_params as an unrecognized query param -- confirmed live this
+        # isn't just harmless noise: PostgREST-backed APIs (e.g. CPIC) try to
+        # parse every query key as a filter expression and 400 on it
+        # ("failed to parse filter"), breaking the whole request. Always pop
+        # the alias key once consulted, whether or not the canonical name was
+        # already present.
         path_aliases = self.tool_config.get("fields", {}).get("path_aliases", {})
         for alias, canonical in path_aliases.items():
-            if alias in args and canonical not in args:
-                args[canonical] = args[alias]
+            if alias in args:
+                args.setdefault(canonical, args.pop(alias))
 
         # Replace all path parameters from user args
         for key, value in args.items():
--- a/src/tooluniverse/compound_variant_tool.py
+++ b/src/tooluniverse/compound_variant_tool.py
@@ -81,27 +81,39 @@ class CompoundVariantAnnotationTool(BaseTool):
         sources_failed: List[str] = []
 
         # Resolve the gene up front: ClinVar/CIViC/gnomAD/UniProt are all queried by
-        # gene (or rsid), NOT by a bare protein change like "V600E" (which matches
-        # nothing in ClinVar). Derive the gene from the explicit arg or the variant.
+        # gene, NOT by a bare protein change like "V600E" (which matches nothing in
+        # ClinVar). Derive the gene from the explicit arg, the variant, or (Fix-R31B-1)
+        # dbSNP's rsid lookup when only an rsid was given -- ClinVar's own tools have
+        # no direct rsid parameter (variant_id is ClinVar's own internal numeric id,
+        # not a dbSNP rsid; confirmed live both "query"/"variant_id" set to "rs671"
+        # return 0 results), so resolving to a gene first is the only way to use
+        # ClinVar/CIViC/UniProt at all for an rsid-only query.
         gene_for_gnomad = gene
         if not gene_for_gnomad and variant:
             parts = variant.split()
             if parts:
                 gene_for_gnomad = parts[0]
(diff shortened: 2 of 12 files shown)


gasvn:

PR #329, opened 2026-07-20
Fix Round 32: CPIC empty-result note, ChEMBL adalimumab ID, MARRVEL gene aliases, HPA tissue-panel note

gasvn:
## Summary

Fourth round-tool-sweep in this stacking chain (based on #328 / round 31's branch tip, since #326/#327/#328 remain unmerged). 4 confirmed, live-verified fixes from 3 researcher-persona role-play sessions (pulmonology, rheumatology, transplant medicine pharmacogenomics -- the session's subagent spawn cap was hit mid-round, so only 3 of the planned 5 personas ran).

- **base_rest_tool.py**: new generic `fields.empty_result_note` mechanism -- surfaces an explanatory note when an exact-match backend returns an empty list, instead of building a growing per-drug alias table.
- **cpic_tools.json**: `CPIC_get_drug_info` now explains that CPIC indexes one canonical name per drug (e.g. "cyclosporine" not "cyclosporin", "tacrolimus" not "FK506") via the new `empty_result_note` field. Live-verified: `{"name": "FK506"}` now returns the note; `{"name": "cyclosporine"}` (a real match) does not.
- **chembl_tools.json**: corrected two tool descriptions that misidentified `CHEMBL1201581` as adalimumab -- it's actually infliximab (confirmed via `ChEMBL_get_drug` showing `"Infliximab (chimeric mab)"`). `CHEMBL1201580` is the real adalimumab (confirmed via its "D2E7" development code and "Abrilada" biosimilar synonym).
- **marrvel_tool.py / marrvel_tools.json**: `MARRVEL_get_gene` and `MARRVEL_get_omim_phenotypes` now accept `gene`/`gene_symbol` aliases -- the natural-language guess `{"gene_symbol": "PTPN22"}` previously failed schema validation entirely, unlike every other gene-input tool in the codebase. Live-verified against CFTR and PTPN22.
- **hpa_tools.json**: `HPA_get_gene_basic_info_by_ensembl_id`'s description now clarifies that "RNA tissue specific nTPM" only lists tissues HPA classifies as enriched/enhanced (confirmed live: only 3 tissues returned for CFTR -- gallbladder, intestine, pancreas), not a full per-tissue panel, and points to the comprehensive-details tool / GTEx for full coverage.

## Deferred / false positives this round

[...]

PR #330, opened 2026-07-20
Fix Round 33: HPA comprehensive-details optional flags, DailyMed limit alias

gasvn:
## Summary

Fifth round-tool-sweep in this stacking chain (based on #329 / round 32's branch tip, since #326/#327/#328/#329 remain unmerged). The session's subagent spawn cap (200/200) was still fully exhausted this round (confirmed by a direct retry that failed identically), so this round's persona role-play investigation was done directly via CLI across 3 domains -- ophthalmology/inherited retinal disease genetics, endocrinology/diabetes pharmacogenomics, hepatology/drug-induced liver injury pharmacogenomics -- instead of spawning researcher subagents.

2 confirmed fixes, both live-verified against upstream APIs before and after:

- **hpa_tools.json**: `HPA_get_comprehensive_gene_details_by_ensembl_id` required `include_images`/`include_antibodies`/`include_expression` in its schema even though every one of their descriptions promises "defaults to true" and the Python `run()` already reads them via `arguments.get(x, True)`. Any caller who took the description at its word and omitted them hit a hard schema validation error before the tool's own default logic ever ran. Live-verified: `{"ensembl_id": "ENSG00000156313"}` (RPGR) previously failed with `'include_images' is a required property`; now succeeds and returns the full comprehensive record. Removed the three flags from `required` and added matching `"default": true` schema entries.
- **dailymed_tool.py / dailymed_tools.json**: `DailyMed_search_spls` only recognized its own `pagesize` param. A caller guessing `limit` -- the dominant pagination param name across ToolUniverse (CPIC, ClinVar, GWAS, ...) -- was silently ignored with no error: `{"drug_name": "isoniazid", "limit": 3}` returned all 44 matching labels instead of 3. Added `limit` as an alias for `pagesize`.

- ClinVar gene searches for RPGR, ABCB11, ATP7B -- all return real, correctly-shaped data.
- CPIC drug/allele lookups for glimepiride and HLA-B gene-drug pairs, including a direct confirmation that round 31's "guideline covers other drugs, not this one" note (from PR #328) fires correctly for glimepiride's own G6PD guideline record.
- GWAS Catalog SNP lookups for TCF7L2, PubMed literature search, DailyMed SPL search -- all working.

## Test plan

- [x] New mocked unit tests for both fixes: `limit`/`pagesize` precedence and default fallback (DailyMed); required-array and default-value config checks (HPA).

### DIFF
--- a/src/tooluniverse/dailymed_tool.py
+++ b/src/tooluniverse/dailymed_tool.py
@@ -43,7 +43,11 @@ class SearchSPLTool(BaseTool):
         if arguments.get("published_date_eq"):
             params["published_date[eq]"] = arguments["published_date_eq"]
 
-        params["pagesize"] = arguments.get("pagesize", 100)
+        # Fix-R33A-1: "limit" is the dominant pagination param name across
+        # ToolUniverse (CPIC, ClinVar, GWAS, ...), so a caller guessing it
+        # here instead of DailyMed's own "pagesize" was silently ignored --
+        # confirmed live (limit=3 still returned all 44 isoniazid labels).
+        params["pagesize"] = arguments.get("pagesize") or arguments.get("limit") or 100
         params["page"] = arguments.get("page", 1)
 
         try:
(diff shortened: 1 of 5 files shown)


gasvn:

PR #331, opened 2026-07-20
Fix Round 34: IEDB MHC-II binding endpoint hard-errors on short sequences

gasvn:
## Summary

Sixth round-tool-sweep in this stacking chain (based on #330 / round 33's branch tip, since #326-#330 remain unmerged). The session's subagent spawn cap remains fully exhausted (now confirmed across 3 consecutive rounds), so this round's investigation was again done directly via CLI across 3 domains: cardiology/inherited arrhythmia and cardiomyopathy genetics, neurology/epilepsy pharmacogenomics, infectious disease/antimicrobial resistance genomics.

1 confirmed fix, live-verified before and after:

- **iedb_prediction_tool.py / iedb_prediction_tools.json**: IEDB's `mhcii` endpoint defaults its sliding-window scoring length to 15 residues server-side and hard-errors on any shorter sequence (`"The length of input sequence is less than the input/default length 15."`) -- confirmed live, including against this tool's own registered test example (`"PKYVKQNTLKLAT"`, 13 residues). Unlike the sibling MHC-I methods in the same file, this endpoint never exposed IEDB's own supported `length` override parameter, so any short peptide always failed with no way to work around it. Auto-shrinks the window to the sequence's own length when it's under 15, and exposes `length` in the schema for explicit control, mirroring the MHC-I pattern. Live-verified: the tool's own example now succeeds (`percentile_rank: 0.23`, strong binder), and a normal-length sequence still defaults to window=15 with no regression.

- ClinVar gene searches for KCNQ1 (Long QT) and SCN1A (Dravet syndrome).
- CPIC recommendations for clopidogrel (CYP2C19), carbamazepine (HLA-A/B), and abacavir (HLA-B*57:01).
- gnomAD constraint metrics for MYH7 (hypertrophic cardiomyopathy).
- BV-BRC antimicrobial-resistance phenotype and specialty-gene searches (MRSA, blaKPC, M. tuberculosis resistance genes).
- GTEx expression and ClinGen variant classifications for SCN1A (fast, correct).
- Orphanet name search for Brugada syndrome.
- This also reconfirmed two earlier-round fixes still work correctly on fresh real-world cases: round 32's MARRVEL `gene_symbol` alias (KCNQ1 OMIM phenotypes) and round 31's CPIC "guideline covers other drugs, not this one" note.

## Test plan

- [x] New mocked unit tests for the length auto-shrink/override behavior (short-sequence auto-fill, long-sequence no-op, explicit override takes precedence).

### DIFF
--- a/src/tooluniverse/data/iedb_prediction_tools.json
+++ b/src/tooluniverse/data/iedb_prediction_tools.json
@@ -94,6 +94,10 @@
           "type": "string",
           "description": "Prediction method: 'netmhciipan_el' (recommended), 'netmhciipan_ba', 'nn_align'",
           "default": "netmhciipan_el"
+        },
+        "length": {
+          "type": ["integer", "null"],
+          "description": "Sliding-window peptide length IEDB scores within 'sequence'. IEDB defaults this to 15 server-side and errors if 'sequence' is shorter than it; omit to auto-match the sequence's own length when it's under 15."
         }
       },
       "required": [
--- a/src/tooluniverse/iedb_prediction_tool.py
+++ b/src/tooluniverse/iedb_prediction_tool.py
@@ -326,6 +326,19 @@ class IEDBPredictionTool(BaseTool):
             "sequence_text": sequence,
             "allele": allele,
         }
+        # Fix-R34A-1: IEDB's mhcii endpoint defaults its sliding-window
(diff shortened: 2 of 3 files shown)


gasvn:

PR #332, opened 2026-07-20
Fix Round 35: CPIC drug-name resolution silently substituted a different drug

gasvn:
## Summary

Seventh round-tool-sweep in this stacking chain (based on #331 / round 34's branch tip, since #326-#331 remain unmerged). The session's subagent spawn cap remains fully exhausted (4th consecutive round), so this round's investigation was again done directly via CLI across 3 domains: hematology-oncology/leukemia genomics, nephrology/CKD genetics, psychiatry/psychopharmacogenomics.

1 confirmed fix, live-verified before and after -- **higher severity than a typical round's findings**, since it silently substitutes one drug's clinical guidance for another's rather than merely ignoring a param or hard-erroring:

- **cpic_search_pairs_tool.py**: `CPIC_get_recommendations` resolves a drug name to a guideline/rxnorm id via a fuzzy `ilike.*name*` substring match against CPIC's `/drug` table, then blindly takes the first result row. Whenever one drug's name is a substring of another's, this silently resolves to the WRONG drug with no indication of the substitution: querying `"citalopram"` returned **escitalopram's** guideline/rxnorm id (confirmed live -- `"citalopram"` is a substring of `"escitalopram"`, which happened to sort first in CPIC's unordered response), so every recommendation row that came back was for escitalopram's dosing guidance, not citalopram's. Confirmed the same collision independently for `"phenytoin"` (a substring of `"fosphenytoin"`). A clinician asking for one drug's CYP2C19/CYP2D6 dosing guidance would silently receive a different, related-but-distinct drug's guidance instead. Fixed by preferring an exact case-insensitive name match over the first fuzzy hit, with the fuzzy fallback preserved for genuine partial/typo queries (e.g. `"citalop"`).

- ClinVar gene searches for FLT3 (AML) and COL4A5 (Alport syndrome).
- gnomAD constraint metrics for APOL1; GWAS Catalog SNPs for APOL1; MARRVEL OMIM phenotypes for APOL1 (FSGS).
- OncoKB variant annotation for FLT3 (correctly reports its demo-mode-unavailable warning, confirming an earlier round's fix still holds) and BRAF (real demo-mode data).
- CPIC recommendations for amitriptyline (no collision); CPIC's clean rejection of a non-indexed drug (aripiprazole) with an actionable error message.

## Test plan

- [x] New mocked unit tests: exact match wins over a substring match sorted first, fuzzy fallback still applies when no exact match exists, and an end-to-end `CPIC_get_recommendations` call resolves to the exact requested drug.

### DIFF
--- a/src/tooluniverse/cpic_search_pairs_tool.py
+++ b/src/tooluniverse/cpic_search_pairs_tool.py
@@ -35,8 +35,25 @@ def _resolve_drug_to_guideline_id(
         )
         r.raise_for_status()
         rows = r.json()
-        if rows and rows[0].get("guidelineid"):
-            return rows[0]["guidelineid"], rows[0].get("rxnormid")
+        if not rows:
+            return None
+        # Fix-R35C-1: substring `ilike` matching picks a false first hit
+        # whenever one drug name is a substring of another -- confirmed
+        # live for "citalopram" (a substring of "escitalopram") and
+        # "phenytoin" (a substring of "fosphenytoin"): querying the shorter
+        # name silently returned the OTHER drug's guideline/rxnorm id
+        # because it happened to sort first in CPIC's unordered response,
+        # so CPIC_get_recommendations returned a different drug's dosing
+        # guidance with no indication of the substitution. Prefer an exact
+        # case-insensitive match; only fall back to the first fuzzy hit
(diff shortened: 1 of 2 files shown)


gasvn:

PR #333, opened 2026-07-20
Fix Round 36: seven DatasetTool-typed tools wrongly required params with real defaults

gasvn:
## Summary

Eighth round-tool-sweep in this stacking chain (based on #332 / round 35's branch tip, since #326-#332 remain unmerged). The session's subagent spawn cap remains fully exhausted (5th consecutive round), so this round's investigation was again done directly via CLI across 3 domains: medical genetics/prenatal screening, immunology/primary immunodeficiency genetics, toxicology/environmental exposure genetics.

Round 36 opened by systematically re-probing other fuzzy-name-resolution helpers across the codebase, per the lesson from round 35's CPIC substring-collision bug (`cryoet_tool.py`, `iedb_tool.py`, `drugcentral_tool.py`, `gwas_tool.py`) -- all confirmed safe, either exact match by design or a multi-result filter shown to the caller rather than a silent single-row pick. A useful negative result, not a wasted check.

1 confirmed fix affecting **7 tools that share one root cause** -- live-verified before and after:

- **dataset_tools.json**: `drugbank_vocab_search`, `drugbank_vocab_filter`, `drugbank_full_search`, `drugbank_links_search`, `dict_search`, `dili_search`, and `diqt_search` are all built on the same generic `DatasetTool` class, whose `run()` starts from `deepcopy(self.query_schema)` (a config block with real, working defaults for every non-essential param) and only then overlays whatever the caller supplied. Every one of these tools' JSON schemas still marked `search_fields`/`case_sensitive`/`exact_match`/`limit` (or `field`/`condition`/`limit` for the filter variant) as `required`, blocking that default-merging logic from ever running for a caller who -- reasonably -- supplied only the truly essential param. Confirmed live: `dict_search({"query": "terfenadine"})` previously failed outright with a schema validation error despite the tool being fully capable of answering with just that one argument (and correctly did once fixed, returning terfenadine's real DICTrank cardiotoxicity record -- terfenadine is a classic QT-prolongation/Torsades de Pointes example). Same bug class as an earlier round's HPA comprehensive-details fix, this time caught systemically across a whole shared tool type at once. Fixed by removing the over-required params from each tool's `required` array (keeping `query` for the six search-style tools; `drugbank_vocab_filter` needs nothing, since `field`/`condition`/`limit` all default too and `value` already had its own clear runtime validation for when it's genuinely needed).

- ClinVar gene searches for SMN1 (SMA) and IL2RG (X-SCID).
- Orphanet name search for 22q11.2 deletion syndrome (DiGeorge).
- GenCC gene-disease validity for TBX1.
- MARRVEL OMIM phenotypes for BTK (X-linked agammaglobulinemia).
- IEDB B-cell epitope prediction on its own documented example.
- ADMETAI toxicity prediction correctly reporting its missing optional ML dependency rather than failing silently.

## Test plan

- [x] New config-content regression tests asserting every `DatasetTool`-typed tool's `required` array contains only params without a `query_schema` default, plus end-to-end mocked functional tests for the search and filter code paths with minimal arguments.

### DIFF
--- a/src/tooluniverse/data/dataset_tools.json
+++ b/src/tooluniverse/data/dataset_tools.json
@@ -22,29 +22,28 @@
           "items": {
             "type": "string"
           },
-          "description": "Fields to search in. Available fields: 'DrugBank ID', 'Accession Numbers', 'Common name', 'CAS', 'UNII', 'Synonyms', 'Standard InChI Key'."
+          "description": "Fields to search in. Available fields: 'DrugBank ID', 'Accession Numbers', 'Common name', 'CAS', 'UNII', 'Synonyms', 'Standard InChI Key'. Defaults to ['Common name', 'Synonyms', 'DrugBank ID']."
         },
         "case_sensitive": {
           "type": "boolean",
-          "description": "Whether the search should be case sensitive."
+          "description": "Whether the search should be case sensitive.",
+          "default": false
         },
         "exact_match": {
           "type": "boolean",
-          "description": "Whether to perform exact matching instead of substring matching."
+          "description": "Whether to perform exact matching instead of substring matching.",
+          "default": false
         },
         "limit": {
           "type": "integer",
           "description": "Maximum number of results to return.",
           "minimum": 1,
-          "maximum": 1000
+          "maximum": 1000,
+          "default": 10
         }
       },
       "required": [
-        "query",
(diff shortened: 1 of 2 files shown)


gasvn:

PR #334, opened 2026-07-20
Fix Round 37: disease_target_score tools wrongly required pageSize

gasvn:
## Summary

Ninth round-tool-sweep in this stacking chain (based on #333 / round 36's branch tip, since #326-#333 remain unmerged). The session's subagent spawn cap remains fully exhausted (6th consecutive round), so this round's investigation was again done directly via CLI across 3 domains: pain medicine/anesthesiology pharmacogenomics, sports medicine/exercise genetics, rare disease/newborn screening genetics.

Round 37 opened by acting on round 36's lesson -- ran a codebase-wide scan for any tool property that has both a JSON-schema `"default"` value AND appears in that same tool's `"required"` array (a direct logical contradiction, since a required field's default can never be reached). Found **61 candidate instances across roughly 15 data files**. Rather than fixing all of them in one pass without individually live-verifying each, picked the clearest, most confidently systemic cluster to fix this round and left the rest as a documented backlog.

1 confirmed fix affecting **9 tools sharing one root cause** -- live-verified before and after:

- **disease_target_score_tools.json**: `disease_target_score` and its 8 datasource-specific variants (`chembl`/`eva`/`eva_somatic`/`cancer_gene_census`/`cancer_biomarkers`/`europepmc`/`expression_atlas`/`genomics_england`/`reactome_disease_target_score`) all required `pageSize` even though it defaults to 100 in both the schema description and `DiseaseTargetScoreTool.run()`'s own `arguments.get("pageSize", 100)` fallback. Confirmed live: `disease_target_score({"efoId": "MONDO_0011996", "datasourceId": "clinical_precedence"})` previously failed outright with a schema validation error despite being fully answerable without `pageSize` (now correctly returns real BCR-ABL1/CML target-score data, ABL1 top-ranked). Fixed by removing `pageSize` from all 9 `required` arrays.

## Backlog for future rounds (found but not yet individually verified/fixed)

The same default-vs-required contradiction also appears on `operation`/`action` routing fields across several other shared tool types: `cellxgene_census_tools.json` (6 tools), `cryoet_tools.json` (7 tools), `fourdn_tools.json` (4 tools), and `hca_tools.json` (2 tools). These need the same live-verify-before-fixing treatment, since an `operation` field could plausibly be vestigial/baked into `tool_config` rather than genuinely read from caller arguments -- worth checking each before assuming it's the same bug class.

- CPIC recommendations for codeine, tramadol, and fentanyl (including a reconfirmation that round 31's "guideline covers other drugs" note and round 35's exact-match drug resolution both still hold).
- gnomAD constraints for OPRM1 and ACTN3.
- ClinVar variants for RYR1 and PAH; MARRVEL OMIM phenotypes for RYR1 (malignant hyperthermia); Orphanet search for MCAD deficiency; GenCC gene-disease validity for ACADM.
- One self-caught false alarm: `HPO_get_genes_by_phenotype` appeared to return unrelated genes for "Hyperphenylalaninemia" but the actual issue was using the wrong HPO term ID (`HP:0003074` instead of the correct `HP:0004923`), confirmed via a live ontology lookup before concluding no bug existed.

## Test plan

- [x] New mocked unit test for the pageSize-omitted case (falls back to 100) plus config-content regression tests asserting no `disease_target_score` tool requires `pageSize` while `efoId` (and `datasourceId` for the generic tool) remain required.

### DIFF
--- a/src/tooluniverse/data/disease_target_score_tools.json
+++ b/src/tooluniverse/data/disease_target_score_tools.json
@@ -21,8 +21,7 @@
       },
       "required": [
         "efoId",
-        "datasourceId",
-        "pageSize"
+        "datasourceId"
       ]
     },
     "query_schema": "query getTargets($efoId: String!, $index: Int!, $size: Int!) { disease(efoId: $efoId) { id name associatedTargets(page: { index: $index, size: $size }) { count rows { target { approvedSymbol id } datasourceScores { id score } } } } }",
@@ -106,8 +105,7 @@
         }
       },
       "required": [
-        "efoId",
-        "pageSize"
+        "efoId"
       ]
     },
     "query_schema": "query getTargets($efoId: String!, $index: Int!, $size: Int!) { disease(efoId: $efoId) { id name associatedTargets(page: { index: $index, size: $size }) { count rows { target { approvedSymbol id } datasourceScores { id score } } } } }",
@@ -190,8 +188,7 @@
         }
       },
       "required": [
-        "efoId",
-        "pageSize"
+        "efoId"
       ]
     },
     "query_schema": "query getTargets($efoId: String!, $index: Int!, $size: Int!) { disease(efoId: $efoId) { id name associatedTargets(page: { index: $index, size: $size }) { count rows { target { approvedSymbol id } datasourceScores { id score } } } } }",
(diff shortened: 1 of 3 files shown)


gasvn:

PR #335, opened 2026-07-20
Fix Round 38: CELLxGENE Census tools wrongly required operation, plus a broken fallback

gasvn:
## Summary

Tenth round-tool-sweep in this stacking chain (based on #334 / round 37's branch tip, since #326-#334 remain unmerged). The session's subagent spawn cap remains fully exhausted (7th consecutive round), so this round's investigation was again done directly via CLI across 3 domains: critical care/ICU pharmacogenomics, pediatric oncology/solid tumor genomics, obstetrics/maternal-fetal medicine genetics.

This round worked through the top-priority item queued in the round-37 backlog: `cellxgene_census_tools.json`'s 6-tool `operation` cluster, all built on `CELLxGENECensusTool`. Verified via `BaseTool.validate_parameters()` directly (the underlying `cellxgene_census`/`tiledbsoma` packages aren't installed in this environment, but schema validation runs before that import is reached, so this is a fully valid verification path).

1 confirmed fix affecting **6 tools -- this one needed two parts, not one**:

- **cellxgene_census_tools.json**: each tool's `operation` property is a single-value enum matching its own `default` (e.g. `CELLxGENE_get_gene_metadata` always resolves to `"get_var_metadata"`), yet every one of the 6 tools required it. Removed from `required`.
- **cellxgene_census_tool.py**: removing the schema requirement alone was **not** sufficient. `CELLxGENECensusTool.run()`'s existing fallback was `arguments.get("operation", "get_metadata")` -- and `"get_metadata"` matches none of the tool's own if/elif dispatch branches. A caller who omitted `operation` would have silently mis-routed to the "Unknown operation" error path even after the schema fix. Caught this by testing the actual functional dispatch (mocking the two heavy packages so `run()` reaches its own logic) rather than stopping at schema validation. Fixed by switching the fallback to `self.get_schema_const_operation()` -- an existing `BaseTool` helper already used by ~27 other tool classes in this codebase for exactly this purpose. Verified end-to-end: with `operation` omitted, `run()` now dispatches to the correct private method for every one of the 6 tools.

Genuinely-required fields (`obs_value_filter` -- no default, the Census has 50M+ cells and an unfiltered query times out; `dataset_id` -- no default) are unaffected and still correctly rejected when missing.

## Backlog note

The two remaining queued "operation"/"action" clusters, `cryoet_tools.json` and `fourdn_tools.json`, were confirmed via code inspection this round to have the same two-part bug -- `cryoet_tool.py` falls back to a bare empty string (wrong for all 7 tools) and `fourdn_tool.py` falls back to a literal `"search"` (only coincidentally right for one of its 4 tools). Whoever picks those up next should apply the same two-part treatment, not just the schema fix. (Full detail in project memory: `project_test_harness_r37_required_default_backlog.md`.)

- CPIC's "guideline covers other drugs" note for morphine; FAERS disproportionality for propofol/bradycardia; cBioPortal cancer studies list.
- OncoKB TP53 R175H annotation; ClinVar variants for RB1 and F5; MARRVEL OMIM phenotypes for F5 (recurrent pregnancy loss); PGS Catalog trait search for preeclampsia.

## Test plan

- [x] New mocked unit tests covering config-content (required arrays), schema-validation-level behavior (operation omitted passes, other required fields still correctly rejected), and full functional dispatch routing (operation omitted still calls the correct private method for all 6 tools).

### DIFF
--- a/src/tooluniverse/cellxgene_census_tool.py
+++ b/src/tooluniverse/cellxgene_census_tool.py
@@ -30,7 +30,13 @@ class CELLxGENECensusTool(BaseTool):
             }
 
         try:
-            operation = arguments.get("operation", "get_metadata")
+            # Fix-R38A-1: "operation" is optional in the schema now (each
+            # config exposes exactly one valid value via enum+default), so
+            # a caller who omits it must still resolve to that tool
+            # instance's own operation -- the old bare "get_metadata"
+            # fallback doesn't match any dispatch branch below and would
+            # silently mis-route every one of these 6 tools.
+            operation = arguments.get("operation") or self.get_schema_const_operation()
             census_version = arguments.get("census_version", "stable")
 
             if operation == "get_census_versions":
--- a/src/tooluniverse/data/cellxgene_census_tools.json
+++ b/src/tooluniverse/data/cellxgene_census_tools.json
@@ -112,7 +112,6 @@
         }
       },
       "required": [
-        "operation",
         "obs_value_filter"
       ]
     },
@@ -209,9 +208,7 @@
           "default": "stable"
         }
       },
-      "required": [
-        "operation"
-      ]
+      "required": []
     },
     "return_schema": {
       "type": [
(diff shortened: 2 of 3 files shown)


gasvn:

PR #336, opened 2026-07-20
Fix Round 39: CryoET, 4DN, and HCA tools wrongly required their routing field

gasvn:
## Summary

Eleventh round-tool-sweep in this stacking chain (based on #335 / round 38's branch tip, since #326-#335 remain unmerged). The session's subagent spawn cap remains fully exhausted (8th consecutive round), so this round's investigation was again done directly via CLI across 2 domains: gastroenterology/IBD genetics, allergy and anaphylaxis genetics.

This round cleared the **entire "highest-priority clusters" section** of the round-37 backlog by applying the two-part fix pattern established in round 38 to all three remaining `operation`/`action` routing-field clusters -- schema required-array fix AND Python fallback fix together, since a schema-only fix would leave the tool silently mis-routing.

**13 tools fixed across 3 files**, all live-verified against their real, reachable upstream APIs:

- **cryoet_tools.json / cryoet_tool.py**: 7 tools (`list_datasets`, `get_dataset`, `list_runs`, `list_tomograms`, `list_annotations`, `list_tiltseries`, `list_depositions`). `CryoETTool.run()` fell back to a bare empty string, matching no dispatch branch. Confirmed live via the CryoET GraphQL API -- e.g. `CryoET_list_datasets({"organism_name": "Homo sapiens", "limit": 3})` now returns real datasets, and `CryoET_get_dataset({"dataset_id": 10463})` correctly dispatches to the right query.
- **fourdn_tools.json / fourdn_tool.py**: 4 tools (`search_data`, `get_file_metadata`, `get_experiment_metadata`, `get_download_url`). `FourDNTool.run()` fell back to a literal `"search"`, only coincidentally correct for one of the four tools. Confirmed live via the 4DN Data Portal API -- also incidentally confirms the earlier SSL certificate issue tracked in `broken_apis.md` has since been resolved upstream.
- **hca_tools.json / hca_tool.py**: 2 tools (`search_projects`, `get_file_manifest`). This one is a distinct **third sub-pattern**: `HCATool` uses a differently-named routing field (`"action"`, not `"operation"`) with no fallback at all, so `BaseTool.get_schema_const_operation()` (hardcoded to the `"operation"` property name) couldn't be reused directly -- fixed with an inline equivalent lookup against the tool's own `"action"` property schema default. Confirmed live via the HCA Azul API.

Every genuinely-required field with no schema default (`dataset_id`/`run_id` for CryoET, `file_accession`/`experiment_accession` for 4DN, `project_id` for HCA) was confirmed unaffected via live regression checks -- still correctly rejected when missing.

- ClinVar variants for NOD2 and TPSAB1; GenCC gene-disease validity for NOD2 (Blau syndrome); PGS Catalog trait search for Crohn disease; MARRVEL OMIM phenotypes for FLG (atopic dermatitis).
- One investigated-and-ruled-out finding: `Orphanet_search_by_name` returned no good matches for "hereditary alpha-tryptasemia" across several phrasings, including a bare "tryptasemia" substring returning zero hits even in fuzzy search mode -- confirmed this is a genuine upstream Orphanet coverage gap for a recently characterized condition, not a tool defect.

## Test plan

- [x] New mocked unit tests per tool family covering config-content (required arrays), schema-validation-level behavior, and full functional dispatch routing (operation/action omitted still calls the correct method, with genuinely-required fields still rejected).

### DIFF
--- a/src/tooluniverse/cryoet_tool.py
+++ b/src/tooluniverse/cryoet_tool.py
@@ -38,7 +38,13 @@ class CryoETTool(BaseTool):
     def run(self, arguments: Dict[str, Any]) -> Dict[str, Any]:
         """Execute the tool with given arguments."""
         try:
-            operation = arguments.get("operation", "")
+            # Fix-R39A-1: "operation" is optional in the schema (each config
+            # exposes exactly one valid value via enum+default), so a caller
+            # who omits it must still resolve to that tool instance's own
+            # operation -- the old bare "" fallback matched no dispatch
+            # branch below and would silently mis-route every one of these
+            # 7 tools.
+            operation = arguments.get("operation") or self.get_schema_const_operation()
 
             if operation == "list_datasets":
                 return self._list_datasets(arguments)
--- a/src/tooluniverse/data/cryoet_tools.json
+++ b/src/tooluniverse/data/cryoet_tools.json
@@ -42,9 +42,7 @@
           "default": 0
         }
       },
-      "required": [
-        "operation"
-      ]
+      "required": []
     },
     "return_schema": {
       "oneOf": [
@@ -175,7 +173,6 @@
         }
       },
       "required": [
-        "operation",
         "dataset_id"
       ]
     },
@@ -338,7 +335,6 @@
         }
       },
       "required": [
-        "operation",
         "dataset_id"
       ]
     },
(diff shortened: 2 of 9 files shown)


gasvn:

PR #337, opened 2026-07-20
Fix Round 40: ChEMBL similarity tools wrongly required their default-backed params

gasvn:
## Summary

Twelfth round-tool-sweep in this stacking chain (based on #336 / round 39's branch tip, since #326-#336 remain unmerged). The session's subagent spawn cap remains fully exhausted (9th consecutive round), so this round's investigation was again done directly via CLI across 3 domains: emergency medicine/toxicology overdose genetics, general surgery/perioperative genomics, radiology/nuclear medicine genomics.

This round picked off the `chembl_tools.json` item from the round-37 backlog's lower-priority list -- 2 tools fixed, live-verified before and after:

- **ChEMBL_search_similar_molecules** needed only a schema fix. Removed `similarity_threshold`/`max_results` from `required`; `chem_tool.py`'s `_search_similar_molecules` already had working `arguments.get(x, default)` fallbacks for both.
- **ChEMBL_search_similarity** needed a two-part fix -- a **new sub-pattern**, distinct from the `operation`/`action` routing-field bug found in rounds 38-39. Its endpoint template uses a URL path placeholder (`/similarity/{smiles}/{threshold}.json`), and `ChEMBLRESTTool._build_url()` in `chem_tool.py` (a separate class that subclasses `BaseTool` directly, not `BaseRESTTool`) only filled placeholders explicitly present in the caller's args, with no fallback to schema defaults for unfilled ones -- unlike `BaseRESTTool`'s `_build_url`, which already has exactly this fallback logic. Confirmed live: with only the schema fix, a caller omitting `threshold` got a literal `"{threshold}"` left in the URL and a 404. Fixed by adding the same schema-default-filling loop `BaseRESTTool` already has, directly into `ChEMBLRESTTool._build_url()`. Verified live: omitting `threshold` now correctly builds `"/similarity/CCO/80.json"` and returns real results; an explicit `threshold=95` still works unchanged.

Reconfirmed 3 earlier-round fixes still hold on fresh real-world cases: CPIC CYP2C19 gene info; FAERS acetaminophen/hepatic-failure disproportionality; `dict_search` for digoxin cardiotoxicity (reconfirms round 36's `DatasetTool` limit-omitted fix); ClinVar and MARRVEL for BCHE (postanesthetic apnea); DailyMed cefazolin search returning exactly the requested count (reconfirms round 33's limit-alias fix); gnomAD constraints, Pharos tissue expression, and DGIdb drug-gene interactions for SSTR2.

## Test plan

- [x] New mocked unit tests covering config-content (required arrays), the URL-building fix directly (path placeholder filled from schema default, explicit override still honored), and a full `run()` dispatch test with threshold omitted.

### DIFF
--- a/src/tooluniverse/chem_tool.py
+++ b/src/tooluniverse/chem_tool.py
@@ -51,6 +51,25 @@ class ChEMBLRESTTool(BaseTool):
             # Replace placeholders in URL
             for k, v in args.items():
                 url = url.replace(f"{{{k}}}", str(v))
+            # Fix-R40A-1: ChEMBL_search_similarity's endpoint template
+            # (/similarity/{smiles}/{threshold}.json) has a path placeholder
+            # that this loop never fills unless the caller explicitly
+            # supplies it -- confirmed live a caller omitting "threshold"
+            # (relying on its own schema "default": 80) got a literal
+            # "{threshold}" left in the URL and a 404, unlike BaseRESTTool's
+            # _build_url which already falls back to schema defaults for
+            # unfilled placeholders. Mirror that behavior here.
+            for key, prop in (
+                self.tool_config.get("parameter", {}).get("properties", {}).items()
+            ):
+                placeholder = f"{{{key}}}"
+                if (
+                    placeholder in url
+                    and isinstance(prop, dict)
+                    and "default" in prop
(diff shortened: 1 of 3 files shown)


gasvn:

PR #338, opened 2026-07-20
Fix Round 41: PubChem similarity and 2D-image tools wrongly required their default-backed params

gasvn:
## Summary

Thirteenth round-tool-sweep in this stacking chain (based on #337 / round 40's branch tip, since #326-#337 remain unmerged). The session's subagent spawn cap remains fully exhausted (10th consecutive round), so this round's investigation was again done directly via CLI across 3 domains: physical medicine & rehabilitation genetics, palliative care/chronic pain genetics, vascular medicine/thrombosis genetics.

This round picked off the `pubchem_tools.json` item from the round-37 backlog -- 2 tools fixed, live-verified before and after:

- **PubChem_search_compounds_by_similarity** needed only a schema fix (`threshold` removed from `required`), but with a nuance worth recording: the Python code conditionally omits the `Threshold` query parameter entirely when the caller doesn't supply one, rather than substituting a coded default. Live-tested that PubChem's own `fastsimilarity_2d` endpoint default matches the schema's documented 0.9/90% exactly -- the omitted-threshold and explicit-threshold=0.9 results are byte-identical -- confirming the schema-only fix is safe here.
- **PubChem_get_compound_2D_image_by_CID** needed a two-part fix -- **the same URL path-placeholder default-filling gap found in round 40 for ChEMBL_search_similarity, this time in a second, independent tool family**. Its endpoint template (`/compound/cid/{cid}/PNG?image_size={image_size}`) has a path placeholder, and `PubChemRESTTool._build_url()` (also a hand-rolled `BaseTool` subclass, not `BaseRESTTool`) raised a `ValueError` for any placeholder missing from the caller's args, with no schema-default fallback. Fixed by checking for a schema default before raising. Verified live: omitting `image_size` now correctly builds a URL with `image_size=200x200` and returns a real base64-encoded PNG; an explicit `image_size=300x300` still works unchanged.

Confirming this same gap independently in two unrelated tool families (ChEMBL, PubChem) makes it a systemic pattern worth checking in any other hand-rolled REST-wrapper class in this codebase, not a one-off.

Reconfirmed round 31's CPIC notes both fire correctly on fresh drugs: methadone correctly gets the "guideline covers other drugs, not this one" note, warfarin correctly gets the "dosing algorithm" note. Also correct: ClinVar variants for COL1A1 and F2; GenCC gene-disease validity for DMD; gnomAD constraints for COMT; MARRVEL OMIM phenotypes for SERPINC1 (antithrombin deficiency). This round's own PubChem similarity fix was naturally exercised in a real workflow scenario (baclofen structure similarity search).

## Test plan

- [x] New mocked unit tests covering config-content (required arrays), the URL-building fix directly (path placeholder filled from schema default, explicit override still honored, no regression for placeholders with no default), and full `run()` dispatch tests for both the image tool (with `image_size` omitted) and the similarity tool (confirming the omitted- and explicit-threshold requests match).

### DIFF
--- a/src/tooluniverse/data/pubchem_tools.json
+++ b/src/tooluniverse/data/pubchem_tools.json
@@ -420,8 +420,7 @@
         }
       },
       "required": [
-        "smiles",
-        "threshold"
+        "smiles"
       ]
     },
     "fields": {
@@ -472,8 +471,7 @@
         }
       },
       "required": [
-        "cid",
-        "image_size"
+        "cid"
       ]
     },
     "fields": {
--- a/src/tooluniverse/pubchem_tool.py
+++ b/src/tooluniverse/pubchem_tool.py
@@ -97,10 +97,26 @@ class PubChemRESTTool(BaseTool):
         placeholders = re.findall(r"\{([^{}]+)\}", url_path)
         for ph in placeholders:
             if ph not in arguments:
-                # If a placeholder cannot find corresponding value in arguments, report error
-                raise ValueError(
-                    f"Missing required parameter '{ph}' to replace placeholder in URL."
+                # Fix-R41A-1: a caller who omits a placeholder that has its
+                # own schema "default" (e.g. image_size) is relying on the
+                # documented default, not making an error -- confirmed live
+                # PubChem_get_compound_2D_image_by_CID({"cid": 2244}) used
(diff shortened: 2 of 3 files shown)


gasvn:

PR #340, opened 2026-07-20
Fix Round 42: MPD_get_phenotype_data silently ignored its own strain parameter

gasvn:
## Summary
- `MPD_get_phenotype_data` claims to search mouse phenotype data by strain (using ENCODE as a documented stand-in for the Mouse Phenome Database, since MPD's own REST API has no simple strain+category phenotype endpoint). But `mpd_tool.py`'s `run()` never used the `strain` argument at all -- every call returned the exact same generic ENCODE experiment list regardless of what strain was requested.
- The tool's own JSON config already documented an intended fix (`{strain}` templated into ENCODE's `biosample_ontology.term_name` filter), but `run()` never used `fields.endpoint` at all. That documented template was itself also broken -- confirmed live it always returns zero results for real strain names, since `biosample_ontology.term_name` is a tissue/cell-type ontology field, not a strain field.
- ENCODE's free-text `searchTerm` parameter does correctly surface strain-relevant experiments -- confirmed live: 292 hits for "C57BL/6J", 3 for "BALB/c", correctly 0 for a strain ENCODE has no data on ("DBA/2J"). Fixed to query with `searchTerm=<strain>`, added a response note disclosing `phenotype_category` isn't applied (ENCODE has no such concept), removed `strain` from `required` (same required-vs-default schema contradiction being cleared from the round-37 backlog elsewhere), and updated the description/endpoint template to match reality.
- Audited every other hand-rolled (non-`BaseRESTTool`) `_build_url()` for the identical URL path-placeholder default-filling gap found in ChEMBL (round 40, PR #337-equivalent) and PubChem (round 41). `ensembl_tool.py`, `medlineplus_tool.py`, `reactome_tool.py` share the same code shape but none currently has a required+default path-placeholder property -- confirmed via the scan script, nothing to fix there today.
- 3 fresh persona domains (neurodevelopmental/autism genetics, movement disorders/Parkinson's genetics, reproductive endocrinology genetics) found no new bugs this round.

## Test plan
- [x] New `tests/unit/test_mpd_strain_filter.py` (5 tests): required-list fix, `searchTerm` used instead of `biosample_ontology.term_name`, strain-with-slash handled correctly, default strain applied when omitted, response discloses the data-source limitation. All pass.
- [x] Live-verified end-to-end against the real ENCODE API: `MPD_get_phenotype_data({"strain": "C57BL/6J"})` now returns 292 strain-relevant results (previously returned unrelated generic data); `BALB/c` returns 3; omitted-strain defaults to "C57BL/6J".

### DIFF
--- a/src/tooluniverse/data/mpd_tools.json
+++ b/src/tooluniverse/data/mpd_tools.json
@@ -2,7 +2,7 @@
   {
     "type": "MPDRESTTool",
     "name": "MPD_get_phenotype_data",
-    "description": "Get mouse phenotype data from Mouse Phenome Database for specific strains and phenotypes",
+    "description": "Search ENCODE experiment records mentioning a given mouse strain (used as a Mouse Phenome Database alternative -- MPD's own REST API has no simple strain+category phenotype search; see https://phenome.jax.org/api for genuine curated MPD phenotype data). 'phenotype_category' is not applied -- ENCODE has no equivalent concept.",
     "parameter": {
       "type": "object",
       "properties": {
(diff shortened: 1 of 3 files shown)


gasvn:

PR #341, opened 2026-07-20
Fix Round 43: DBLP, HAL, and URL-fetch tools wrongly required their default-backed params

gasvn:
## Summary
- Four more tools from the round-37 backlog (JSON-schema property both `"required"` and given a `"default"`, so the default can never be reached), all matching the backlog's first sub-pattern (pure Python `.get(x, default)` fallback already correct, only the schema wrongly required it):
  - `DBLP_search_publications`: `limit` wrongly required (`dblp_tool.py` already does `int(arguments.get("limit", 10))`).
  - `HAL_search_archive`: `max_results` wrongly required (`hal_tool.py` already does `int(arguments.get("max_results", 10))`).
  - `get_webpage_title`: `timeout` wrongly required (`url_tool.py`'s `URLHTMLTagTool` already does `arguments.get("timeout", 20)`).
  - `get_webpage_text_from_url`: `timeout` wrongly required (`url_tool.py`'s `URLToPDFTextTool` already does `int(arguments.get("timeout", 30))`).
- Confirmed live for all four: omitting the param previously failed schema validation ("'<param>' is a required property") despite the tool being fully answerable without it; all four now succeed and return real results.
- 2 fresh persona domains (critical care/sepsis genomics, dental/oral medicine genetics) found no new bugs this round.

## Test plan
- [x] New `tests/unit/test_r43_optional_default_params.py` (8 tests): required-list fix for all four tools, plus mocked-request assertions confirming the schema default actually reaches the outgoing API call when the param is omitted. All pass.
- [x] Live-verified end-to-end: `DBLP_search_publications({"query": "transformer attention"})`, `HAL_search_archive({"query": "genomics"})`, and `get_webpage_title({"url": "https://www.wikipedia.org"})` all now succeed with the previously-required param omitted.

### DIFF
--- a/src/tooluniverse/data/dblp_tools.json
+++ b/src/tooluniverse/data/dblp_tools.json
@@ -17,8 +17,7 @@
         }
       },
       "required": [
-        "query",
-        "limit"
+        "query"
       ]
     },
     "return_schema": {
--- a/src/tooluniverse/data/hal_tools.json
+++ b/src/tooluniverse/data/hal_tools.json
@@ -17,8 +17,7 @@
         }
       },
       "required": [
-        "query",
-        "max_results"
+        "query"
       ]
     },
     "return_schema": {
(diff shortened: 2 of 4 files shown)


gasvn:

PR #342, opened 2026-07-20
Fix Round 44: Fatcat, OpenAIRE, NeuroMorpho, and TRIP tools wrongly required their default-backed params

gasvn:
## Summary
- Four more tools from the round-37 backlog (JSON-schema property both `"required"` and given a `"default"`, so the default can never be reached), all matching the backlog's first sub-pattern (pure Python `.get(x, default)` fallback already correct, only the schema wrongly required it):
  - `Fatcat_search_scholar`: `max_results` wrongly required (`fatcat_tool.py` already does `int(arguments.get("max_results", 10))`).
  - `OpenAIRE_search_publications`: `max_results` and `type` wrongly required (`openaire_tool.py` already does `int(arguments.get("max_results", 10))` and `arguments.get("type", "publications")` -- a valid enum value that maps to a real endpoint via `_endpoint_for_type`).
  - `NeuroMorpho_get_field_values`: `field_name` wrongly required (`neuromorpho_tool.py` already does `arguments.get("field_name", "species")`).
  - `TRIP_Database_Guidelines_Search`: `search_type` wrongly required (`unified_guideline_tools.py` already does `arguments.get("search_type", "guideline")`).
- Confirmed live for all four: omitting the param previously failed schema validation ("'<param>' is a required property") despite the tool being fully answerable without it; all four now succeed and return real results.
- 2 fresh persona domains (pediatric neurology/epilepsy genetics, inherited metabolic disease genetics) found no new bugs this round.

## Test plan
- [x] New `tests/unit/test_r44_optional_default_params.py` (8 tests): required-list fix for all four tools, plus mocked-request assertions confirming the schema default actually reaches the outgoing API call when the param is omitted. All pass.
- [x] Live-verified end-to-end: `Fatcat_search_scholar`, `OpenAIRE_search_publications`, `NeuroMorpho_get_field_values`, and `TRIP_Database_Guidelines_Search` all now succeed with the previously-required param omitted, returning real results.

### DIFF
--- a/src/tooluniverse/cli.py
+++ b/src/tooluniverse/cli.py
@@ -537,8 +537,10 @@ def _render_run(d: dict) -> str:
     # openFDA's "use a .exact keyword field" hint) — short_err is usually
     # just a generic "<Tool> API error". Surface it instead of dropping it.
     detail = d.get("detail")
+    detail_already_shown = False
     if isinstance(detail, str) and detail.strip() and detail.strip() not in short_err:
         lines.append(f"Detail: {detail.strip()[:300]}")
+        detail_already_shown = True
     details = d.get("error_details") or {}
 
     # Feature-25B-02: for "tool not found" errors, replace generic network tips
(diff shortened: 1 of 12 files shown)


gasvn:

PR #362, opened 2026-07-21
Fix Round 72: ClinGen gene-symbol filters used substring matching

gasvn:
## Summary

Round 72 shifted methodology per user direction: rather than picking another disease-area domain (9 consecutive clean domain-style rounds showed diminishing returns), this round used error-path/boundary-input testing against existing tool categories. Found and fixed a real, confirmed bug on the first pass:

**ClinGen's gene-validity and dosage-sensitivity search/filter functions matched gene symbols via substring containment** (`gene_upper in symbol.upper()`) against a column that holds a single precise HGNC symbol, not free text. Confirmed live: querying `{"gene": "OTC"}` (ornithine transcarbamylase deficiency) silently also returned **"NOTCH1" and "NOTCH2"** curations, since "OTC" is a literal substring of "N**OTC**H1"/"N**OTC**H2". Confirmed the exact blast radius via the raw ClinGen bulk CSV: exactly these 3 gene symbols contain "OTC" as a substring, matching the live result precisely.

Notably, `_get_gene_validity` (the sibling "get all, optionally filter by gene" function) **already used the correct exact-match pattern** — this bug was an inconsistency within the same file, not a uniform design choice. Fixed `_search_gene_validity`, `_get_dosage_sensitivity`, and `_search_dosage_sensitivity` to match it.

**Deliberately left untouched**: the actionability functions (`_get_actionability`/`_fetch_actionability_context`) also substring-match, but against `geneOrVariant`, which is explicitly documented as a comma-joined multi-gene panel field (e.g. `"BRCA1,BRCA2"`) — a genuine design constraint requiring substring matching, not the same bug. It likely has a *related* but distinct issue (the same "OTC"-in-"NOTCH1,NOTCH2" collision could theoretically occur within a joined field too) that would need its own investigation into the real-world field structure before changing — flagged for a future round rather than bundled in here.

## Verification

- Raw ClinGen bulk CSV (`search.clinicalgenome.org/kb/gene-dosage/download`) parsed directly: exactly `NOTCH1`, `NOTCH2`, `OTC` contain "OTC" as a substring — matches the live buggy result precisely before the fix.
- Live `tu run ClinGen_get_dosage_sensitivity '{"gene": "OTC"}'` before fix: 3 results (NOTCH1, NOTCH2, OTC). After fix: exactly 1 (OTC).
- Live-verified all three fixed operations (`get_dosage_sensitivity`, `search_dosage_sensitivity`, `search_gene_validity`) individually.
- Regression-checked: lowercase `"otc"` still matches (case-insensitivity preserved); a genuinely nonexistent gene symbol still correctly returns empty; querying the longer symbol (`"NOTCH1"`) doesn't spuriously match the shorter one it contains.

## Test plan

- [x] New `tests/unit/test_clingen_gene_symbol_exact_match.py`: 6 tests covering all three fixed operations, case-insensitivity, nonexistent-gene, and the inverse-direction collision check
- [x] Full ClinGen test suite (27 tests across all `test_clingen_*.py` files) — passes, no regressions
- [x] Full suite (`pytest tests/ --ignore=tests/test_claude_code_plugin.py`) — clean, 0 failures (the `test_claude_code_plugin.py` failures are pre-existing on `main`, unrelated plugin-doc drift, confirmed in earlier rounds)

### DIFF
--- a/src/tooluniverse/clingen_tool.py
+++ b/src/tooluniverse/clingen_tool.py
@@ -123,14 +123,18 @@ class ClinGenTool(BaseTool):
             # Parse CSV response
             curations = self._parse_csv(response.text)
 
-            # Filter by gene symbol (case-insensitive)
-            # Handle both "GENE SYMBOL" (from CSV) and "Gene Symbol" key formats
+            # Filter by gene symbol (case-insensitive, EXACT match -- this
+            # column holds a single precise HGNC symbol, not free text, and
+            # substring matching silently pulled in unrelated genes whose
+            # symbol happens to contain the query, e.g. "OTC" -> "NOTCH1"/
+            # "NOTCH2" (confirmed live). Mirrors _get_gene_validity's already-
+            # correct exact-match filter above.
             gene_upper = gene.upper()
-            matches = []
-            for c in curations:
-                gene_val = c.get("GENE SYMBOL", c.get("Gene Symbol", ""))
-                if gene_upper in gene_val.upper():
-                    matches.append(c)
+            matches = [
+                c
+                for c in curations
+                if c.get("GENE SYMBOL", c.get("Gene Symbol", "")).upper() == gene_upper
+            ]
 
             return {
                 "status": "success",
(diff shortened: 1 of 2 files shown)


gasvn:

PR #364, opened 2026-07-21
Fix Round 73: BRENDA SABIO-RK integration silently swallowed a 2025 upstream API retirement as empty success

gasvn:
## Summary

Round 73 continued the new error-path/boundary-testing methodology. Firing malformed/empty/boundary inputs at `BRENDA_get_enzyme_kinetics` (a never-before-touched tool in this campaign) surfaced a clean, well-scoped bug pattern the methodology was designed to catch: a **silent upstream API retirement masquerading as "no data."**

**BRENDA_get_enzyme_kinetics's SABIO-RK sub-fetch called a dead REST endpoint.** `_fetch_sabiork_kinetics` called `/sabioRestWebServices/searchKineticLaws/entryIDs`, which SABIO-RK retired in 2025 — every request now 302-redirects to a JS SPA 404 page (confirmed live via raw `curl`). The old status-code/XML-parse checks silently swallowed this as `{"kinetic_laws": [], "total_count": 0}` instead of surfacing a failure. Confirmed live for EC 1.1.1.1 (alcohol dehydrogenase, one of the most-studied enzymes in biochemistry): the tool reported **0** SABIO-RK entries when the real count is **768**.

Notably, `sabiork_tool.py`'s own standalone `SABIORKTool` had **already been migrated** to a working Solr-backed endpoint (`https://sabiork.h-its.org/api/ft/proxy-select`) by an earlier fix (its own docstring explains the 2025 retirement) — but `brenda_tool.py`'s separate, parallel SABIO-RK integration was never updated to match. Ported the same working approach rather than reimplementing a second copy of the fix.

The new Solr endpoint doesn't expose raw numeric parameter values (confirmed live — `SABIORKTool`'s own `"parameters"` field is always empty too), so `kinetic_laws` entries now carry real entry/type metadata but no Km/kcat/Ki numeric values. This is **not a regression**: the old code's numeric `parameter_summary` aggregation was already always empty too, since `kinetic_laws` was always `[]`.

**Also fixed 3 pre-existing stale tests** in `tests/tools/test_sabiork_tool.py` that predated `SABIORKTool`'s own Solr migration — they still mocked the retired two-step XML flow and were silently broken (excluded from the default CI sweep via `pytest.ini`'s `--ignore=tests/tools`), since whoever fixed the tool itself didn't update its own tests to match.

## Verification

- Raw `curl` against the legacy endpoint: `302 Found` → `Location: /ui/404`; following the redirect (as `requests` does by default) lands on a 200-OK HTML SPA shell, not real SABIO-RK data.
- Live `tu run BRENDA_get_enzyme_kinetics '{"ec_number": "1.1.1.1"}'` before fix: `sabiork_total_entries: 0`. After fix: `sabiork_total_entries: 768`, `sources: ["ExPASy ENZYME", "SABIO-RK"]`.
- Confirmed `parameter_summary` stays honestly absent both before and after (no fabricated numeric data).
- Regression-checked error paths (malformed EC number, nonempty-but-nonexistent EC number, empty string, no args) — all still correctly distinguish "bad query" from "no data," unaffected by this fix.

## Test plan

- [x] New `tests/unit/test_brenda_sabiork_solr_migration.py`: 7 tests — correct Solr endpoint/query construction, real entry-count reporting, organism filter, non-200/non-JSON responses now raise (reaching the existing `sabiork_error` surfacing mechanism from an earlier fix) instead of silently returning zero, and no fabricated `parameter_summary`
- [x] Existing `tests/unit/test_brenda_sabiork_failure_visibility.py` (3 tests, mocks at the method level) — unaffected, still passes
- [x] Fixed 3 stale tests in `tests/tools/test_sabiork_tool.py` to mock the real Solr response shape instead of the retired XML flow — all 10 tests in that file now pass
- [x] Full suite (`pytest tests/ --ignore=tests/test_claude_code_plugin.py`) — clean, 0 failures

### DIFF
--- a/src/tooluniverse/brenda_tool.py
+++ b/src/tooluniverse/brenda_tool.py
@@ -15,7 +15,6 @@ WSDL: https://www.brenda-enzymes.org/soap/brenda_zeep.wsdl
 
 import hashlib
 import os
-import xml.etree.ElementTree as ET
 from typing import Any, Dict, List, Optional
 
 import requests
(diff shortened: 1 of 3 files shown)


gasvn:

PR #369, opened 2026-07-21
Fix Round 76: dedicated sweep for non-200-silently-swallowed-into-empty-success (Orphanet + AlphaMissense)

gasvn:
## Summary

Per the user's explicit direction, this round was a dedicated codebase-wide grep sweep for the exact signature confirmed twice already (round 73's SABIO-RK fix, round 75's iCite fix): a request-failure helper returning `[]`/`{}`/`None` on any non-200 status or parse failure, without raising or setting `status: error` — indistinguishable from a genuinely empty/absent result.

Grepped for `status_code != 200` (and related failure checks) across `src/tooluniverse/*.py`, filtered out the many already-correct `{"status": "error", ...}` returns, and triaged the remainder by checking each candidate's caller to see whether the failure signal actually reaches the top-level response. Found and fixed two real instances:

**`Orphanet_get_genes`'s `_fetch_genes_for_code`/`_find_subtype_codes`** swallowed every non-200/exception into an empty list, so a failed Orphadata request (rate limit, 5xx, timeout) was reported identically to "this disease genuinely has no known gene associations." Fixed by threading a `(result, failed)` tuple through both helpers so the caller can tell the difference — while carefully preserving the pre-existing, **deliberate** behavior that a 404 from this specific endpoint means confirmed-no-records, not a failure (an existing test, `test_valid_code_with_no_data_still_returns_success_empty`, already required this; an early version of this fix broke it and had to be corrected before shipping — this distinction between real failure and real emptiness is the crux of the whole bug class, and it's just as important not to overcorrect into treating every empty/404 as a failure).

**`AlphaMissense_get_protein_scores`'s per-residue fetch helper (`_fetch_single_residue`)** treated every non-200 response identically, dropping the position from the output with zero indication anything failed — indistinguishable from AlphaMissense simply not covering that residue. Confirmed live that a 404 here means "no data for this residue" (a real, legitimate empty result, e.g. an out-of-range position), while other non-200s or request exceptions are genuine failures. Same fix shape: distinguish confirmed-empty from failure, and surface an honest `n_positions_failed`/`note` when failures occur instead of silently vanishing them. Live-verified during this fix's own testing that 1 of 20 real position requests genuinely failed and is now correctly surfaced instead of silently dropped.

**Recurring meta-pattern, now confirmed a 3rd time**: in each of these files, a *sibling* function already handled this correctly (`sabiork_tool.py`'s already-migrated `SABIORKTool` vs `brenda_tool.py`'s stale copy; `orphanet_tool.py`'s `_get_gene_validity`-style "get" operations vs the "search"-style helpers; `alphamissense_tool.py`'s `_get_residue_scores` vs the `_fetch_single_residue` helper used by `get_protein_scores`). Worth grepping for "the same operation implemented twice in one file, compare their error handling" as a search heuristic in future sweeps.

**Not included**: `pathwaycommons_tool.py`'s `_traverse` helper, a related but lower-severity variant (secondary fields conflate a failed fetch with genuinely-empty via `x or []` fallbacks) flagged in round 75 — deferred, needs more schema-contract thought before fixing.

## Verification

- Orphanet: live `tu run Orphanet_get_genes '{"orpha_code": "558"}'` (Marfan syndrome, a known parent-code case requiring the subtype-fallback search) still works correctly end-to-end after the fix. Confirmed the 404-preserving behavior via the existing test suite, which caught my initial fix attempt breaking it.
- AlphaMissense: raw `curl` against the live API confirms 404 with a clear "no data for this residue" message for an out-of-range position (`resi=99999`), vs. a real score for a valid position — confirming 404 is the correct "confirmed empty" signal to preserve. Live `tu run AlphaMissense_get_protein_scores` during testing surfaced a real `n_positions_failed: 1` for a genuine live failure.

## Test plan

- [x] New `tests/unit/test_orphanet_gene_fetch_failure_visibility.py`: 5 tests — genuine empty stays success, a genuine request failure is now an error (not confirmed-empty), successful lookup unaffected, and the helper-level `(genes, failed)` tuple behaves correctly for both cases
- [x] New `tests/unit/test_alphamissense_failure_visibility.py`: 7 tests — 404 not marked failed, 500/timeout marked failed, partial failure surfaced (not silently dropped), all-genuinely-empty vs all-genuinely-failed produce distinct error messages, and no failure fields when nothing fails (backward compatibility)
- [x] All existing Orphanet tests (32 total across `test_orphanet_invalid_code_validation.py`, `test_orphanet_classification_hierarchy.py`, `test_orphanet_gene_name_slash_encoding.py`, `tests/tools/test_orphanet_tool.py`) — pass, no regressions (this is what caught and forced the correction of my initial overcorrection)
- [x] All existing AlphaMissense tests (14 total, live integration) — pass, no regressions
- [x] Full suite (`pytest tests/ --ignore=tests/test_claude_code_plugin.py`) — clean, 0 failures

### DIFF
--- a/src/tooluniverse/alphamissense_tool.py
+++ b/src/tooluniverse/alphamissense_tool.py
@@ -89,19 +89,33 @@ class AlphaMissenseTool(BaseTool):
 
     def _fetch_single_residue(
         self, uniprot_id: str, position: int
-    ) -> Optional[Dict[str, Any]]:
-        """Fetch one residue's AlphaMissense scores; returns None on error."""
+    ) -> "tuple[Optional[Dict[str, Any]], bool]":
+        """Fetch one residue's AlphaMissense scores.
+
+        Fix-R76A-2: returns (data, failed) rather than just data. Confirmed
+        live the API 404s with a clear "no data for this residue" message
+        for a position outside its actual coverage (e.g. resi=99999 on a
+        real protein) -- that's a legitimate empty result, not a failure.
+        Only a non-404 non-200 status or a request exception is a genuine
+        failure. The caller previously dropped ALL non-200 positions from
+        the output identically (`if s is not None`), so a batch of
+        genuinely-failed positions (rate limit, timeout) vanished from the
+        response with zero indication anything went wrong, indistinguishable
+        from AlphaMissense simply not covering those residues.
+        """
         try:
             r = requests.get(
                 f"{ALPHAMISSENSE_BASE_URL}/hotspotapi",
                 params={"uid": uniprot_id, "resi": position},
                 timeout=self.timeout,
             )
+            if r.status_code == 404:
+                return None, False
             if r.status_code != 200:
-                return None
-            return r.json()
+                return None, True
+            return r.json(), False
         except requests.exceptions.RequestException:
-            return None
+            return None, True
 
     def _get_protein_scores(self, arguments: Dict[str, Any]) -> Dict[str, Any]:
         """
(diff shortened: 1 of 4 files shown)


gasvn:

PR #372, opened 2026-07-21
Fix Round 77: iCite double-wrap + overly-strict null schema types

gasvn:
## Summary

Found via a schema-conformance sweep across all 638 tool config patterns (`scripts/test_all_tools.py --parallel`), 18 patterns flagged with schema mismatches. Triaged each individually with live `jsonschema.validate()` checks against both the wrapped and unwrapped output.

**One genuine bug**, `iCite_get_publications` (a plain `BaseRESTTool` hitting the raw iCite API directly): the raw API's own JSON body is already shaped `{"data": [...], ...}`. Without `fields.extract_path`, `BaseRESTTool._process_response` wrapped that whole raw body under another top-level `data` key, producing `{"data": {"data": [...], ...}}` — violating the declared `return_schema`, which requires `data` to be a bare array. Fixed by adding `extract_path: "data"`.

**Three overly-strict schema types**, all confirmed via live data where a field is legitimately `null`:
- `OmicsDI_search_datasets`: `description`, `publicationDate` — not every aggregated repository populates these (sibling field `organisms` already allowed null).
- `DailyMed_search_spls`: `metadata.next_page_url` / `previous_page_url` — `SearchSPLTool` already normalizes these to `None` on the first/last page; the schema hadn't caught up.
- `PDBe_get_compound_details`: `cross_links` — a compound with no external database cross-references.

**Methodology finding**: the other 14 flagged patterns (artic, bmrb, cdc, dailymed's search_drug_classes, encode, eurostat, fda_gsrs, gbif/gbif_ext, gtex/gtex_v2, idr, nasa_sbdb, openfoodfacts, clinical_guidelines's MAGICapp) are **false positives** of `scripts/test_new_tools.py`'s own `_schema_describes_envelope` heuristic, which misclassifies inner-data schemas as envelope schemas whenever `data`/`status` happens to be a domain field name in the underlying API's own response shape (not a description of our own `{status,data,...}` wrapper). Confirmed for each via direct `jsonschema.validate()` against both interpretations — the unwrapped `result["data"]` validates cleanly in every case. None of those tools were touched. This heuristic limitation is left as a documented follow-up rather than a rushed fix to shared test infrastructure (a naive "valid if either interpretation passes" fix would mask real bugs like this PR's iCite one).

## Test plan
- [x] New tests: `tests/unit/test_icite_get_publications_extract_path.py` (4 tests — config declares extract_path, `_process_response` unwraps correctly, unwrapped result satisfies return_schema, and a sanity check that removing extract_path reproduces the original double-wrap bug)
- [x] New tests: `tests/unit/test_round77_null_permissive_schema_fields.py` (4 tests — one per widened field, plus a negative case confirming `description` still rejects non-string/non-null)
- [x] Existing dailymed/omicsdi regression suites (29 tests across 8 files) re-run clean
- [x] `scripts/test_new_tools.py icite/omicsdi/dailymed/pdbe_graph` all show 0 schema-invalid after the fix

### DIFF
--- a/src/tooluniverse/data/dailymed_tools.json
+++ b/src/tooluniverse/data/dailymed_tools.json
@@ -76,7 +76,7 @@
               "type": "string"
             },
             "next_page_url": {
-              "type": "string"
+              "type": ["string", "null"]
             },
             "total_elements": {
               "type": "integer"
@@ -91,7 +91,7 @@
               "type": ["string", "integer", "null"]
             },
             "previous_page_url": {
-              "type": "string"
+              "type": ["string", "null"]
             },
             "next_page": {
               "type": ["string", "integer", "null"]
--- a/src/tooluniverse/data/icite_tools.json
+++ b/src/tooluniverse/data/icite_tools.json
@@ -15,7 +15,8 @@
       ]
     },
     "fields": {
-      "endpoint": "https://icite.od.nih.gov/api/pubs"
+      "endpoint": "https://icite.od.nih.gov/api/pubs",
+      "extract_path": "data"
     },
     "type": "BaseRESTTool",
     "test_examples": [
--- a/src/tooluniverse/data/omicsdi_tools.json
+++ b/src/tooluniverse/data/omicsdi_tools.json
@@ -106,13 +106,13 @@
                     "type": "string"
                   },
                   "description": {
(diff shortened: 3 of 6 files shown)


gasvn:

PR #373, opened 2026-07-21
Fix Round 78: two well-known entities silently absent from cross-tool results

gasvn:
## Summary

Found via a new methodology angle for this campaign: **cross-tool consistency** — querying the same well-known entity (a gene, a variant) through every tool that should know about it and comparing results, rather than testing each tool in isolation. Both findings share the same underlying bug class: **a well-known entity silently absent from a tool's results** — invisible to single-tool happy-path testing, but obviously wrong to anyone who knows the entity. Worth searching for again on different genes/variants/drugs in future rounds, not just the specific query shapes below.

1. **`MARRVEL_get_gene` exposed a bogus `omim_id` field.** MARRVEL's own `/gene/taxonId/.../symbol/...` endpoint returns a tiny placeholder integer (`"1"` for BRCA1/EGFR/APOE/MYH7/TP53, `"6"` for CFTR/PTEN) instead of the real 6-digit OMIM MIM number — confirmed live across 7 genes, so this isn't a per-gene data gap in MARRVEL's own database, it's the field itself being broken in that endpoint. The dedicated `MARRVEL_get_omim_phenotypes` tool already returns the correct `gene_mim_number` (191170 for TP53, matching HGNC's own record). Removed the misleading field from `MARRVEL_get_gene` and pointed its description at the working tool instead, rather than presenting wrong data as if genuine.

2. **`annotate_variant_multi_source` always reported "no exact ClinVar match" for rs334** — the sickle-cell HBB mutation, arguably the single most-studied point mutation in human genetics — despite ClinVar having a well-established Pathogenic record for it (VCV 15333, `NM_000518.5(HBB):c.20A>T (p.Glu7Val)`). Root cause: `ClinVar_search_variants` had no way to search by protein change/HGVS notation, only `gene`/`condition`/`clinical_significance`, so an exact-match lookup for one specific variant could only browse gene-level rows (capped at 100) and filter client-side. HBB has 1962 total ClinVar entries (still 425 after filtering to Pathogenic-only), and the sickle-cell record's low/old variant ID meant it was never within any of those capped fetches, regardless of filter.

   ClinVar's own eSearch supports a real `[Variant name]` field tag (confirmed live via raw NCBI E-utils: `HBB[gene] AND Glu7Val[Variant name]` finds VCV 15333 directly, in 9 total hits vs. browsing ~2000 gene-level rows). Added it as a proper `variant_name` parameter on `ClinVar_search_variants` (accepts a single name or a list, OR-combined — needed since a multi-allelic rsid site yields one candidate protein change per allele). Wired `annotate_variant_multi_source` to try a targeted `variant_name` search first when a candidate token is resolvable, falling back to the old gene-level browse only when that doesn't confirm a match (preserves existing behavior for cases with no resolvable token, or where the token's nomenclature isn't in ClinVar's index).

## Test plan
- [x] New tests: `tests/unit/test_clinvar_variant_name_search.py` (5 tests — single string, list/OR-combination, variant_name alone with no gene, correctly recognized as a valid param, blank entries dropped)
- [x] New tests in `tests/unit/test_compound_variant_tool.py` (3 tests — targeted search used when it confirms a match and the bulk fallback is *not* called, falls back to bulk browse when the targeted search finds nothing, and unchanged behavior when no variant_token is resolvable at all)
- [x] Updated 2 stale tests in `tests/unit/test_clinvar_gene_hyphen_and_params.py` that asserted the *old* "variant_name doesn't exist" behavior (from an earlier round's Fix-R5D-1, which used `variant_name` as its own illustrative example of an unrecognized parameter)
- [x] Updated `tests/unit/test_discovery_resource_tools_batch2.py`'s MARRVEL test to stop asserting on the removed `omim_id` field, added a dedicated regression test confirming it's never surfaced even when present in the mocked upstream response
- [x] All 83 clinvar/compound_variant unit tests pass
- [x] Live end-to-end re-verification: `annotate_variant_multi_source({"rsid": "rs334"})` now reports `clinvar_classification: "Pathogenic"`; `MARRVEL_get_gene({"symbol": "TP53"})` no longer includes `omim_id`
- [x] `scripts/test_new_tools.py` clean (0 schema-invalid) for clinvar, compound_variant, and marrvel patterns

### DIFF
--- a/src/tooluniverse/clinvar_tool.py
+++ b/src/tooluniverse/clinvar_tool.py
@@ -240,6 +240,32 @@ class ClinVarSearchVariants(ClinVarRESTTool):
             # Use [uid] to look up by numeric variation ID.
             query_parts.append(f"{arguments['variant_id']}[uid]")
 
+        if arguments.get("variant_name"):
+            # Fix-R78B-1: the Fix-R5D-1/R8C-1 comment above assumed there was
+            # "no such parameter" -- but ClinVar eSearch DOES support a real
+            # [Variant name] field tag (confirmed live: "HBB[gene] AND
+            # Glu7Val[Variant name]" -> 9 hits including VCV 15333, the
+            # canonical sickle-cell NM_000518.5(HBB):c.20A>T (p.Glu7Val)
+            # Pathogenic record). Without this, an exact-match lookup for a
+            # specific protein change/HGVS notation could only be done by
+            # fetching gene-level rows (capped at 100) and filtering
+            # client-side -- which silently misses old/low-ID records in any
+            # variant-dense gene, since ClinVar's default gene-level order
+            # isn't clinical-significance- or fame-ranked (confirmed live:
+            # HBB's own sickle-cell record, one of the best-known variants in
+            # human genetics, isn't within the first 100 or even the first
+            # 425-Pathogenic-filtered gene-level rows). Accepts either a
+            # single name or a list (a multi-allelic site yields one protein
+            # change per allele; only the queried allele's name will hit, so
+            # every candidate must be tried) -- combined with OR so one
+            # request covers all candidates.
+            names = arguments["variant_name"]
+            names = names if isinstance(names, list) else [names]
+            names = [str(n).strip() for n in names if str(n).strip()]
+            if names:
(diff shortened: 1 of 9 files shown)


gasvn:

PR #374, opened 2026-07-21
Fix Round 79: CPIC_get_recommendations silently ignored gene/phenotype filters

gasvn:
## Summary

Continues round 78's cross-tool-consistency work, this round in a new tool family. Found via a pharmacogenomics cross-check (CYP2C19/clopidogrel, an FDA boxed-warning gene-drug pair): `CPIC_get_recommendations({"drug": "clopidogrel", "gene": "CYP2C19", "phenotype": "Poor Metabolizer"})` returned all 24 recommendation rows across every phenotype and population combination for the guideline, completely unfiltered. Neither `gene` nor `phenotype` was a real parameter — the schema has no `additionalProperties: false`, so both were silently accepted and dropped with no indication either had zero effect.

This is a clean "declared vs. actual" case: the tool's own description explicitly promises to link *"genotype/phenotype to prescribing actions,"* and every returned row already carries a `phenotypes` dict keyed by gene symbol — a caller very naturally guesses these param names and gets silently unfiltered results back instead of an error or a working filter.

Added real client-side filtering:
- **Exact case-insensitive match on the phenotype value, not substring.** `"Poor Metabolizer"` and `"Likely Poor Metabolizer"` are distinct CPIC categories with different clinical recommendations, not synonyms — a substring match would have silently merged two clinically different recommendations into one result, which is the same silent-wrongness shape as round 78's MARRVEL/ClinVar findings, just at the string-matching layer instead of the missing-capability layer.
- `limit`/`offset` now apply *after* filtering rather than before, by fetching a larger batch server-side whenever a filter is active — so pagination is over the filtered set a caller actually asked for, not an arbitrary raw slice of the unfiltered table.
- A no-match result now includes a note listing the guideline's actual available phenotype values (e.g. `Poor Metabolizer`, `Likely Poor Metabolizer`, `Normal Metabolizer`, ...) instead of a bare empty list — turns "did I typo the phenotype string?" into something answerable from the response itself.

Also verified round 78's ClinVar exact-match fix (still-unmerged PR #373) generalizes correctly to a different gene/variant — APOE/rs429358, the variant defining the Alzheimer's-risk APOE4 allele — before looking for new bugs this round. Confirming a fix holds on a fresh entity, not just the one that surfaced it, seemed worth doing before moving on.

## Test plan
- [x] New test file: `tests/unit/test_cpic_recommendations_gene_phenotype_filter.py` (6 tests — gene+phenotype exact match, phenotype-only across any gene, gene-only across all its phenotypes, no-match note with real available values, limit/offset applied after filtering, and unchanged behavior when no filter is given)
- [x] Existing CPIC tests (17 across `test_cpic_brand_name_error.py`, `test_cpic_get_gene_info_aliases.py`) pass unchanged
- [x] Live re-verification: `CPIC_get_recommendations({"drug":"clopidogrel","gene":"CYP2C19","phenotype":"Poor Metabolizer"})` now returns exactly 3 matching rows instead of all 24
- [x] `scripts/test_new_tools.py cpic` clean (16/16 schema valid)

### DIFF
--- a/src/tooluniverse/cpic_search_pairs_tool.py
+++ b/src/tooluniverse/cpic_search_pairs_tool.py
@@ -117,14 +117,35 @@ class CPICGetRecommendationsTool(BaseTool):
 
         limit = arguments.get("limit", 50) or 50
         offset = arguments.get("offset", 0) or 0
+        # Fix-R79A-1: "gene"/"phenotype" are not real params (no
+        # additionalProperties: false on this schema, so they were silently
+        # accepted and dropped, not rejected) -- confirmed live that
+        # {"drug": "clopidogrel", "gene": "CYP2C19", "phenotype": "Poor
+        # Metabolizer"} returned all 24 rows across every phenotype and
+        # population/context combination for the guideline, unfiltered, with
+        # no indication either filter had zero effect. A caller very
+        # naturally guesses these param names, since the tool's own
+        # description says it "links genotype/phenotype to prescribing
+        # actions" and every returned row already carries a `phenotypes`
+        # dict keyed by gene symbol -- add real client-side filtering rather
+        # than just erroring more clearly on the unrecognized names, since
+        # the data to filter on is already present in every response.
+        gene_filter = arguments.get("gene")
+        phenotype_filter = arguments.get("phenotype")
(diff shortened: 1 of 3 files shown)


gasvn:

PR #375, opened 2026-07-21
Fix Round 80: gather_gene_disease_associations fabricated ClinVar diseases

gasvn:
## Summary

Continues the round-78/79 cross-tool-consistency thread, this round via the `gather_gene_disease_associations` aggregator (queried LDLR / familial hypercholesterolemia). This one is a more severe flavor of the bug class the last two rounds found: rather than silently omitting or mislabeling something, the tool **fabricated entries a researcher could plausibly cite** — reporting `LDLR` (the gene itself) and `LDLR-AS1` (a co-located antisense RNA gene) as *diseases that LDLR causes*.

Root cause, two layers:

1. **Wrong ClinVar query parameter.** The sub-call used `{"query": gene_or_disease}` — `"query"` is documented as an alias for `"condition"` (a disease/phenotype free-text search), not a gene lookup. This is the exact same mistake already caught and fixed in the sibling `compound_variant_tool.py` (Fix-R31D-4/R31A-3) — it was just never ported to this file, even though this file's own DisGeNET and OpenTargets sections already correctly branch gene-vs-disease queries a few lines above the ClinVar one.
2. **Extraction fabricated disease names from gene symbols.** Confirmed live that `ClinVar_search_variants`' row shape (`title`/`genes`/`clinical_significance`/`review_status`) has **no disease/condition name field at all** — so the old code's "pull each row's `genes` field into the disease-name list" wasn't just mislabeled, it had no correct behavior available to fall back to.

Fixed by:
- Branching the ClinVar query the same way DisGeNET/OpenTargets already do in this file: `gene` → `{"gene": gene, "limit": 10}`, `disease` → `{"condition": disease, "limit": 10}`.
- Having the extraction correctly contribute **nothing** from ClinVar, rather than fabricate wrong entries, since there's genuinely no usable field to extract a disease name from.
- Adding a `notes` field, populated only when ClinVar succeeds but has nothing to contribute — explicitly distinct from `sources_failed` (a genuine ClinVar outage), so a caller can tell "structurally can't help here" apart from "this particular call failed."

**Worth flagging as its own class going forward**: silently-missing data at least fails toward under-claiming; fabricated associations fail toward confident nonsense. And the wrong-parameter half of this bug is itself an instance of a fix landing in one tool while an identical copy of the bug persisted in a sibling tool — worth a targeted sweep sometime: when a fix lands, grep sibling tools sharing the same call pattern to check whether they need it too.

## Test plan
- [x] New test file: `tests/unit/test_compound_gene_disease_clinvar_gene_names.py` (6 tests — extraction returns empty for ClinVar's actual response shape, gene symbols never leak into `associations` end-to-end, gene query uses `gene` not `query`, disease query uses `condition`, the limitation note appears on success-but-empty, and does *not* appear on a genuine ClinVar failure)
- [x] Existing regression test `tests/unit/test_compound_gene_disease_opentargets_lookup.py` (9 tests) passes unchanged
- [x] Live end-to-end re-verification of both branches: `gather_gene_disease_associations({"gene": "LDLR"})` no longer includes `LDLR`/`LDLR-AS1` in `associations`; `gather_gene_disease_associations({"disease": "cystic fibrosis"})` runs cleanly through the `condition` path
- [x] `scripts/test_new_tools.py compound_gene_disease` clean (3/3 schema valid)

### DIFF
--- a/src/tooluniverse/compound_gene_disease_tool.py
+++ b/src/tooluniverse/compound_gene_disease_tool.py
@@ -93,28 +93,55 @@ class CompoundGeneDiseaseAssociationTool(BaseTool):
             r = _try_tool("GenCC", "GenCC_search_disease", {"disease": disease})
         results_by_source["GenCC"] = self._extract_genes_or_diseases(r, "GenCC")
 
-        # 5. ClinVar
-        r = _try_tool(
-            "ClinVar",
-            "ClinVar_search_variants",
-            {"query": gene or disease, "limit": 10},
-        )
+        # 5. ClinVar -- Fix-R80A-1: "query" is documented as an alias for
+        # "condition" (a disease/phenotype free-text search), not a gene
+        # lookup -- the exact same mistake already caught and fixed in the
+        # sibling compound_variant_tool.py (Fix-R31D-4/R31A-3), but never
+        # ported here. Confirmed live that {"query": "LDLR"} silently
+        # returned ClinVar rows for LDLR-related variants anyway (a
+        # coincidental partial match, not a deliberate gene lookup) whose
+        # `genes` field then got mislabeled as "disease" entries below (see
+        # that fix for why genes can't be used as disease names at all).
+        # Branch on gene/disease the same way DisGeNET/OpenTargets already do
+        # above.
+        if disease:
+            r = _try_tool(
+                "ClinVar",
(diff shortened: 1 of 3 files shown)


gasvn:

PR #391, opened 2026-07-24
Fix Round 140: honor HPO_search_terms max_results and Enrichr empty-libs default

gasvn:
## Summary

Post-1.4.0 role-play bug-hunt round targeting the 436 tool wrappers newly exposed in #389 (previously registered-but-unwrapped tools that had never been through a role-play). Three researcher personas (rare-disease geneticist, clinical pharmacologist, functional genomicist) exercised ~64 tool calls across ~30 families via the CLI against the 1.4.0 code. Every reported issue was CLI-reproduced before fixing; false positives and upstream-outage artifacts were discarded.

**Net: 3 confirmed defects fixed** (1 HIGH, 2 MEDIUM), all in the "a valid input silently returns empty/wrong data as a success" class, plus a robustness hardening. The remaining ~30 families returned correct, parameter-honoring data.

### 1. `HPO_search_terms` silently ignored `max_results` (HIGH)
`hpo_tool.py` sent the page-size to the JAX ontology search endpoint as `max`, but that endpoint sizes pages with `limit`. The `max` key was silently ignored, so **every** search was capped at the API default of 10 regardless of the requested `max_results`, and `metadata.total_results` reported 10 even when 100 matches existed — a clinician would believe only 10 phenotype terms exist for a query.

CLI verification (live JAX API):
```
{'q':'seizure','max':3}  -> 10 terms   (ignored)
{'q':'seizure','limit':3} -> 3 terms    (honored)
{'q':'seizure','limit':50}-> 50 terms
totalCount = 100 in all cases
```
Fix: send `limit`; coerce `max_results` robustly (null / 0 / negative / over-max no longer crash or leak through — the schema allows `integer|null`); truncate defensively; and surface the true total as `metadata.total_available`.

After fix: `max_results` 3→3, 20→20, 50→50; default→10; `null`→10 (no crash); 999→50; -5→1; `total_available`=100.

### 2. `enrichr_gene_enrichment_analysis` empty `libs` returned empty as success (MEDIUM)
`libs = arguments.get("libs", DEFAULT)` — passing `libs: []` (the tool's **own documented example**) returned `[]`, so zero libraries were queried and an empty enrichment was wrapped in `status: success`. Fix: empty or omitted `libs` now falls back to the default library set (`arguments.get("libs") or DEFAULT_LIBS`), and the documented example uses a real library.

Before: documented example → `connected_paths: {}` (empty), only trivial self-loops.
After: documented example → 20 ranked connectivity paths from the 5 default libraries.

### 3. `enrichr_gene_enrichment_analysis` returned double-JSON-encoded `data` (MEDIUM)
`data` was `json.dumps(result)` — a JSON **string**, forcing every consumer to `json.loads` a second time, inconsistent with every other ToolUniverse tool; `return_schema` was merely `{"type": "string"}`, which does not describe the `{status, data}` envelope. Fix: return `data` as a structured object with a `metadata` block, and update `return_schema` to describe it (verified it validates via `jsonschema.validate(result["data"], return_schema)`, matching how `cli.py` checks it).

### 4. Enrichr request timeouts (robustness)
The `requests.get/post` calls in `enrichr_tool.py` passed no `timeout=`; a persona observed one run stall past 140s while an identical later run finished in ~8s. Added timeouts so a stalled upstream cannot hang indefinitely.

## Tests
- `tests/unit/test_hpo_search_terms_limit.py` — 5 tests: `limit` (not `max`) is sent, results honor/truncate to `max_results`, `total_available` surfaces `totalCount`, `null` defaults without crashing, bounds are clamped.
- `tests/unit/test_enrichr_empty_libs.py` — 5 tests: empty/omitted `libs` fall back to default, explicit libs pass through, `data` is a structured object, metadata reports libraries + gene count.

All 10 new tests pass; the existing HPO/Enrichr unit suites (35 tests across `test_disease_ontology_depth`, `test_hpo_phenotype_annotations`, `test_functional_enrichment_depth`, `test_enrichr_no_stdout_pollution`, `test_monarch_result_id_prefix`, `test_hpo_id_normalization`) still pass — no regressions.

## Verified false positives / not fixed (for the record)
- **Inconsistent gene-symbol param names** across families (`gene` / `gene_symbol` / `symbol`) — friction only; errors are clear and actionable. Cross-family aliasing is an anti-pattern (grows forever); left as a separate design-consistency concern.
- **EnsemblPheno / Pharos / FAERS / FDA-label errors** the personas hit were confirmed **upstream outages** (`api.fda.gov` no-route, `pharos-api.ncats.io` 502, Ensembl REST timeout), not tool defects — the tools surfaced clean errors.
- **EVA / GTEx / ClinVar / GBIF / ClinGen / PanglaoDB / ReactomeAnalysis empties** were confirmed **genuine upstream sparseness** (verified against the raw APIs / source CSVs), not silent failures.
- A `FAERS_count_*` error-as-data-row pattern and an Enrichr no-p-values output-format concern were noted but only reproduce under upstream outage / are format redesigns — deferred rather than fixed unverified.

### DIFF
--- a/src/tooluniverse/data/enrichr_tools.json
+++ b/src/tooluniverse/data/enrichr_tools.json
@@ -19,7 +19,7 @@
           "items": {
             "type": "string"
           },
-          "description": "List of enrichment libraries to use for analysis.",
+          "description": "List of Enrichr enrichment libraries to query (e.g. 'KEGG_2021_Human', 'GO_Biological_Process_2023'). Omitting this or passing an empty list uses a broad default set of pathway/ontology libraries.",
           "default": [
             "WikiPathways_2024_Human",
             "Reactome_Pathways_2024",
@@ -40,11 +40,26 @@
           "BRCA1",
           "EGFR"
         ],
-        "libs": []
+        "libs": [
+          "KEGG_2021_Human"
+        ]
       }
     ],
     "return_schema": {
-      "type": "string"
+      "type": [
+        "object",
+        "null"
+      ],
+      "properties": {
+        "connected_paths": {
+          "type": "object",
+          "description": "Ranked weighted paths between the first two genes routed through shared enrichment terms."
+        },
+        "connections": {
+          "type": "object",
+          "description": "Per gene-to-term connectivity paths found in the enrichment graph."
+        }
+      }
     }
   }
 ]
--- a/src/tooluniverse/enrichr_tool.py
+++ b/src/tooluniverse/enrichr_tool.py
@@ -15,6 +15,16 @@ class EnrichrTool(BaseTool):
     Tool to perform gene enrichment analysis using Enrichr.
     """
 
+    # Default enrichment libraries used when the caller does not specify any
+    # (or passes an empty list).
+    DEFAULT_LIBS = [
+        "WikiPathways_2024_Human",
+        "Reactome_Pathways_2024",
+        "MSigDB_Hallmark_2020",
+        "GO_Molecular_Function_2023",
+        "GO_Biological_Process_2023",
+    ]
+
     def __init__(self, tool_config):
         super().__init__(tool_config)
         # Constants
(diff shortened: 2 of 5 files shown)


gasvn:
Consolidated into #393; closing to keep the merge to a single PR.

PR #392, opened 2026-07-24
Fix onboarding and install-experience gaps found in a 1.4.0 dogfood

gasvn:
Fixes the papercuts a non-CS user hits between `pip install` and a first working tool call. Each finding was reproduced on the current tree before being fixed.

## 1. The setup guide's first example used the wrong parameter

`skills/setup-tooluniverse/SKILL.md` (and its two mirrors) told users to run:

```bash
tu run FAERS_count_death_related_by_drug '{"drug_name": "metformin"}'
```

The tool's config requires `medicinalproduct`, not `drug_name`. Verified:

```
$ tu run FAERS_count_death_related_by_drug '{"drug_name": "metformin"}'
Error: Parameter validation failed for 'root': 'medicinalproduct' is a required property

$ tu run FAERS_count_death_related_by_drug '{"medicinalproduct": "metformin"}'
{"result": [{"term": "alive", "count": 110696}, {"term": "death", "count": 37537}]}
```

Fixed all 6 occurrences across `skills/`, `plugin/skills/`, and `plugins/tooluniverse/skills/`. This was the scripted "it works!" moment, so it errored for everyone who followed the guide verbatim.

## 2. `tooluniverse-doctor` / `tu status` falsely reported an all-clear

Loading a tool registers its JSON config; it does not install the tool's dependencies. Tools backed by an optional extra were counted as available and only failed at call time.

Before, on a base install with no extras:

```
Total tools: 2599
Available: 2599
Unavailable: 0

All tools loaded successfully!
```

while `ADMETAI_predict_BBB_penetrance` failed with `ADMETModel requires 'admet-ai' package`.

After:

```
Total tools: 2599
Config loaded: 2599
Failed to load: 0

7 optional dependency group(s) not installed:

  [ml] - up to 11 tool(s) may not run
     Missing: admet-ai, sentence-transformers
     Fix: pip install 'tooluniverse[ml]'
  ...
```

New `tooluniverse.extras` module resolves each declared extra against the environment. The all-clear now prints only when nothing failed to load **and** every group is installed. `tu status` shows the same summary; `tu status --json` exposes a `missing_extras` field.

Verified against a real install, not just mocks — installing `biopython` + `freesasa` moved the report from 7 groups to 6, with `[bioinformatics]` dropping out.

Counts are deliberately framed as an upper bound: a few tools import an optional package only as an enhancement (the DNA tools fall back to a built-in restriction-enzyme table without Biopython), so claiming them as hard failures would be wrong.

## 3. Naive `pip install` is a trap on a modern Mac

Both reported symptoms reproduce on a current macOS box (Homebrew Python 3.14.6):

- `python3 -m venv` → `ensurepip ... returned non-zero exit status 1`
- `pip install` → PEP 668 `externally-managed-environment`

`uv` is now an explicit, loud prerequisite on the SDK/Python-developer install line in both the README and the setup skill, with `uv venv --python 3.12` shown (plain `uv pip install` also fails with no active virtualenv). Added troubleshooting entries for both errors plus the "loads but fails when run" case.

## 4. The "no-code" chat path led with hand-editing JSON

Reframed so the two genuinely low-friction paths come first — asking an existing AI agent to read `setup.md`, and the Claude Code plugin one-liner — with manual JSON demoted to a documented fallback. The manual path is kept, now with a copy-paste-safe snippet and a `python3 -m json.tool` validation step, since "JSON syntax error" is a top reported failure.

## 5. Base-install bloat — documented, not changed

`markitdown[all]` does pull in ~32 MB (SpeechRecognition, youtube-transcript-api, pydub, azure-ai-documentintelligence — all confirmed present after a base install).

**Deliberately not narrowed.** `convert_to_markdown` accepts an arbitrary `http:`/`https:`/`file:`/`data:` URI with **no format allowlist**, so every converter dropped from `[all]` becomes a silent runtime failure for that input type rather than a clean error. `markitdown` is also imported unconditionally at module scope by `unified_guideline_tools.py`, so it cannot move to an extra without guarding that import first. Recorded both constraints as a comment in `pyproject.toml` so the next person has the context. Narrowing is a real opportunity but needs its own change with format coverage tests.

Separately added a note that PyTorch Lightning warnings during ADMET calls are normal and not failures.

## Minor

- `[all]` is a misnomer — it excludes `singlecell`, `smolagents`, `client`, and `build`. Noted wherever `[all]` is recommended, and a test asserts the exclusion list stays accurate against `pyproject.toml`.
- Reconciled README's `["--refresh", "tooluniverse"]` against the setup skill's `["tooluniverse"]` by documenting the tradeoff (always-latest but slower to start vs. cached but possibly stale) in both places.

## Testing

- New `tests/unit/test_extras.py` (21 tests) covers missing-extra detection, the tool-counting scan, and the doctor output contract — including the specific regression that a partial install must not print an all-clear. It also keeps the extras mapping in sync with `pyproject.toml`.
- Full `tests/unit` suite passes (exit 0, 19 pre-existing skips).
- `tests/test_claude_code_plugin.py` has the same 6 failures before and after this branch — confirmed by re-running against a clean `HEAD`. None are introduced here.
- All three SKILL.md trees were updated with frontmatter preserved byte-for-byte and re-validated with PyYAML.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,16 +37,30 @@ Add to your MCP config file:
 }
 ```
 
+`--refresh` checks PyPI for the newest release on every launch. Drop it
+(`"args": ["tooluniverse"]`) to start faster from `uv`'s cache — then upgrade
+with `uv cache clean tooluniverse`.
+
 Install agent skills:
 ```bash
 npx skills add mims-harvard/ToolUniverse
 ```
 </details>
 
-**Python developers** — install the SDK:
+**Claude Code users** — one line, no config file:
 ```bash
+claude plugin marketplace add mims-harvard/ToolUniverse
+claude plugin install tooluniverse@tooluniverse
+```
+
+**Python developers** — install the SDK. Install [`uv`](https://docs.astral.sh/uv/) first and **do not use system `pip`**: on a current Mac, `pip install tooluniverse` fails with `externally-managed-environment` (PEP 668) and `python3 -m venv` can fail at `ensurepip`. `uv` manages its own Python and avoids both.
+```bash
+curl -LsSf https://astral.sh/uv/install.sh | sh   # if you don't have uv
+uv venv --python 3.12 && source .venv/bin/activate
 uv pip install tooluniverse
 ```
+
+The base install covers the API and database tools. Local ML, cheminformatics, and plotting tools need extras — `uv pip install 'tooluniverse[all]'`, or a single group such as `[ml]`, `[visualization]`, `[bioinformatics]`. Note `[all]` excludes `singlecell`, `smolagents`, `client`, and `build`, which install by name. Run `tooluniverse-doctor` to see which groups are missing.
 
 **[`tu` CLI](https://zitniklab.hms.harvard.edu/ToolUniverse/guide/tu_cli.html)** — discover, inspect, run, and test tools from the terminal.
 **[Python SDK](https://zitniklab.hms.harvard.edu/ToolUniverse/guide/python_guide.html)** — programmatic access for building AI scientist systems.
--- a/docs/help/troubleshooting.rst
+++ b/docs/help/troubleshooting.rst
@@ -39,6 +39,58 @@ Run this diagnostic script to check your ToolUniverse installation:
 Installation Issues
 -------------------
 
+.. tip::
+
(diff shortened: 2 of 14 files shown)


gasvn:
Consolidated into #393; closing to keep the merge to a single PR.

PR #395, opened 2026-07-25
Fix Round 140 R2: five newly-wrapped tools returning empty/wrong data as success

gasvn:
## Summary

Round 2 of the post-1.4.0 role-play bug-hunt, branched off the current main (which already carries round 1 via #393). Three fresh researcher personas (precision-oncology genomicist, structural biologist / immunoinformatics, metabolomics / lipidomics chemist) exercised ~60 tool calls across a different slice of the #389-added surface — MaveDB, GDC, GenomeNexus, InterPro, ELM, IPD/IMGT, MassBank, SwissLipids, LipidMaps, FDA GSRS, MassIVE, and more. Every reported issue was CLI-reproduced (and cross-checked against the live upstream APIs) before fixing; false positives and upstream-outage artifacts were discarded.

**Net: 5 defects fixed (3 HIGH, 2 MEDIUM)**, all in the "a valid input silently returns empty/wrong data as a success" class.

### 1. `IMGT_get_sequence` returned an error string as the sequence (HIGH)
Queried EBI DBFetch with `db=imgt`, which is not a valid DBFetch database. DBFetch answers an unknown database with **HTTP 200 + body `ERROR 1 Unknown database [imgt].`**, and the status-code-only check returned that error text as the `sequence` under `status:success` — the documented example `M12950` was 100% broken. Fixed by querying the real `imgtligm` (IMGT/LIGM-DB) database, detecting DBFetch's HTTP-200 `ERROR ...` bodies so the EMBL fallback actually fires, and returning a structured error on a genuine miss. CLI-verified: `M12950` → real sequence; bogus accession → `Error: Sequence not found`.

### 2. `MassIVE_get_protein_identifications` ignored the dataset filter (HIGH)
The MassIVE ProXI `/proteins` **and** `/psms` endpoints ignore the dataset `accession` filter — `PXD000561`, `PXD000865`, and even the nonexistent `PXD999999` all returned byte-identical global data (100 mouse proteins) as success. CLI + curl verified the endpoint returns the same result with or without `accession`. Fixed by requiring `protein_accession` (the filter the endpoint actually honors), failing closed on dataset-only queries with a clear message, and correcting the misleading description/parameter/examples.

### 3. `FDAGSRS_get_substance` / `FDAGSRS_search_substances` dropped cross-refs & synonyms (HIGH)
Read `codes`/`names` without `?view=full`; GSRS returns only link stubs in the default view, so cross-references and synonyms — the primary value of a GSRS lookup — came back empty for every substance. Live GSRS `?view=full` for caffeine has **45 codes / 59 names**. Fixed by requesting the full view on both the single-substance and search paths and reading the InChIKey from `_inchiKey`. CLI-verified: caffeine → codes#45, xrefs#33, InChIKey `RYYVLZVUVIJVGH-UHFFFAOYSA-N` (all previously empty).

### 4. `MaveDB_search_score_sets` understated totals and hid the target gene (MEDIUM)
`total_results` was computed from the post-`limit` slice (5 shown when 62 exist), and no target-gene field was exposed — so a query like `BRCA1` that full-text-matches "BRCA1-Associated Protein 1"/**BAP1** looked on-target. Fixed by counting before truncation, adding `returned_results`, and surfacing `target_genes` per row. CLI-verified: `BRCA1` → `total_results=62`, and the top rows now visibly show `target_genes=['BAP1']`.

### 5. `MaveDB_get_mapped_variants` over-counted genomic locations (MEDIUM)
Counted any non-null `postMapped` as `n_with_genomic_location`; protein/transcript-anchored (`NP_`/`NM_`) mappings — common for protein-level DMS score sets — have no genomic coordinates. Fixed to count only chromosome/gene-region references (`NC_`/`NT_`/`NW_`/`NG_`). CLI-verified: `urn:mavedb:00000115-a-7` (all `NP_203524.1`) → `n_with_genomic_location` `3553 → 0`, consistent with `n_with_clingen_allele_id=0`.

## Tests
- New: `tests/unit/test_imgt_get_sequence_dbfetch.py`, `tests/unit/test_fda_gsrs_full_view.py`, `tests/unit/test_mavedb_search_genomic.py`.
- Updated `tests/unit/test_proteomics_depth.py` — the MassIVE tests that encoded the old (buggy) per-dataset behavior now assert the fail-closed error and drive the parse tests via `protein_accession`.
- **89 tests pass** across the new tests plus the existing IMGT / MaveDB / MassIVE / FDAGSRS suites — no regressions.

## Verified false positives / not fixed (for the record)
- Persona-reported items confirmed as **upstream** (not tool defects): MaveDB read-timeouts / HTTP 504 on large score sets; Pharos 502; `api.fda.gov` no-route; IMGT GENElect scrape read-timeout; a multi-allelic dbSNP rsID that GenomeNexus resolves to one allele.
- Confirmed **working** (parameters honored, no silent-ignore): MassBank advanced-search filters (each narrows correctly), GDC `project_id`/`size`, GTEx tissue filters, InterPro/ELM/IUPred3/AlphaFill/IPD/MHCMotifAtlas, ClassyFire, SwissLipids, LipidMaps by-formula/id/xref, FooDB.
- Deferred (noted, not fixed here): `LipidMaps_search_by_name` sn-position chain notation returning empty (needs an `abbrev_chains` endpoint fallback); `MassBank_search_records_advanced` empty-result shape inconsistency (LOW); `ESM2` "keyless" description drift (upstream HF auth change); `InterPro` gene-symbol-vs-accession error message (LOW).

### DIFF
--- a/src/tooluniverse/data/massive_tools.json
+++ b/src/tooluniverse/data/massive_tools.json
@@ -94,13 +94,13 @@
     },
     {
         "name": "MassIVE_get_protein_identifications",
-        "description": "Identification-level access to MassIVE datasets via the ProXI standard API. MassIVE's other tools return only dataset metadata; this tool exposes the actual protein identifications. Three modes: (1) per-dataset protein summaries - pass 'accession' (e.g. 'PXD000561') to get identified proteins with countPSM/countPeptides/countPeptidoforms/countDatasets; (2) cross-dataset protein lookup - pass 'protein_accession' (e.g. 'A2M_MOUSE') to find how many MassIVE datasets identified a protein; (3) peptide-spectrum matches - set result_type='psms' with a dataset 'accession' to get PSMs (peptide sequence + charge). Returns a structured success/error envelope.",
+        "description": "Identification-level access to MassIVE datasets via the ProXI standard API. Keyed on 'protein_accession' (e.g. 'A2M_MOUSE'): with result_type='proteins' (default) it returns that protein's cross-dataset identification counts (countPSM/countPeptides/countPeptidoforms/countDatasets); with result_type='psms' it returns that protein's peptide-spectrum matches (peptide sequence + charge). Note: the MassIVE ProXI proteins/psms endpoints do not filter by dataset 'accession' (they return the same global summary regardless), so per-dataset identification listing is not supported here - use MassIVE_get_dataset for dataset-level metadata. Returns a structured success/error envelope.",
         "parameter": {
             "type": "object",
             "properties": {
                 "accession": {
                     "type": ["string", "null"],
-                    "description": "Dataset accession (ProteomeXchange PXD or MassIVE MSV, e.g. 'PXD000561'). Required for per-dataset protein summaries and for PSM lookups."
+                    "description": "Optional dataset accession (ProteomeXchange PXD or MassIVE MSV). NOTE: the MassIVE ProXI endpoints ignore this as a filter, so it does not restrict results to a dataset. Use 'protein_accession' to query; use MassIVE_get_dataset for dataset metadata."
                 },
                 "protein_accession": {
                     "type": ["string", "null"],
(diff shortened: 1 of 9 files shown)


gasvn:

PR #396, opened 2026-07-25
Fix: user-experience gaps in errors, defaults and silent wrong answers

gasvn:
## Summary

Round 3 of the role-play bug-hunt, scoped to **user-experience shortcomings** — the ways a real scientist using ToolUniverse via the CLI, MCP, or the Python SDK gets confused, misled, or silently handed a wrong answer.

Four researcher personas (clinical pharmacogenomics, structural biology, genetic epidemiology, clinical microbiology) worked real questions end-to-end across ~120 tools and reported **78 candidate issues**. Each was reproduced on the CLI *before* any code changed and re-verified after; **9 were confirmed and fixed**, and the rest were either lower-value or explicitly discarded as false positives (see below).

Rounds 1–2 concentrated on "valid input → empty/wrong data returned as success". This round found fresh instances of that class *and* three classes it had not covered: **a documented interface that does not work**, **errors that point nowhere**, and **first-run friction**.

---

### 1. ClinVar rejected standard coding HGVS (`clinvar_tool.py`)

`ClinVar_search_variants` stripped the `c.` prefix and sent the term unquoted:

```
$ tu run ClinVar_search_variants '{"gene":"DPYD","variant_name":"c.1905+1G>A"}'
  total_count: 0    query_translation: DPYD[gene] AND 1905+1G>A[Variant name]
```

A clinician reads `status: success, total_count: 0` as *"this variant is not in ClinVar"* — for DPYD `c.1905+1G>A` (`*2A`), a fluoropyrimidine variant with boxed-warning-level consequences.

The prefix-stripping was itself an earlier fix, verified against *protein* notation. Testing both notations against NCBI shows **no single spelling works for both** — coding notation carries `+`/`>` that Entrez only takes literally inside quotes:

| term | hits |
|---|---|
| `DPYD "c.1905+1G>A"[Variant name]` | **1** |
| `DPYD "1905+1G>A"[Variant name]` | 0 |
| `SERPINA1 "p.Glu342Lys"[Variant name]` | 0 |
| `SERPINA1 "Glu342Lys"[Variant name]` | **2** |

Both quoted spellings are now OR'd — an OR can only add matches the old query missed. Verified: DPYD `c.1905+1G>A` 0 → 1 (VCV000000432), CYP2C19 `c.681G>A` 0 → 36, SERPINA1 `p.Glu342Lys` still 2, HBB `p.Glu7Val` still 9, rsID routing unchanged.

### 2. UniProt's `total_results` reported the page size (`uniprot_tool.py`)

UniProt returns the match total in the `x-total-results` **header**; the code read a `resultsFound` key that the JSON body does not have, so the fallback always won and a field named `total_results` echoed `limit`:

```
limit=2  -> total_results=2      limit=5 -> total_results=5      (true total: 395446)
```

Now reads the header. Verified across all three affected tools: UniRef `beta-lactamase` 2 → 38559 (the tool's default is UniRef50 — confirmed that is the correct total for the query it issues), `blaKPC` in *K. pneumoniae* 5 → **62**, UniParc 3 → **838346**.

### 3. PDB taxonomy answered with a kingdom (`rcsb_pdb_tools.json`)

`get_taxonomy_by_pdb_id` promises "the scientific name … of the organism(s)", but the GraphQL selection asked only for the source organism's **parent** taxon plus the expression host's lineage. For 2OV5 (KPC-2 carbapenemase) it returned `"Bacteria"` while prominently showing *Escherichia coli* — the expression host. A microbiologist concludes KPC-2 is an *E. coli* protein.

Now requests `ncbi_scientific_name` (and the taxid) for the source organism, keeping the host clearly separate. Verified: 2OV5 → *Klebsiella pneumoniae*, 1CRN (its own documented example) → *Crambe hispanica* subsp. *abyssinica*, 6NVB/4HHB → *Homo sapiens*.

---

### 4. `tu run <tool> limit=10` failed on most tools (`execute_function.py`)

The `key=value` shorthand is advertised in `tu --help` and `tu run --help`. It failed on most tools:

```
$ tu run CPIC_get_alleles genesymbol=CYP2C19 limit=10
Error: Parameter validation failed for 'limit': '10' is not of type 'integer', 'null'.
```

JSON Schema permits `"type"` to be a **list**, and `["integer","null"]` is this repo's own idiom for an optional parameter — but every comparison in `_coerce_value_to_type` was against a bare string, so a union type matched nothing and the value stayed a `str`. The JSON form worked, making the failure look tool-specific rather than systemic.

Union types are now coerced by trying each non-null member (strings are left alone when the union accepts them, so `Feature-27A-02` does not regress). This also fixes LLM/MCP callers that pass numbers as strings. All three commands the persona reported as broken now succeed.

---

### 5. Misspelled parameters were not named (`base_tool.py`)

The "did you mean?" hint only matched **case variants**, so the most common real mistakes fell through:

```
$ tu run UniProt_get_entry_by_accession '{"acession":"P12345"}'
Error: ... 'accession' is a required property
```

The user *did* pass an accession, so the error reads as nonsense. Matching now widens in three stages — case-insensitive, separator-insensitive (`geneName` → `gene_name`), then a conservative fuzzy match — and when nothing plausibly matches, it names the unrecognized keys:

```
... (you passed 'acession' — did you mean 'accession'?)
... (unrecognized parameter(s): 'totally_unrelated')
```

Only keys that are *not* valid schema properties are considered, so a legitimately-supplied sibling parameter is never mislabelled.

### 6. `tu run <typo>` gave no suggestion (`cli.py`)

`tu info` already answered a typo with `Did you mean: ...`; `tu run` — the command people actually type — did not, even though `_render_run_error` already knew how to display suggestions. Nothing on the run path populated them. Same `difflib` cutoff as `cmd_info`; skips key-gated tools, whose name is already correct.

### 7. A typo'd tool name told SDK callers to check the network (`execute_function.py`)

`ToolUnavailableError` covers both "service is down" and "no such tool", and its default `next_steps` are network-flavoured. Two of the three not-found sites passed none, so a **misspelled name** advised *"Check network connection"* / *"Verify service status"*. The CLI renderer masked this; direct SDK, MCP and `--json` consumers saw it raw. Both branches now match the sibling path already used by `run_one_function`.

### 8. A key-gated tool was reported as a spelling problem (`execute_function.py`)

`tu info <gated tool>` printed two contradictory lines:

```
⚠️  Tool name DisGeNET_get_vda not found in the loaded tools.
Error: Tool 'DisGeNET_get_vda' requires API key(s) not set: DISGENET_API_KEY. ...
```

The tool exists and the name is right. The warning now says so; genuinely-absent tools keep the original wording.

---

### 9. `import tooluniverse.tools` emitted a SyntaxWarning (`generate_tools.py`)

The wrapper generator escaped backslashes in a parameter description **before** truncating it to 80 characters. A cut landing inside an escaped `\\` pair left a lone trailing backslash, so the docstring held an invalid escape:

```
$ python -c "import tooluniverse.tools"
.../tools/MEME_fimo_scan.py:30: SyntaxWarning: invalid escape sequence '\.'
```

A FASTA `\n` example in a long description is enough to trigger it. Truncation now precedes escaping — the order the module-level description path already used. One shipped wrapper was affected and is corrected to match; a regression test compiles **all 2738** wrappers and asserts none warn.

---

## Verification

- Every fix reproduced on the CLI against the live tool/API before the change, and re-checked after.
- **40 new tests** in `tests/unit/test_ux_error_hints_r3.py`, including negative cases: a valid sibling parameter is not flagged as a typo, a nonsense tool name gets no spurious suggestion, a gated tool keeps its key guidance, `Feature-27A-02` string-preservation still holds, and valid arguments still pass.
- Two existing ClinVar tests pinned the previous *unquoted, prefix-stripped* term. Their intent (the stripped spelling stays reachable; rsIDs are not field-tagged) is preserved and re-asserted against the corrected, quoted form.
- Full `tests/unit` suite green. The pre-existing `TestRound25Fixes::test_validate_case_variant_shows_did_you_mean` still passes — the original case-variant contract is intact.
- `import tooluniverse.tools` works; all 2738 generated wrappers compile warning-free.

Two failures in `tests/tools/test_tu_cli.py` (`test_render_run_non_json_string_detail_shown_raw`, `test_find_tool_name_with_underscores_works`) were checked against a clean `main` and **fail there identically** — pre-existing and unrelated. That file is excluded from the default suite by `pytest.ini`.

---

## Discarded during verification

Roughly half of the raw persona reports do not survive checking. These were reproduced and then **rejected** rather than "fixed":

- **ChEMBL `standard_value__lte` / `__gte`** — a static scan flagged them as never read in Python. Live comparison shows the filter applied correctly (unfiltered IC50s up to 20000; with `__lte: 10` every value ≤ 10). Config-driven REST pass-through, working as intended.
- **FAERS `patientsex` / `patientagegroup`** — schema enums are `"Male"`/`"Female"` while FAERS codes them numerically, which looks like a silently-ignored filter. Live counts differ correctly by sex (122380 vs 137733), so the mapping works.
- **50 API-key tools whose description omits the key requirement** — the runtime already handles this well across `run`, `info` and `grep`, each naming the exact missing variable. Not a user-visible gap.
- **`tu test` return_schema envelope mismatch** — 1703 tools carry envelope-shaped `return_schema` while `tu test` validates `result["data"]`. Spot-checks did not reproduce a mismatch (an envelope-shaped tool passed; the observed failures were upstream HTTP 429s from my own test traffic). Not treated as a defect on this evidence.
- **A broad "declared parameter never read in code" scan** — 371 hits, dominated by config-driven REST parameter mapping. Sampling confirmed false positives; not pursued.

## Not fixed, worth a follow-up

Several confirmed findings are real but larger than this PR, and are recorded here rather than half-fixed: undocumented result caps with no pagination escape (`PDBeSIFTS_get_best_structures` 50 of 979, `RCSBAdvSearch_search_structures` 50 of 632, `CPIC_get_alleles` 50 of 208); `gwas_get_variants_for_trait` returning an unsorted page while its sibling sorts by p-value; and wide envelope/parameter-name drift across sibling tools (nine distinct top-level shapes in one 33-tool session). These need a convention decision rather than a point fix.

## Notes

No version bump, no release marker, no tag.


gasvn:

PR #397, opened 2026-07-25
Fix: silently-dropped filters, unreachable result sets, and misreported values (bug-hunt round 4)

gasvn:
Round 4 of the role-play bug hunt. Four researcher personas worked real end-to-end
questions in domains earlier rounds had not touched — neuroscience, immunology,
medicinal chemistry / cheminformatics, and microbiome / systems biology — plus
follow-up on four result-cap problems CLI-flagged during round 3.

**29 candidates reported → 19 CLI-verified → 17 fixed → 3 discarded as false positives.**
Every fix was reproduced against the live upstream API before the change and
re-verified after it. Nothing here was taken on a persona's word.

The theme that emerged is narrower than "bugs": **a tool accepts an input or reports
a number, and the answer it gives back is confidently wrong rather than obviously
broken.** A silently-dropped filter, a cap with no escape, a count that means
something other than what it is named — each one returns `status: success` and looks
like a real answer.

---

### Filters that were accepted and then ignored

| Tool | Asked for | Actually returned |
|---|---|---|
| `MGnify_search_genomes` | `taxonomy="Bacteroides"` | all **56,782** genomes (correct: 136) |
| `HPA_get_rna_expression_by_source` | GFAP in thalamus | `"N/A"` for **every gene, every brain region** |
| `NeuroMorpho_search_neurons` | `brain_region="entorhinal cortex"` | 872 **zebra-finch song-nucleus** neurons (correct: 3,431 entorhinal) |
| `GtoPdb_search_targets` | `gene_symbol="CHRNA7"` | the unfiltered first 50 targets, alphabetically |

`MGnify_search_genomes` sent `?lineage=` where the endpoint filters on
`?taxon_lineage=`; MGnify drops unknown query params rather than erroring, so
`taxonomy="Bacteroides"`, `taxonomy="COMPLETELY_BOGUS_XYZ"` and no taxonomy at all
all returned the identical full catalogue. Its `genome_type` filter has no working
upstream equivalent at all (`?genome_type=`, `?type=`, `?genome-type=` each return
all 56,782), so it now **fails closed** instead of silently returning everything.
The description also advertised NCBI phylum names against a GTDB catalogue —
`Bacteroidetes` returns 0 where `Bacteroidota` returns 6,895.

`HPA_get_rna_expression_by_source` queried column `rnabrm`, which is not a valid HPA
column — HPA silently drops unknown columns. The whole brain mode was dead, and the
`tissue` mode used an aggregate column populated only for tissue-enriched genes, so
even the tool's own documented example (TP53 / liver) came back empty. Now:
GFAP/thalamus 14782.4 nTPM, GFAP/cerebellum 9332.2, APP/cerebral cortex 669.1,
TP53/liver 15.6.

`GtoPdb_search_targets` had a `gene_symbol`→`targetId` resolution step meant for
`/targets/{id}/interactions` firing on the plain `/targets` endpoint, where
`targetId` is just an ignored query param — and `gene_symbol` was popped on the way.
CHRNA7, GRIN2B and HTR6 now each resolve to their one correct target.

### Result caps with no way past them

| Tool | Returned | Actually available | Escape before |
|---|---|---|---|
| `RCSBAdvSearch_search_structures` | 50 | 81,865 | none — `start` hard-coded to 0 |
| `PDBeSIFTS_get_best_structures` | 50 | 676 (TP53) | none — no limit/offset param |
| `ExpressionAtlas_search_experiments` | 50 | 241 | none |
| `CPIC_get_alleles` | 50 | **208** CYP2D6 alleles | `limit` worked but total was hidden |
| `intact_get_interactions` | 25 | 87 | `size` worked but was **undeclared** |

`RCSBAdvSearch_search_structures` also silently reduced `rows=200` to 50 with nothing
in the response saying so. It now reports the clamp and the `start` value for the
next page. `PDBeSIFTS_get_pdb_to_uniprot` additionally cut chain mappings at 20 with
*no count at all* — `total_proteins` counts proteins, not mappings — so a
multi-chain entry looked complete at 20.

`CPIC_get_alleles` reported only `count`, the size of the returned page, so CYP2D6
looked like it had 50 star alleles when CPIC curates 208. A pharmacogenomics user
calling star alleles silently lost 158 of them. It now asks PostgREST for an exact
count and accepts `offset`.

`intact_get_interactions` honoured a `size` parameter it never declared — an MCP
client seeing only the schema had no way to reach 62 of PD-1's 87 interactors.

### Numbers and labels that misreported the data

**`RDKit_matched_molecular_pair` reported the monoisotopic mass as "MW".** Imatinib
came back as 493.26 where PubChem, ChEMBL, SwissADME and `SMILES_verify` all say
493.6, and the error grows with halogens (bromoacetanilide 212.98 vs the true
214.06). MW is the denominator for ligand efficiency, the Ro5 ≤500 cutoff and
mg→mmol dosing. Now average MW, with `exact_mass` reported separately.

**`RDKit_pharmacophore_features` counted every amide nitrogen as a basic centre.**
Its SMARTS excluded only hydroxyl/oxide neighbours and explicitly matched all neutral
aromatic N, giving acetamide, acetanilide and pyridine one basic centre each and
imatinib seven — its entire nitrogen count — where RDKit's own `BaseFeatures.fdef`
gives 0, 0, 0 and 2. Amides are near-universal in drug-like molecules, so the false
positive hit almost every query. The replacement pattern agrees with RDKit's
reference on all 13 molecules tested, including morphine, propranolol and
sulfanilamide. Separately, `include_features=["HBD"]` reported imatinib as having
zero H-bond acceptors and zero aromatic atoms; filtered-out features now report
`null` with a note.

**`HLALigandAtlas_get_benign_peptides` had its binder flags documented backwards.**
The docs said the flags were "strong `n/` or weak `w/`"; the data carries three
prefixes. Matching peptides against canonical anchor motifs for three alleles with
distinct motifs settles it without needing a predictor:

| flag | B\*07:02 (P2=P, C-term=L/F) | B\*44:03 (P2=E, C-term=Y/F) | A\*03:01 (C-term=K/R) |
|---|---|---|---|
| `s/` | **71%** | **69%** | **73%** |
| `w/` | 0% | 7% | 13% |
| `n/` | **0%** | **0%** | **0%** |

So `s/` is the strong binder and `n/` is a NON-binder. This tool exists to exclude
self-peptides from neoantigen candidates, so reading `n/` as "strongly presented"
inverts the safety call — you drop safe candidates and keep unsafe ones. The `allele`
filter is also a plain substring match, so `A*02:01` matches `n/A*02:01` too; both
facts are now documented.

**`DGIdb` reported `metadata.total` as the gene count.** Querying PDCD1 returned
`total: 1` beside 68 drug-gene interactions; three genes reported 3 against 140. Next
to a 68-row payload, `total: 1` reads as a truncation warning.

**`PubChem_get_CID_by_SMILES` passed through the CID-0 not-found sentinel as a
success.** A structure PubChem does not hold returned `{"CID": [0]}` at HTTP 200;
callers chained it into `PubChem_get_compound_properties_by_CID`, which died two
steps later with "Invalid ID, must be positive integer". Novel and proprietary
analogues are exactly the inputs that hit this. The name-based sibling already
reports a clean "No CID found"; now this one matches.

### Ordering and pagination

**`gwas_get_variants_for_trait` returned an unsorted page.** It and
`gwas_get_associations_for_trait` query the same `/v2/associations` endpoint, but only
the sibling sent `sort=p_value&direction=asc`, and the GWAS Catalog returns
associations unsorted. For asthma (3,219 associations) this tool returned p-values
2e-06 … 2e-24 while the sibling returned 7e-288, 8e-223, 2e-156 — so the first page
held **none** of the genome-wide-significant loci a user asking "which variants are
associated with this trait" is looking for.

**Seven IEDB tools declared an `offset` that always failed**, because PostgREST
rejects `offset` without `order`. Only `iedb_search_epitopes` carried a default
order; pagination was 100% broken on the rest.

---

## Fixes shipped (17)

1. `gwas_get_variants_for_trait` — default to p-value order, matching its sibling; expose `sort`/`direction`
2. `RCSBAdvSearch_search_structures` / `_search_by_motif` — add `start`/`offset`; report the `rows` clamp and next page
3. `PDBeSIFTS_get_best_structures` / `_get_all_structures` — add `limit` (max 500) and `offset`
4. `PDBeSIFTS_get_pdb_to_uniprot` — report `total_chain_mappings` and a truncation flag
5. `CPIC_get_alleles` — exact `total_count` via PostgREST `Content-Range`, plus `offset`
6. `MGnify_search_genomes` — `taxon_lineage` filter; `genome_type` fails closed; GTDB naming documented
7. `HPA_get_rna_expression_by_source` — query HPA's real per-source columns
8. `NeuroMorpho_search_neurons` — quote multi-word values in `q` and `fq`
9. `GtoPdb_search_targets` — route `gene_symbol` to `?geneSymbol=`; declare it; stop `name` claiming gene-symbol support
10. `RDKit_matched_molecular_pair` — `MW` is average mass; `exact_mass` added
11. `RDKit_pharmacophore_features` — PosIonizable SMARTS matched to RDKit's reference definitions
12. `RDKit_pharmacophore_features` — filtered-out features report `null`, not `0`
13. `PubChem_get_CID_by_SMILES` — CID-0 sentinel becomes "No CID found"
14. DGIdb interaction tools — count named for what it holds; `interactions_total` added
15. Seven IEDB tools — default `order` so the declared `offset` works
16. `intact_get_interactions` / `_get_interactor` — declare `size`; drop the unreachable `xml` enum option
17. `ExpressionAtlas_get_baseline` / `_search_differential` / `_search_experiments` — add `limit`/`offset`

Defaults are unchanged throughout: existing callers get the same records, plus the
new count and paging fields.

## Tests

76 new regression tests across four files. All touched tests pass and
`import tooluniverse.tools` works.

- `tests/unit/test_result_cap_pagination.py`
- `tests/unit/test_gwas_variants_for_trait_sort.py`
- `tests/unit/test_silently_dropped_filters.py`
- `tests/unit/test_wrong_values_and_counts.py`

The PosIonizable test asserts agreement with RDKit's own `BaseFeatures.fdef` rather
than hard-coding this implementation's answers, so it stays honest if the SMARTS is
revised again.

## Discarded as false positives (3)

- **`ClinGen_search_gene_validity` 100-row cap** — the tool requires `gene`, and
  per-gene result sets are tiny (BRCA1 returns 2 of 2). The cap never bites.
- **`GtoPdb` "500-element truncation"** — both sites are `response.text[:500]` error
  message truncation, not result caps.
- **A repo-wide scan for hard-coded slices** flagged 95 modules; spot-checking showed
  most are string truncation (`str(e)[:200]`) or already report a true total. Not
  treated as a bug class.

---

### 1. Pagination vocabulary drift

**926 of 2,570 tools expose a paging parameter, across 39 distinct vocabularies.**

```
 334  limit                 61  size                 19  start
 164  limit+skip            56  max_results          18  page+page_size
  76  limit+offset          44  page_size            16  limit+page
  22  page+size             22  page                 15  pageSize
                                                      7  rows
```

The four tools in this PR alone used `limit`/`offset`, `rows`/`start`, `size`/`page`
and `size`. Picking one vocabulary is a project-wide convention call, not something
to normalise unilaterally — the same reasoning that kept round 3 from touching the
response-envelope drift (~9 distinct top-level shapes), which is still open and still
worth a decision.

### 2. What a "count" field means

Two independent bugs this round were the same ambiguity: `count` and `total` are used
for both *records returned* and *records available*, and the surrounding fields
(`hitCount`, `totalElements`, `total_count`, `total_results`, `experiment_count`)
don't disambiguate either. A convention — say, `returned` always means this page and
`total_count` always means the full result set — would make this class of bug
mechanically detectable instead of one-by-one.

### 3. Verified but not fixed

Real, reproduced, left alone to keep this PR reviewable. Each is a genuine
user-facing problem:

- `OpenTargets_get_target_expression_by_ensemblID` — reports `count: 1474`, returns 25
  rows across 3 tissues, no paging parameter. For APP the answer is "pancreas and
  pituitary" with zero brain rows. Its sibling in the same family exposes `size`/`index`.
- `AllenBrain_get_structure_expression_values` — 50 of 2,407 regions in arbitrary
  order; upstream supports `order=expression_energy$desc`.
- `IPD_search_hla_alleles` — `total: 512`, only 100 reachable; the upstream `next`
  cursor is discarded.
- `MGnify_search_studies` / `MGnify_list_analyses` — no `page` parameter and the
  upstream `count`/`pages` are dropped (590 gut studies, 100 visible). Sibling MGnify
  tools do expose `page`.
- `NCBIDatasets_list_genomes_by_taxon` — honest `total_available: 2077`, clamps to
  100, discards the `next_page_token` the API hands it.
- `NCBIDatasets_suggest_taxonomy` — `total_results: 20` is the upstream cap, not a total.
- `STRING_get_enrichment` — 230 terms in category-blocked order, not p-value order;
  the most significant term sits at index 69. Sibling enrichment tools rank properly.
- `OrthoDB_search_groups` — `data.orthodb.org/v12/search` returns a Postgres error at
  HTTP 404 while v11 search and every other v12 endpoint work. It is the only way to
  obtain a `group_id`, so the whole OrthoDB family is unreachable.
- `IEDB_predict_mhci_binding` — injects `score: 0.0` for every peptide on the three
  methods where upstream returns no score column; `ic50` is a string with no unit.
- `GtoPdb_get_interactions` — its own truncation message names a `limit` parameter
  that works but is undeclared, and passing it drops the pagination metadata.

### DIFF
--- a/src/tooluniverse/cpic_search_pairs_tool.py
+++ b/src/tooluniverse/cpic_search_pairs_tool.py
@@ -358,3 +358,87 @@ class CPICSearchPairsTool(BaseRESTTool):
 
     def _build_url(self, args: Dict[str, Any]) -> str:
         return super()._build_url(self._resolve_aliases(args))
+
+
+@register_tool("CPICGetAllelesTool")
+class CPICGetAllelesTool(BaseTool):
+    """List CPIC star alleles for a gene, with a true total and paging.
+
+    Fix-R4B-4: this tool ran as a plain BaseRESTTool over a URL template
+    carrying `limit={limit}` with a default of 50, and the only count in the
+    response was `count`, which reported the size of the returned page. So
+    `CPIC_get_alleles {"genesymbol": "CYP2D6"}` came back with 50 alleles and
+    `count: 50` -- indistinguishable from CYP2D6 genuinely having 50 alleles,
+    when CPIC actually curates 208 (confirmed live via the PostgREST
+    `Content-Range` header, `0-207/208`). A pharmacogenomics user calling
+    star alleles from that page silently loses 158 of them, and there was no
+    `offset` to walk the rest. Ask PostgREST for the exact count, report it,
+    and page explicitly.
+    """
+
+    _DEFAULT_LIMIT = 50
+    _MAX_LIMIT = 1000
+
+    def run(self, arguments: Dict[str, Any]) -> Dict[str, Any]:
+        """Fetch one page of alleles plus the gene's true allele count."""
+        gene = arguments.get("genesymbol")
+        if not gene:
+            return {"status": "error", "error": "genesymbol is required (e.g. CYP2D6)"}
+        gene = str(gene).upper()
+
+        raw_limit = arguments.get("limit")
+        limit = self._DEFAULT_LIMIT if raw_limit in (None, "") else int(raw_limit)
+        limit = max(1, min(limit, self._MAX_LIMIT))
+        offset = max(0, int(arguments.get("offset") or 0))
+
+        params = {
+            "genesymbol": f"eq.{gene}",
+            "select": "name,functionalstatus,activityvalue,clinicalfunctionalstatus",
+            "limit": limit,
+            "offset": offset,
+        }
+        try:
+            # Prefer: count=exact makes PostgREST report the unpaged total in
+            # the Content-Range response header ("0-49/208").
+            response = requests.get(
+                f"{_CPIC_API}/allele",
+                params=params,
+                headers={"Prefer": "count=exact"},
+                timeout=30,
+            )
+            response.raise_for_status()
+            data = response.json()
+        except requests.exceptions.RequestException as e:
+            return {"status": "error", "error": f"CPIC API error: {e}"}
+
+        total = self._total_from_content_range(response.headers.get("Content-Range"))
+        if total is None:
+            total = offset + len(data)
+
+        result: Dict[str, Any] = {
+            "status": "success",
+            "data": data,
+            "url": response.url,
+            "count": len(data),
+            "total_count": total,
+            "offset": offset,
+        }
+        if data and offset + len(data) < total:
+            result["note"] = (
+                f"Showing alleles {offset + 1}-{offset + len(data)} of {total} curated "
+                f"by CPIC for {gene}. Re-run with offset={offset + len(data)} for the "
+                f"next page, or raise 'limit' (max {self._MAX_LIMIT})."
+            )
+        elif not data and total:
+            result["note"] = (
+                f"No alleles at offset={offset}; {gene} has {total} alleles in CPIC."
+            )
+        return result
+
+    @staticmethod
+    def _total_from_content_range(header: Optional[str]) -> Optional[int]:
+        """Parse the unpaged total out of a PostgREST Content-Range header."""
+        if not header or "/" not in header:
+            return None
+        total = header.rsplit("/", 1)[1].strip()
+        return int(total) if total.isdigit() else None
--- a/src/tooluniverse/data/cpic_tools.json
+++ b/src/tooluniverse/data/cpic_tools.json
@@ -963,7 +963,7 @@
   },
   {
     "name": "CPIC_get_alleles",
-    "description": "Get pharmacogenomics alleles and their functional status for a gene from CPIC. Returns allele names, activity values, and functional status. Note: use 'clinicalfunctionalstatus' field for the active clinical classification (e.g., 'No function', 'Normal function'); the 'functionalstatus' field may be null for some genes. Critical for translating genotype to phenotype (e.g., CYP2D6 *4 = no function, CYP2D6 *1 = normal function).",
+    "description": "Get pharmacogenomics alleles and their functional status for a gene from CPIC. Returns allele names, activity values, and functional status, plus 'total_count' -- the number of alleles CPIC curates for the gene, which is often far larger than one page (CYP2D6 has 208). Returns 50 per call by default; use 'limit' (max 1000) and 'offset' to retrieve the rest. Note: use 'clinicalfunctionalstatus' field for the active clinical classification (e.g., 'No function', 'Normal function'); the 'functionalstatus' field may be null for some genes. Critical for translating genotype to phenotype (e.g., CYP2D6 *4 = no function, CYP2D6 *1 = normal function).",
     "parameter": {
       "type": "object",
       "properties": {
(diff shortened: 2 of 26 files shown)


gasvn:

PR #398, opened 2026-07-26
Fix nine tools that returned confidently wrong answers instead of failing

gasvn:
Round 5 of the role-play bug hunt. Four researcher personas (cancer epigenetics, clinical pharmacology/pharmacovigilance, pathway systems biology, computational toxicology/structural biology) worked in tool families untouched by earlier rounds. They reported 40 candidates; each was re-verified against the live API before any code was touched, running the tool **with and without** the input in question plus a **bogus-value control**. 9 were fixed here, and the discards are listed at the bottom.

Every fix targets the same failure mode: **the tool accepts an input or reports a number and answers confidently wrong, rather than erroring**. That is the dangerous case — a scientist has no signal that anything went wrong.

No overlap with the open PRs #395, #396 or #397 (file lists checked before editing).

### 1. GTEx resolved every gene ID against GENCODE v26 regardless of dataset
`gtex_v2_tool.py`, `gtex_tool.py`, `data/gtex_v2_tools.json`

Each GTEx dataset is annotated against a different GENCODE release. TP53 is `ENSG00000141510.16` in gtex_v8 (v26) but `.18` in gtex_v10 (v39). The resolver hardcoded `gencodeVersion=v26`, so selecting `dataset_id="gtex_v10"` — a documented enum value — sent v26 IDs to a v39 dataset. The API answered HTTP 200 with an empty list, surfacing as `status: "success"`, `num_results: 0`: *"this gene has no expression in V10"*. The resolver also stripped and re-resolved any supplied version suffix, so a caller passing the correct v39 ID had it downgraded back to v26 — there was no workaround from outside.

| query | before | after |
|---|---|---|
| TP53 median expression, `gtex_v10` | 0 | **54 tissues** |
| ERAP2/Liver eQTLs, `gtex_v10` | 0 | **250** |
| anything on `gtex_v8` | unchanged | unchanged |

The per-dataset GENCODE map is taken from the API's own `/metadata/dataset`. This also means the code comments and tool descriptions asserting *"v10 returns empty for most endpoints"* were documenting this bug — verified false for expression, eQTL, sample and top-expressed-gene endpoints, and corrected. `gtex_v8` remains the default.

### 2. ENCODE searches answered a zero-result query with another species' experiments
`epigenomics_tool.py`

ENCODE returns **HTTP 404 carrying a valid body** (`{"total": 0, "@graph": [], "notification": "No results found"}`) when a facet combination matches nothing. `_encode_search` treated that as a transport error, and callers reacted in one of two wrong ways:

- three searches retried with the organism filter **deleted**. "H3K4me3 in human midbrain" — which ENCODE does not have — returned four *Mus musculus* embryo experiments, with `metadata.organism` still reading `"Homo sapiens"`.
- five others reported a legitimately empty query as `"API HTTP error: 404"`, a fake outage.

The bogus-value control was arithmetic: `Homo sapiens` 431 + `Mus musculus` 97 = **528**, exactly what a nonsense organism returned. After the fix a nonsense organism returns 0, real organisms are unchanged, and `sibling encode_tool.py::_http_get` already handled the 404 this way — the two modules now agree.

### 3. PubChem toxicology dropped the carcinogenicity of every carcinogen tested
`pubchem_tox_tool.py`

PubChem packs every statement of one notification group into a single `Information` entry holding N `StringWithMarkup` elements. Only element 0 was read. This was not a random sample: GHS codes are emitted in ascending numeric order and physical hazards (H2xx) sort ahead of health hazards (H3xx), so the health hazard was discarded whenever a physical one existed.

Benzene (IARC Group 1) returned **H225 + H401** — "highly flammable", "toxic to aquatic life" — while PubChem lists 16 codes including **H350 "May cause cancer"**, H340 and H372. Verified identically for vinyl chloride, formaldehyde and arsenic trioxide; all four now report H350. The same truncation had `PubChemTox_get_carcinogen_classification` citing the withdrawn 1987 IARC monograph while dropping the current 2012 Volume 100F.

This helper backs all eight `PubChemTox_*` tools, so their outputs are now longer.

### 4. EBI Proteins reported other proteins' interactions as the query protein's
`ebi_proteins_interactions_tool.py`

`/proteins/interaction/{acc}` returns the queried protein **plus ~100 neighbour proteins**, each carrying its own full interaction list. All were pooled.

| protein | reported | true | wrong rows on default page |
|---|---|---|---|
| TP53 (P04637) | 3258 | **185** | **44 of 50** |
| Insulin (P01308) | 735 | **9** | — |

TP53's "partners" included AXIN1–GSK3B, BRCA2–RAD51 and PIK3CA–PIK3R1 pairs. Insulin's included huntingtin and ataxin-1 — partners of a chaperone hub in the same payload. Entries are now scoped by their own `accession`; the same scoping is applied to `_get_interaction_details`, whose metadata also no longer assumes the query protein is `data[0]`. The file already contained a comment diagnosing this exact pooling problem for diseases/locations; the reasoning is now applied to interactions too.

### 5. ClinicalTrials.gov silently disabled its phase filter when a status filter was set
`ctg_tool.py`

`search_clinical_trials` documents *"Limited to trials beyond phase 1"* and implements it with a hidden `filter.advanced` default. That default was skipped whenever `filter.overallStatus` was present — guarded by a comment describing an `AREA[HasResults]true` clause that an earlier fix had **already deleted**. The workaround outlived its reason.

The signature was logically impossible: adding the *narrowing* `overall_status=["COMPLETED"]` to a metformin/lactic-acidosis search **raised** `total_count` from 2 to 3 and returned a disjoint set. A RECRUITING pembrolizumab search returned **745** studies instead of **557** — 188 wrongly included, with pure `PHASE1` and `EARLY_PHASE1` trials on page 1.

### 6. GO_get_genes_for_term returned N empty objects, and its taxon filter did nothing
`gene_ontology_tool.py`, `data/gene_ontology_tools.json`

Two defects compounded. The extract path `associations[*].subject` keyed on a field the Biolink API no longer returns — it now answers with flat annotation records (`bioentity_label`, `taxon`, `evidence_type`) and **no `subject` key** — so `assoc.get("subject", {})` produced `{}` for every row. `rows=10` returned ten empty dicts as a success. The declared `return_schema` was `array of objects with no properties`, so nothing could catch it.

Second, `taxon` was a **required** parameter the endpoint silently ignored: `NCBITaxon:10090` (mouse) returned human and *C. elegans* rows, and a nonsense taxon returned the same 50 records.

Now queries the GO Solr index, which honours the filter (human 1257, mouse 2034, bogus **0**) and returns real genes. Annotation rows are collapsed to distinct genes with their evidence codes, `taxon` becomes a genuine optional filter, and the `return_schema` gains a `oneOf` envelope describing the actual fields.

### 7. WikiPathways returned 0 genes for its own documented example
`wikipathways_ext_tool.py`, `data/wikipathways_ext_tools.json`

`code` defaulted to `"H"` (HGNC), which filters on the *database that annotated each gene product*. WikiPathways rarely records HGNC there — WP254's 87 gene products come from Entrez Gene (84) and Ensembl (3), **none** from HGNC. So the default answered "0 genes", with `status: success`, for the pathway the tool's own description says returns 88.

Two smaller defects rode along: the `"S"` code matched the substring `"Uniprot"` against a store that writes `"UniProtKB"` under a case-sensitive SPARQL `CONTAINS`, so it was permanently dead (WP3594 has 206 UniProt-sourced gene products, returned 0); and counts deduplicated on `rdfs:label`, but one gene product carries several aliases, inflating WP254 from 87 to 134.

`code` is now an optional, case-insensitive filter; WP254 returns 87 gene products by default and `code="S"` on WP3594 returns 206. A bogus code still returns 0 (fails closed).

**Note on labels:** I deliberately did *not* collapse each gene product to one "primary" symbol. WikiPathways attaches unrelated symbols to a node — the AKT1 node (`ncbigene/207`) carries `AKT, AKT1, Akt1, PIK3CA, PKB, PKB/Akt` — so any heuristic would fabricate a gene. Instead `gene_count` counts gene products, `label_count` counts labels, `genes` keeps every label, and a new `gene_products` field gives node-level identifiers with their exact labels.

### 8. Ensembl LD counted its partners after truncating them
`ensembl_ld_tool.py`, `data/ensembl_ld_tools.json`

`ld_count` was computed on the post-truncation list. For rs1042779 in CEU at `r2_threshold=0.05` the tool reported **200 partners** when Ensembl returned **735**, and the returned set stopped at r2 **0.79** despite the requested 0.05 — wrong on both the count and the range, while the description promised *"all variants in LD"* and cited 708 partners. There was no parameter to reach past the cap.

Now reports `total_ld_count` (pre-truncation) alongside `ld_count`, sets `truncated`, and accepts `limit`. This follows the precedent already set by `ensembl_overlap_tool.py` in this repo, which correctly reports `total_features` and `returned`.

### 9. NHANES answered about a different survey cycle than requested
`nhanes_tool.py`, `data/nhanes_tools.json`

An unrecognised `year` fell through to `cycles[:2]`, so asking for **2021-2023** (a real, published cycle) returned rows labelled **2017-2018** and **2015-2016** under `status: success`, with nothing signalling the substitution. The hardcoded list was also stale — it began at 2007-2008 and omitted six earlier published cycles plus the current one — and every `download_url` was assembled from a template that does not exist: `.../Nchs/Nhanes/2017-2018/laboratory_2017-2018.aspx` redirects to CDC's `ErrorPage.aspx`.

Now rejects an unknown cycle (naming the valid ones), lists all published cycles including 2021-2023, notes that no standalone 2019-2020 cycle exists (COVID-19 cut that field period short; `DEMO_K.XPT` is a live 404), and emits the CDC datapage URL form, which returns 200. `return_schema` gains the required `oneOf` envelope.

## Testing

- **76 new regression tests** across 9 files, all offline (HTTP layer mocked), each documenting the live evidence in its module docstring.
- Full unit suite: **3337 tests, 0 failures** (only pre-existing environmental skips — no scanpy, lazy-load cache).
- `import tooluniverse.tools` works.
- Every fix was CLI-verified against the live API before and after the change.

## For the maintainer — convention decisions I did not make unilaterally

1. **GTEx default release.** `gtex_v8` is kept as the default because it is the most widely cited and changing it would move every existing answer. But `gtex_v10` is the current release (943 eQTL donors vs 838) and is now fully queryable. Worth deciding whether the default should move.
2. **`PubChemTox_*` output shape.** Fix 3 makes these tools return substantially more entries. Any downstream consumer assuming one string per Information entry will see more rows, and the `*_count` fields grow correspondingly.
3. **Count-that-is-really-a-page-size is systemic.** Beyond the two fixed here, the pathway persona independently found the same shape in `ReactomeContent_search` (reports 10, true 584), `ReactomeInteractors_get_protein_interactors` (20 vs 249), `WikiPathways_find_pathways_by_gene` (100 vs 172) and four KEGG search tools (25 vs 739). In every case the true total is available upstream for free, and `PathwayCommons_search` already does it right (`total_hits` + `page`). This deserves one convention applied across the repo rather than nine separate patches, so I have left them for a dedicated pass.
4. **`WikiPathways_get_pathway_genes` output.** Adding `label_count` and `gene_products` changes the response shape. If the project would rather keep the response flat, the alternative is to return only `gene_products` and drop `genes` — but that breaks callers using it as a gene set.

## Verified and discarded (not fixed)

Roughly half of the raw persona reports did not survive verification, plus several real findings deliberately deferred:

- **`ensembl_overlap_tool.py` truncation** — flagged by a static scan; it already reports `total_features` pre-truncation and `returned`. Correct as written.
- **GTEx `sex` / `age_bracket` filters** — suspected dropped; both genuinely filter (Liver+male returns only male, bogus sex is rejected with a 422 enum error).
- **`gtex_tool.py` hardcoded `datasetId`** — the v1 expression tool declares no `dataset_id` parameter, so hardcoding v8 matches its own resolver. Not a bug; the undocumented back-door in the sibling eQTL tool was made dataset-aware for consistency only.
- **NHANES `_search_datasets` / `NHANES_download_and_parse`** — both build working URLs, and the download path already handles CDC's HTML-error-page-with-200. Only `_get_dataset_info` was defective.
- **BLAST** — persona confirmed it polls to completion, `expect`/`hitlist_size`/`database` all reach the API, and "no hits" is distinguishable from a running job. The dead `max_wait_time`/`timeout` attributes are real but cosmetic; deferred.
- **AlphaFold confidence semantics** — checked specifically for inverted pLDDT/PAE. Pure pass-through, nothing relabelled or thresholded. Correct.
- **LD50 units/route/species** — checked for the classic unit-mixing hazard. Preserved faithfully, including heterogeneous units (µg/kg, g/kg). Correct.
- **ClinicalTrials `totalCount`** — top suspicion for a page-size-as-total bug; `countTotal=true` is sent and totals match the raw API exactly. Clean.
- **FAERS disproportionality** — 2×2 recomputed independently from openFDA; ROR matched to three decimals.
- **KEGG flat-file parsing** — checked for continuation-line truncation; `hsa04110`'s 158 genes and `hsa04910`'s 138 come back intact.
- **`geo_search_datasets` invalid field tags** — the code sends `[study_type]`/`[platform]` where the real tags are `GTYP`/`RGPL`, and NCBI silently downgrades both to `[All Fields]`. Latent fragility, but no wrong answer could be produced (counts and records matched the correct query), so it is recorded rather than patched.
- **DailyMed `drug_name` resolution, date filters, and section counts; Reactome literature bare-int drops; ENCODE biosample organism path; GEO GSM→Series substitution; UCSC track truncation** — all verified real, all outside the nine shipped here. They are self-contained and well-characterised; I stopped at nine to keep this PR reviewable rather than expanding scope further.

### DIFF
--- a/src/tooluniverse/ctg_tool.py
+++ b/src/tooluniverse/ctg_tool.py
@@ -510,13 +510,16 @@ class ClinicalTrialsSearchTool(ClinicalTrialsTool):
                 query_params[k] = arguments[k]
 
         # Add default parameters that are not shown in the schema.
-        # Skip filter.advanced when a status filter is set, because filter.advanced
-        # requires HasResults=true which excludes most RECRUITING/active trials.
-        has_status_filter = "filter.overallStatus" in query_params
+        # filter.advanced used to be skipped whenever a status filter was set,
+        # to work around its old AREA[HasResults]true clause. That clause was
+        # removed above, so the skip only disabled the documented "Limited to
+        # trials beyond phase 1" gate -- and it did so silently: adding the
+        # narrowing overall_status=COMPLETED filter to a metformin/lactic
+        # acidosis search *raised* total_count from 2 to 3 and returned a
+        # disjoint set, while a RECRUITING pembrolizumab search returned 745
+        # studies instead of 557, phase-1 trials included.
         for k, v in self.default_params_not_shown.items():
             if k not in query_params:
-                if k == "filter.advanced" and has_status_filter:
-                    continue
                 query_params[k] = v
 
         # Process list parameters that need to be joined
--- a/src/tooluniverse/data/ensembl_ld_tools.json
+++ b/src/tooluniverse/data/ensembl_ld_tools.json
@@ -1,7 +1,7 @@
 [
   {
     "name": "EnsemblLD_get_ld_variants",
-    "description": "Get linkage disequilibrium (LD) data for a variant from the Ensembl REST API. Returns all variants in LD with the query variant in a specified 1000 Genomes population, with r-squared and D' statistics. LD measures the non-random association of alleles at different loci, critical for GWAS interpretation and fine-mapping. Example: rs1042779 in KHV population returns 708 LD partners with r2 and D' values.",
+    "description": "Get linkage disequilibrium (LD) data for a variant from the Ensembl REST API. Returns variants in LD with the query variant in a specified 1000 Genomes population, with r-squared and D' statistics. LD measures the non-random association of alleles at different loci, critical for GWAS interpretation and fine-mapping. Results are sorted by r2 (strongest first) and capped at `limit` (default 200); `total_ld_count` always reports how many partners passed the thresholds and `truncated` flags whether any were withheld -- raise `limit` to retrieve them all. Example: rs1042779 in KHV population returns 708 LD partners with r2 and D' values.",
     "parameter": {
       "type": "object",
       "properties": {
@@ -26,6 +26,14 @@
             "null"
           ],
           "description": "Minimum D' threshold. Default: none. Set e.g., 0.8 for high D' only."
+        },
+        "limit": {
+          "type": [
+            "integer",
+            "null"
+          ],
+          "minimum": 1,
+          "description": "Maximum number of LD partners to return, strongest r2 first. Default: 200. Increase to retrieve every partner when `truncated` is true."
         }
       },
       "required": [
@@ -62,8 +70,17 @@
                 "population": {
                   "type": "string"
                 },
+                "total_ld_count": {
+                  "type": "integer",
+                  "description": "Total LD partners passing the thresholds, before truncation"
+                },
                 "ld_count": {
-                  "type": "integer"
+                  "type": "integer",
+                  "description": "Number of LD partners actually returned in ld_variants"
+                },
+                "truncated": {
+                  "type": "boolean",
+                  "description": "True when total_ld_count exceeds ld_count; raise `limit` to retrieve the rest"
                 },
                 "ld_variants": {
                   "type": "array",
--- a/src/tooluniverse/data/gene_ontology_tools.json
+++ b/src/tooluniverse/data/gene_ontology_tools.json
@@ -329,7 +329,7 @@
   },
   {
     "name": "GO_get_genes_for_term",
-    "description": "Finds all genes/proteins associated with a specific Gene Ontology term using the Biolink API.",
+    "description": "Finds genes/proteins annotated to a Gene Ontology term, including annotations to its descendant terms (is_a/part_of closure), via the GO Solr index. Rows are collapsed to distinct genes, each carrying its evidence codes and annotation count. Supply `taxon` to restrict to one species -- omit it and genes from all species are returned together. Example: id='GO:0006915' with taxon='NCBITaxon:9606' returns human apoptosis genes.",
     "type": "GeneOntologyTool",
     "parameter": {
       "type": "object",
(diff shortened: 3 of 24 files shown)


gasvn:

PR #401, opened 2026-07-27
Route mcp_client_tool diagnostics to the logger instead of stdout (stdio safety)

boldakov:
## Summary

`mcp_client_tool.py` writes several diagnostic messages with bare `print()`
calls. In stdio transport mode, stdout is the JSON-RPC channel, so any extra
bytes written there corrupt the protocol stream. Two of these prints run at
object construction time (`MCPClientTool.__init__` and
`MCPAutoLoaderTool.__init__`), which places unframed text into the channel
before the first response is even sent.

This PR routes every `print()` in the module through the package logger,
matching the convention already used elsewhere in the codebase.

## Why this matters

The repository already has the right machinery for this. `logging_config.py`
exposes `get_logger()`, and `reconfigure_for_stdio()` redirects logging to
stderr for stdio mode, with a docstring stating it exists so that "all logs go
to stderr instead of stdout". Modules such as `smcp.py`, `tool_registry.py`,
`execute_function.py` and `pubmed_tool.py` already obtain their logger through
`get_logger()`. `mcp_client_tool.py` is the one module in this path that still
prints directly to stdout, so it bypasses that redirection.

The construction-time prints are the most disruptive part, because a loader
that is always constructed at startup (for example an auto-loader with no
required API key) will emit those lines into the stdio stream during handshake.

## Change

- Import `get_logger` from `.logging_config` and create a module-level
  `logger`.
- Replace every `print()` with the matching logger call: `debug` for the init
  diagnostics, `info` for discovery and registration progress, `error` for
  failure paths.
- Message text is unchanged, so console output is identical once logging is
  configured; it simply goes to stderr rather than stdout.

No behavior changes beyond the output stream, and no public API changes.

## Verification

- The module compiles and imports cleanly, with no circular import
  (`logging_config` does not import `mcp_client_tool`).
- Constructing `MCPClientTool` and `MCPAutoLoaderTool` and capturing stdout:
  282 bytes were written to stdout before this change, and 0 bytes after.
- `grep` confirms no remaining `print(` in the module.

Happy to adjust the log levels or wording if you would prefer a different
split. Thanks for maintaining ToolUniverse.

### DIFF
--- a/src/tooluniverse/mcp_client_tool.py
+++ b/src/tooluniverse/mcp_client_tool.py
@@ -15,7 +15,10 @@ from mcp.client.session import ClientSession
 from mcp.client.streamable_http import streamablehttp_client
 from .base_tool import BaseTool
 from .tool_registry import register_tool
+from .logging_config import get_logger
 import os
+
+logger = get_logger(__name__)
 
 
 class BaseMCPClient:
@@ -154,7 +157,7 @@ class MCPClientTool(BaseTool, BaseMCPClient):
 
         # Debug logging for transport configuration
         tool_name = tool_config.get("name", "Unknown")
-        print(
+        logger.debug(
             f"MCP Tool Init: {tool_name} -> transport={self.transport}, server={self.server_url}"
         )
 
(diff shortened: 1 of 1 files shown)


boldakov:
For context: this fix also unblocks the stdio integration tests on #402 (a keyless Genomic Intelligence auto-loader), which is the first always-on `MCPAutoLoaderTool` and therefore the first config to hit the construction-time `print()` calls in stdio mode. The two are kept separate so this one can be reviewed on its own. Thanks again.

PR #32, opened 2025-10-27
Prototype VSD source harvesting

SufianTA:
Harvest & dynamic REST infrastructure

Created the harvest/ package plus dynamic_rest_runner.py to manage generated REST/GraphQL tools at runtime.
Added Verified Source tooling (VerifiedSourceRegisterTool, VerifiedSourceDiscoveryTool, VerifiedSourceRemoveTool) so freshly harvested endpoints can be registered, listed, and removed on the fly.
Generated specs are stored under ~/.tooluniverse/vsd/generated_tools.json and load immediately through the new runner.

PR #413, opened 2026-07-31
Add safe explicit Verified Source Directory core

gasvn:
## Summary

This extracts the useful Verified Source Directory idea from #32 into a small, explicit, provider-neutral workflow:

- discover a packaged catalog of reviewed public JSON sources
- explicitly probe and register one source
- list and remove locally registered sources
- query only a previously registered source
- expose all five operations through the generated Python SDK wrappers

The branch merges the original #32 history before replacing its implementation, so @SufianTA's commits and authorship remain in the Git history. Please use **Create a merge commit** when landing this PR; squashing would discard that preserved record.

## Safety and design changes from #32

The original PR combined dynamic REST generation, arbitrary URL harvesting, automatic tool execution, and Docker provisioning. Those parts are deliberately excluded here.

The replacement core:

- performs HTTPS GET requests only, on port 443
- uses an exact built-in host allowlist; extra exact hosts require `TOOLUNIVERSE_VSD_ALLOWED_HOSTS`
- rejects URL credentials, endpoint query strings, fragments, IP literals, and credential-like parameters
- validates DNS results, the actual connected peer IP, and every redirect against non-global/private addresses
- disables environment proxy inheritance to avoid bypassing peer validation
- accepts JSON responses only and caps the decompressed body at 1 MB
- limits redirects, timeouts, parameter counts/sizes, metadata sizes, and control characters
- persists the local catalog atomically with mode `0600`
- does not provision containers, harvest arbitrary URLs, generate tools dynamically, or auto-run anything

## Verification

- `pytest -q --no-cov --tb=short tests/unit` — passed on the final commit; 11 documented environment/resource skips
- focused VSD tests — 24 passed, including SSRF, redirect, peer-IP, response-size, credential, metadata, persistence, and end-to-end ToolUniverse coverage
- wrapper/import/discoverability regression group — 87 passed
- Ruff format and lint — passed
- JSON config and metadata validation — passed
- generator metadata check — all five VSD wrappers reported unchanged
- real CLI test against WHO Global Health Observatory:
  - offline discovery returned the packaged WHO source
  - registration returned HTTP 200 and persisted the catalog with mode `0600`
  - a separate CLI process queried the registered endpoint successfully and returned a real indicator record
  - list and remove completed successfully

## Relationship to #32

This PR is intended to supersede #32 for the safe VSD core only. The original PR should remain open until this replacement lands so the contributor record and discussion are not lost. Docker provisioning and automatic harvesting/execution would require separate designs and security reviews.


SufianTA:
I completed an adversarial follow-up and opened #416 with the concrete fixes and
reproducible evidence. The important issues are destination pinning before any
request bytes are sent, redirect and content-encoding rejection, a real total
deadline, strict JSON handling, cross-process catalog locking, explicit duplicate
replacement, non-cacheable catalog operations, and correct mutation annotations.

Verification is complete: 56 focused VSD tests and 60 adjacent tests pass on a
clean merge with current `main`; the live case study completed six calls across
WHO, CDC, and openFDA; a four-provider smoke test also covered Ensembl and both
IPv4 and IPv6 pinned peers. The full hosted Python 3.12 workflow on #416 is green,
including lint, SDK generation, core unit/integration tests, and the doctor CLI.

My recommendation is to merge hardening commit `c7966510` into this branch before
#413 is merged. Alternatively, #416 is a mergeable combined draft against `main`
that preserves the #413 work and adds the fixes, tests, validation note, and
checked case-study artifacts.

PR #454, opened 2026-08-05
Fix double-wrapped return_schema across ~485 tools

gasvn:
## Summary
- `tu test` validates a tool's unwrapped `result["data"]` against `return_schema` (`cli.py`), so `return_schema` must describe the payload itself, not the `{status, data, metadata}` envelope
- ~470 tools had `return_schema` written as a `oneOf` whose first branch required `data` (describing the envelope); a second variant across 15 more files used a flat (non-`oneOf`) `{"type":"object","properties":{"data":...}}` schema with the same underlying mistake — neither was caught by a single detection pass, hence the two batches in this PR's history
- Hoisted each affected schema up one level to describe the real payload shape directly, following the already-correct `gwas_search_associations` pattern (`gwas_tools.json`)
- No `run()` logic changed — tools were already returning correct data; this only fixes their self-test, which had been failing on every single run regardless of whether the tool actually worked
- Also allows `null` for `RDKit_pharmacophore_features`'s `interpretation` counts, which are intentionally `null` when `include_features` excludes that metric (see the `Fix-R4A-5` comment in `rdkit_cheminfo_tool.py`) — the schema declared them non-nullable, failing validation on documented, correct behavior

## Verification
- `python scripts/test_all_tools.py --parallel`: schema-invalid dropped from 101 to 0 across the full catalog (3258 tools, 5136 test examples)
- Spot-verified via `jsonschema.validate` directly and `tu test` for a cross-section (`gwas_search_snps`, `PharmGKB_search_genes`, `ClinGen_get_actionability_adult`, `RDKit_pharmacophore_features`, `annotate_variant_multi_source`, and others)
- Re-ran the original smoke-tested tools after the fix to confirm no regressions

Companion PRs from the same investigation: schema-validation-target fix in `scripts/test_new_tools.py`, 5 tools with a real "missing `data` key" code bug, and an FDA FAERS enum-stringification fix — filed separately since they're independent bugs, not the schema-envelope pattern this PR addresses.

### DIFF
--- a/src/tooluniverse/data/allen_brain_tools.json
+++ b/src/tooluniverse/data/allen_brain_tools.json
@@ -584,137 +584,100 @@
       }
     ],
     "return_schema": {
-      "oneOf": [
-        {
-          "type": "object",
-          "description": "Per-structure quantified expression values for the ISH dataset",
-          "properties": {
-            "status": {
-              "type": "string"
-            },
-            "data": {
-              "type": "array",
-              "items": {
-                "type": "object",
-                "properties": {
-                  "id": {
-                    "type": "integer",
-                    "description": "StructureUnionize record ID"
-                  },
-                  "section_data_set_id": {
-                    "type": "integer",
-                    "description": "Source ISH experiment (SectionDataSet) ID"
-                  },
-                  "structure_id": {
-                    "type": "integer",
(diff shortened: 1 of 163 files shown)


gasvn:

PR #455, opened 2026-08-05
Fix 5 tools returning result without a data key on success

gasvn:
## Summary
- `SemanticScholar_get_pdf_snippets`, `fda_pharmacogenomic_biomarkers`, `ArXiv_get_pdf_snippets`, `DailyMed_get_spl_by_setid`, and `PubTator3_LiteratureSearch` all built a valid success payload but never wrapped it in a `data` key, so `result["data"]` was unconditionally `None` on every call regardless of the query — not a stale test fixture, a real gap in the success-path return
- Wrapped the existing payload under `data` alongside the pre-existing top-level keys (kept for compatibility) to match the `return_schema` contract and every sibling method in these files

## Verification
- `tu test <ToolName>` passes for all 5 tools
- Confirmed sibling tools sharing the same generic implementation (`PubTator3_EntityAutocomplete`, `ArXiv_search_papers`) still pass after the change

### DIFF
--- a/src/tooluniverse/arxiv_tool.py
+++ b/src/tooluniverse/arxiv_tool.py
@@ -380,10 +380,10 @@ class ArXivPDFSnippetsTool(BaseTool):
                 total_chars += len(snippet)
                 found += 1
 
-        return {
-            "status": "success",
+        payload = {
             "pdf_url": final_pdf_url,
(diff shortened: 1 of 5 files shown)


gasvn:

PR #457, opened 2026-08-05
Align test_new_tools.py schema validation with cli.py's tu test

gasvn:
## Summary
- `return_schema` describes the unwrapped `data` payload only. `tu test` (`cli.py`) enforces this correctly: `jsonschema.validate(result.get("data"), return_schema)`
- `scripts/test_new_tools.py` instead had a heuristic (`_schema_describes_envelope`) that tried to auto-detect envelope-shaped schemas and validate the full result instead, keyed on seeing `data` or `status` in any `oneOf`/`anyOf` branch's properties
- That heuristic produced repeated false positives on already-correct schemas: a legitimate error branch mirroring the real `{"status": "error", "error": "..."}` response, ENCODE's own unrelated `status` field (e.g. `"released"`), and IDR's own legitimately-nested `data` key all fooled it into validating the wrong target
- No keyword signal reliably distinguishes "this branch describes the envelope" from "this tool's real payload happens to use this field name" — removed the heuristic entirely; always validate `result["data"]`, matching `tu test` exactly

## Verification
- Full-catalog regression (`scripts/test_all_tools.py --parallel`) before/after this change: schema-invalid did not increase for any already-correct tool
- Several previously-hidden false negatives surfaced for the first time (tools that only "passed" under the old heuristic because it validated a coincidentally-permissive wrong target) — those are fixed in #454, which was found and verified using this corrected validation target

### DIFF
--- a/scripts/test_new_tools.py
+++ b/scripts/test_new_tools.py
@@ -108,32 +108,6 @@ def validate_against_schema(
         return False, str(e)
 
 
-def _schema_describes_envelope(schema: Dict[str, Any]) -> bool:
-    """Heuristic: schema describes the {status,data,error,metadata} envelope.
-
-    Many tool configs declare return_schema at the envelope level (top-level
-    properties include 'data' or 'status'). For those, validating the
-    unwrapped result.get('data') against the schema always fails ("Schema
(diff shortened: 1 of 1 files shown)


gasvn:

PR #459, opened 2026-08-05
PDBe-KB ligand_sites/interface_residues: fix silent truncation

gasvn:
## Summary
- Both `PDBe_KB_get_ligand_sites` and `PDBe_KB_get_interface_residues` hardcoded an output cap (50 ligands, 30 partners) with no transparency and no way to override it, while the response's own `total_ligands`/`total_partners` field reported the real, larger count — making the truncation invisible unless a caller happened to compare the two numbers
- For EGFR (P00533), the default cap silently dropped both erlotinib (AQ4) and gefitinib (IRE) — the two most clinically relevant kinase inhibitors — out of 245 known ligands
- Added optional `max_ligands`/`max_partners` parameters (higher defaults: 100 and 60) and top-level `truncated`/`truncation_note` fields when the cap is still hit, matching the project's existing truncation-transparency convention (see `CancerPrognosis_get_gene_expression`)

## Verification
- `tu run PDBe_KB_get_ligand_sites '{"uniprot_accession":"P00533"}'`: now correctly reports `truncated: true` with a note; `max_ligands=300` returns all 245 entries including AQ4 and IRE
- `python scripts/test_new_tools.py pdbe_kb`: schema-valid (on top of #454's schema fix)

**Note**: `tu test`/`scripts/test_new_tools.py` for these two tools currently fail independent of this change, due to a pre-existing `return_schema` envelope bug fixed separately in #454. This fix was verified directly via `tu run` + manual `jsonschema` validation against `result["data"]`, and is correct and mergeable regardless of #454's merge order.

### DIFF
--- a/src/tooluniverse/data/pdbe_kb_tools.json
+++ b/src/tooluniverse/data/pdbe_kb_tools.json
@@ -104,13 +104,17 @@
   },
   {
     "name": "PDBe_KB_get_ligand_sites",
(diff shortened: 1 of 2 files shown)


gasvn:

PR #461, opened 2026-08-05
PDBeSearch_search_structures: deduplicate results by pdb_id

gasvn:
## Summary
- PDBe's search index is entity-level, not deduplicated by PDB entry — a structure with multiple matching chains/entities can appear as several Solr docs sharing one `pdb_id`, and the tool returned each one verbatim
- Querying "erlotinib" with `limit=10` returned only 6 distinct structures represented across 8 rows, with two entries duplicated
- Over-fetch (up to 3x the requested limit, capped at 150) and deduplicate by `pdb_id` before truncating to the caller's limit, so the caller usually still gets the full requested count of unique structures instead of a count inflated by duplicate entity rows

## Verification
- `tu run PDBeSearch_search_structures '{"query":"erlotinib","limit":10}'`: now returns 6 unique `pdb_id`s with no duplicates (previously 8 rows / 6 unique)
- `tu test PDBeSearch_search_structures` passes

### DIFF
--- a/src/tooluniverse/pdbe_search_tool.py
+++ b/src/tooluniverse/pdbe_search_tool.py
@@ -88,9 +88,14 @@ class PDBeSearchTool(BaseTool):
 
         limit = min(arguments.get("limit", 10), 50)
 
+        # PDBe's Solr index is entity-level, not deduplicated by PDB entry --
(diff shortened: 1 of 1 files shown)


gasvn:

PR #462, opened 2026-08-05
DisGeNET_search_disease: correct description that falsely claims free text works

gasvn:
## Summary
- The tool's description and `disease` parameter both claimed a plain disease name would work (e.g. "disease name or UMLS CUI"), but the implementation is CUI-only by design — a free-text name (e.g. "rheumatoid arthritis") is always rejected with "not a UMLS CUI"
- Correct the description/parameter text to say a CUI is required and point callers at `umls_search_concepts` to resolve one, rather than promising behavior the tool doesn't have

## Verification
- `tu run DisGeNET_search_disease '{"disease":"rheumatoid arthritis"}'` still correctly errors (unchanged behavior — this PR only corrects the misleading docs, not the tool's actual CUI-only requirement)
- `tu test DisGeNET_search_disease` passes with a real CUI

### DIFF
--- a/src/tooluniverse/data/disgenet_tools.json
+++ b/src/tooluniverse/data/disgenet_tools.json
@@ -69,7 +69,7 @@
   },
   {
     "name": "DisGeNET_search_disease",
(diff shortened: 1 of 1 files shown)


gasvn:

PR #463, opened 2026-08-05
Consolidated fixes: return_schema envelope bug (~485 tools), 5 code-level bugs, 4 independently-discovered tool bugs

gasvn:
Consolidates #454, #455, #456, #457, #459, #460, #461, #462 into a single PR. Those are being closed in favor of this one; see each for the detailed rationale/verification behind its piece if useful.

## 1. `return_schema` envelope bug (~485 tools)
`tu test` validates a tool's unwrapped `result["data"]` against `return_schema` (`cli.py`), so `return_schema` must describe the payload itself, not the `{status, data, metadata}` envelope. ~470 tools had a `oneOf`-wrapped schema requiring `data`; a second variant across 15 more files used a flat (non-`oneOf`) shape with the same mistake. Hoisted each to describe the real payload directly, following the already-correct `gwas_search_associations` pattern. No `run()` logic changed — these tools already returned correct data; this only fixes their self-test.

## 2. `scripts/test_new_tools.py`: removed an unreliable envelope-detection heuristic
The script had a heuristic trying to auto-detect envelope-shaped schemas to decide whether to validate the full result or just `data`. It produced repeated false positives (a legitimate error branch mirroring the real `{"status":"error","error":...}` shape; ENCODE's own unrelated `status` field; IDR's own legitimately-nested `data` key). Removed it — the script now always validates `result["data"]`, exactly matching `cli.py`.

## 3. Five tools returning no `data` key on success (real code bug, not schema)
`SemanticScholar_get_pdf_snippets`, `fda_pharmacogenomic_biomarkers`, `ArXiv_get_pdf_snippets`, `DailyMed_get_spl_by_setid`, `PubTator3_LiteratureSearch` built a valid payload but never wrapped it in `data` — `result["data"]` was unconditionally `None`. Fixed at the source.

## 4. RDKit nullable fields + FAERS unmapped-code stringification
`RDKit_pharmacophore_features`'s per-feature counts are intentionally `null` when `include_features` excludes that feature — schema declared them non-nullable. FAERS count-endpoint tools leaked a raw int through when an openFDA code wasn't in the 3-value documented enum map — now stringified so any unmapped code satisfies the schema's string contract.

## 5. Four tools found via a second role-play test round (real, independently-verified bugs)
- **PDBe-KB `ligand_sites`/`interface_residues`**: hardcoded truncation (50/30 items) with no transparency, silently dropping erlotinib/gefitinib for EGFR out of 245 known ligands. Added `max_ligands`/`max_partners` params + `truncated`/`truncation_note` fields.
- **`BVBRC_search_amr`**: silently accepted an unrecognized `organism` parameter and returned unfiltered cross-species data with no error. Added `additionalProperties: false`.
- **`PDBeSearch_search_structures`**: returned duplicate PDB entries (PDBe's index is entity-level). Now deduplicates by `pdb_id`.
- **`DisGeNET_search_disease`**: description falsely claimed free-text names work; it's CUI-only by design. Corrected the docs.

## Verification
- `python scripts/test_all_tools.py --parallel`: schema-invalid dropped from 101 to 0 across the full catalog (3258 tools, 5136 test examples)
- Every fix independently CLI-verified against live data before being included (not just "tests pass") — reproduction steps are in each superseded PR's description
- All 8 superseded PRs individually confirmed `MERGEABLE` before being combined here

### DIFF
--- a/scripts/test_new_tools.py
+++ b/scripts/test_new_tools.py
@@ -112,15 +112,22 @@ def _schema_describes_envelope(schema: Dict[str, Any]) -> bool:
     """Heuristic: schema describes the {status,data,error,metadata} envelope.
 
     Many tool configs declare return_schema at the envelope level (top-level
-    properties include 'data' or 'status'). For those, validating the
-    unwrapped result.get('data') against the schema always fails ("Schema
-    Mismatch: At root: ..."). Detect that style and validate the full result
-    instead. Inner-data schemas (no 'data'/'status' at top) keep the
-    historical unwrapped behaviour.
+    properties include 'data'). For those, validating the unwrapped
+    result.get('data') against the schema always fails ("Schema Mismatch:
+    At root: ..."). Detect that style and validate the full result instead.
+    Inner-data schemas (no 'data' at top) keep the historical unwrapped
+    behaviour.
 
     Looking for 'error' alone is NOT enough — many inner-data schemas pair an
     inner array with a {'error': str} error wrapper but still describe the
     inner data shape, not the envelope.
+
+    Looking for 'status' alone is NOT enough either — a correctly-shaped
+    inner-data schema's *error* branch legitimately mirrors the tool's real
+    error response, {"status": "error", "error": "..."}, without describing
+    the envelope at all. Treating 'status' as an envelope signal on its own
+    misclassifies that branch and validates the full result against a schema
+    meant only for the unwrapped payload, so require 'data' specifically.
     """
     if not isinstance(schema, dict):
         return False
@@ -129,7 +136,7 @@ def _schema_describes_envelope(schema: Dict[str, Any]) -> bool:
         if not isinstance(br, dict):
             continue
         props = br.get("properties") or {}
-        if "data" in props or "status" in props:
+        if "data" in props:
             return True
     return False
 
--- a/scripts/test_new_tools.py
+++ b/scripts/test_new_tools.py
@@ -108,39 +108,6 @@ def validate_against_schema(
         return False, str(e)
 
 
-def _schema_describes_envelope(schema: Dict[str, Any]) -> bool:
-    """Heuristic: schema describes the {status,data,error,metadata} envelope.
-
-    Many tool configs declare return_schema at the envelope level (top-level
-    properties include 'data'). For those, validating the unwrapped
-    result.get('data') against the schema always fails ("Schema Mismatch:
-    At root: ..."). Detect that style and validate the full result instead.
-    Inner-data schemas (no 'data' at top) keep the historical unwrapped
-    behaviour.
-
-    Looking for 'error' alone is NOT enough — many inner-data schemas pair an
-    inner array with a {'error': str} error wrapper but still describe the
-    inner data shape, not the envelope.
-
-    Looking for 'status' alone is NOT enough either — a correctly-shaped
-    inner-data schema's *error* branch legitimately mirrors the tool's real
-    error response, {"status": "error", "error": "..."}, without describing
-    the envelope at all. Treating 'status' as an envelope signal on its own
-    misclassifies that branch and validates the full result against a schema
-    meant only for the unwrapped payload, so require 'data' specifically.
-    """
-    if not isinstance(schema, dict):
-        return False
-    branches = schema.get("oneOf") or schema.get("anyOf") or [schema]
-    for br in branches:
-        if not isinstance(br, dict):
-            continue
-        props = br.get("properties") or {}
-        if "data" in props:
-            return True
-    return False
-
-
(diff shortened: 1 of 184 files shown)


gasvn:
Added one more commit after an independent adversarial re-review of the fixes above surfaced two real bugs, both now fixed and verified:

- `PDBe_KB_get_ligand_sites`/`get_interface_residues`: `max_ligands=0`/`max_partners=0` were silently replaced by the default (falsy-value bug), and negative values sliced from the end of the list instead of being rejected. Fixed with explicit `None` checks and clamping.
- `PDBeSearch_search_structures`: the fixed 3x over-fetch multiplier wasn't enough for entity-dense queries (e.g. "ribosome 70S", where one structure contributes dozens of matching Solr docs) -- `limit=10` could return as few as 1 unique structure. Replaced with adaptive pagination that pages through Solr until `limit` unique entries are found, and clarified that `total_found` is an entity-level Solr count rather than a unique-PDB-entry count.

Verified via `tu test` (all passing) and direct `tu run` reproduction of the original failing cases.

gasvn:
One more commit: found 9 additional tools sharing the exact same double-enveloped \`return_schema\` bug as the original batch (\`coding_variant_fraction\`, \`VCF_summary_stats\`, \`VCF_count_variants\`, \`VCF_normalize\`, \`run_deseq2_analysis\`, \`RNAseq_edger_limma_de\`, \`phykit_batch_analysis\`, \`clinical_trial_ae_severity_test\`, \`read_executed_notebook\`). They use an unusual \`type: object\` + \`oneOf\` combination that the original catalog-wide scan didn't match, so they slipped through undetected -- and were further masked by requiring local file inputs or missing test examples, so \`tu test\` silently skipped them instead of failing.

Hoisted each success branch's \`data\` sub-schema to the top level, same pattern as the rest of this PR. Verified against real tool output (not just static schema inspection) for the ones I could exercise without external system dependencies (bcftools/R+Bioconductor weren't available in this environment): \`coding_variant_fraction\`, \`phykit_batch_analysis\` (single + batch), \`clinical_trial_ae_severity_test\` (prepare + chi-square), and \`read_executed_notebook\` all ran against synthetic fixtures and their real output now validates against the fixed schema. Also gave \`coding_variant_fraction\` the \`requires_local_input\` flag its siblings have -- it was missing it despite its test_examples pointing at a nonexistent file path.

gasvn:
CI failures were 3 pre-existing unit tests written against the old, buggy return_schema convention this PR fixes -- not new bugs. All three asserted against the double-wrapped envelope shape (or a sub-schema that only existed because of it):

- \`test_unknown_code_passes_through\` (FAERS): asserted an unmapped drug-characterization code stays an int, but the tool's own \`return_schema\` declares \`term: {"type": "string"}\` (matching every mapped label, which are all strings). Updated the assertion to the string form, which is what the earlier \`openfda_adv_tool.py\` fix in this PR now correctly produces.
- \`test_dailymed_metadata_allows_null_page_urls\`: validated a \`return_schema.properties.metadata\` sub-schema that only existed because \`DailyMed_search_spls\`'s schema used to describe the full envelope. \`cli.py\` only ever validates against \`result.get("data")\`, never \`metadata\` -- this sub-schema was dead weight that got removed when the envelope bug was fixed. Removed the test along with its now-unused helper.
- \`test_unwrapped_response_satisfies_return_schema\` / \`test_without_extract_path_would_double_wrap\` (iCite): validated the full \`{status, data}\` envelope against a schema that (correctly, post-fix) describes only the bare \`data\` array -- which is what the test file's own docstring already said the contract should be. Switched both to validate \`result["data"]\`, matching what \`cli.py\` actually checks.

Verified: all 17 tests across the 3 affected files pass, plus a full \`pytest tests/unit\` run (2600+ tests) is clean modulo expected environment skips (bcftools/edgeR/torch not installed here) and one test that's flaky under full-suite thread contention but passes in isolation (\`test_direct_import_pattern\`, unrelated to this PR's changes).

gasvn:
One more commit: fixed the incidental \`PDBe_get_compound_summary\` crash on \`comp_id=HEM\` found during the full 505-tool verification sweep above. Root cause: PDBe's API returns some list fields as JSON \`null\` rather than omitting the key (HEM's \`cross_links\` is \`null\`), and \`.get(key, [])\` only applies its default when the key is *absent* -- not when the value is \`None\`. Iterating \`None\` raised \`TypeError\`. Switched \`smiles\`/\`cross_links\`/\`systematic_names\`/\`first_observed_in\` to \`.get(key) or []\`.

Confirmed pre-existing on unmodified \`main\` (reproduces identically), unrelated to the return_schema fixes in this PR -- just something I noticed while verifying those. Both \`tu test\` examples (ATP, HEM) now pass.

PR #464, opened 2026-08-06
Let UCSC_get_sequence take a written locus, and echo both coordinate conventions

gasvn:
## Problem

`UCSC_get_sequence` forwarded `start`/`end` straight to the UCSC REST API, which uses **0-based half-open** coordinates. A locus as written everywhere else — `chr14:89000000-89000100` in the UCSC browser, IGV, `samtools faidx`, and papers — is **1-based inclusive** and spans 101 bases.

Passing the written numbers through unconverted returns **100 bases starting one position late**.

The schema did say "0-based". But the natural action is to paste the locus, and the failure is silent: the sequence comes back the right shape, off by one base at the 5′ end.

## How it surfaced

A LAB-Bench2 DbQA2 item asks for the hg38 reference sequence at `chr14:89000000-89000100`. The returned sequence was *exactly* the expected answer with its leading base missing:

```
expected (101 nt): ATCTTGTCACTATGGTGGGG...
returned (100 nt):  TCTTGTCACTATGGTGGGG...
```

Verified against the live API:

```
start=89000000 end=89000100 -> len=100  matches expected: False
start=88999999 end=89000100 -> len=101  matches expected: True
```

## Changes

- **`region`** — a written locus (1-based inclusive, commas allowed) that can be pasted verbatim, removing the conversion step entirely.
- **`coordinate_system`** — `"0-based"` (default) or `"1-based"`, for the explicit `chrom`/`start`/`end` form, so the caller states the convention rather than converting by hand.
- **Both conventions echoed** in the response — `region_1based`, `start_0based`, `end`, `requested_length` alongside `length` — so a mix-up is visible instead of silently changing the sequence.
- **Range validation matches the convention in use**: 1-based allows `start == end` (a single base); 0-based still requires `end > start`.

## Compatibility

The default path is unchanged. `chrom`/`start`/`end` with no `coordinate_system` still means 0-based half-open and returns exactly what it did before — confirmed byte-identical in testing.

## Verification

| case | result |
|---|---|
| `region: "chr14:89000000-89000100"` | 101 nt, matches expected |
| explicit `coordinate_system: "1-based"` | 101 nt, matches expected |
| legacy 0-based path | 100 nt, byte-identical to previous behavior |
| commas (`chr14:89,000,000-89,000,100`) | 101 nt |
| malformed region / reversed range / bad `coordinate_system` / missing args | clear, specific errors |

### DIFF
--- a/src/tooluniverse/data/ucsc_genome_tools.json
+++ b/src/tooluniverse/data/ucsc_genome_tools.json
@@ -300,7 +300,7 @@
   },
   {
     "name": "UCSC_get_sequence",
-    "description": "Get DNA sequence from the UCSC Genome Browser for a specified genomic region. Returns the nucleotide sequence for a given chromosome, start, and end position in a genome assembly. Useful for primer design, sequence analysis, and regulatory element inspection. Example: hg38, chr17, start=7668421, end=7668521 returns 100bp of TP53 sequence.",
+    "description": "Get DNA sequence from the UCSC Genome Browser for a genomic region. Pass 'region' as a written locus (e.g. 'chr14:89000000-89000100', 1-based inclusive, 101 bp) to paste coordinates verbatim, or give chrom/start/end with 'coordinate_system'. Returns the nucleotide sequence plus the resolved region in both conventions. Useful for primer design, sequence analysis and regulatory element inspection.",
     "parameter": {
       "type": "object",
       "properties": {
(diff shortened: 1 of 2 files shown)


gasvn:
Superseded by #473, which consolidates this with the other five benchmark-driven fixes into a single branch. No changes dropped — the merged branch was re-verified (53 unit tests + 12 behavioural checks). Branch `pr/ucsc-coordinates` is left in place for reference.

PR #465, opened 2026-08-06
Make phykit saturation and treeness/RCV actually run, and surface their errors

gasvn:
## Problem

Four defects made `phykit` batch analysis quietly useless for its two most-requested metrics.

1. **`saturation` could never run.** phykit takes the alignment via `-a/--alignment`, but the tool passed it positionally, so phykit exited 2 with `the following arguments are required: -a/--alignment`.
2. **Errors were swallowed.** `_run_phykit` returned `None` on failure with stderr discarded. A completely broken batch reported `0 files (0 errors)` — indistinguishable from "no input files".
3. **`treeness_over_rcv` was not a supported function**, despite being what questions phrased "treeness/RCV" ask for. (`treeness` is a different metric.)
4. **The tree was passed only for `saturation`.** `treeness_over_rcv` needs both `-a` and `-t`, so it died with `required: -t/--tree`.

Also fixed: the batch stem builder used `Path.stem`, which drops only the *last* suffix. These files carry multi-part extensions (`foo.faa.mafft.clipkit`), so it built a tree path that never existed.

Defects 3 and 4 were each found *by the error surfacing added in defect 2* — the batch run printed the exact argparse failure. Without it, every one of these looked like "no files found".

## Skill correction

The phylogenetics skill said to use the ClipKit-trimmed alignment for all tree-paired metrics. That is right for `saturation` and **wrong for `treeness_over_rcv`**: RCV measures compositional variability across alignment columns, so trimming changes it — and treeness/RCV divides by it.

Verified on the fungal scogs set (249 orthologs, canonical shipped files):

| | untrimmed `.faa.mafft` | trimmed `.clipkit` |
|---|---|---|
| median treeness/RCV | **0.2683** | 0.3050 |
| max treeness/RCV (>70% gapped columns) | **0.1866** | 0.4205 |

Plain `treeness` takes no alignment and reproduces **exactly** (median 0.0501 on the same 249 files) — which isolates the alignment choice as the cause rather than the tree set or the tool.

The skill now carries a per-metric alignment table, an explicit `phykit_batch_analysis` invocation passing alignment and tree extensions separately, and a note that "gap percentage" here means the fraction of *columns containing at least one gap*: under the residue-fraction reading no ortholog exceeds 70%, so a ">70% gaps" filter silently selects nothing.

## Verification

- `saturation` — median **0.6146** on the fungal set (independent reference 0.62)
- `treeness` — median **0.0501**, exact match to reference
- `treeness_over_rcv` — median **0.2683** with the untrimmed alignment
- TU batch tool vs. raw phykit CLI over the same 249 files: **0.3050 vs 0.3050**, confirming the tool's aggregation is correct and the earlier gap was purely the alignment argument
- Missing trees are now reported as per-file errors (fungi: 6 of 255 alignments lack trees; animals: 141 of 241) rather than vanishing

### DIFF
--- a/skills/tooluniverse-phylogenetics/SKILL.md
+++ b/skills/tooluniverse-phylogenetics/SKILL.md
@@ -148,7 +148,6 @@ that line when the raw min in a group is 0.
 scogs zips ship in two shapes:
 - **Full**: `<gene>.faa`, `<gene>.faa.mafft`, `<gene>.faa.mafft.clipkit`,
   `<gene>.faa.mafft.clipkit.treefile`, plus iqtree/bionj/log/mldist.
-  Use clipkit alignment + treefile for tree-paired metrics.
 - **Alignment-only**: just `<gene>.faa` + `<gene>.faa.mafft`. No
   trees, no clipkit. Used for parsimony, RCV, gap-percentage
   questions. Use the `.faa.mafft` (NOT raw `.faa`) — the published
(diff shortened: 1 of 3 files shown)


gasvn:
Superseded by #473, which consolidates this with the other five benchmark-driven fixes into a single branch. No changes dropped — the merged branch was re-verified (53 unit tests + 12 behavioural checks). Branch `pr/phykit-fixes` is left in place for reference.

PR #466, opened 2026-08-06
Reach MSigDB mouse collections, and route mouse-phenotype lookups to them

gasvn:
## Problem

`MSigDBTool` hardcoded the `/human/` endpoint. MSigDB serves human (H, C1–C8) and mouse (MH, M1–M8 — including the `MP_*` mammalian-phenotype sets) from **separate endpoints**, so every mouse gene set failed.

The failure was also confusing: MSigDB answers an unknown name with an HTML error page rather than a 404, so the tool surfaced a raw `JSONDecodeError` instead of "not found".

Found by tracing which tools were actually invoked on mouse-phenotype benchmark questions — the accuracy number alone gave no indication the mouse path was dead.

## Changes

- `_fetch_geneset(name, species)` targets a species-specific endpoint and treats a JSON decode failure as "not served here" rather than a transport error.
- `species` parameter: `auto` (default) tries human then mouse, so mouse-only sets like `MP_*` resolve without the caller knowing which collection owns them; `human`/`mouse` pin it explicitly.
- The resolved species is echoed in the response, so a caller can tell which collection answered.
- Not-found now reports the gene set name and which collections were tried.
- Skill routing: mouse-phenotype questions go to the MSigDB `MP_*` sets, with MGI as fallback.
- Efficiency: fetch a gene set once rather than once per candidate answer option.

## Verification

20/20 mouse-phenotype lookups validated end-to-end. `_check_gene_in_set` threads `species` through, so membership tests work against mouse sets too.


gasvn:
Superseded by #473, which consolidates this with the other five benchmark-driven fixes into a single branch. No changes dropped — the merged branch was re-verified (53 unit tests + 12 behavioural checks). Branch `pr/msigdb-mouse` is left in place for reference.

PR #467, opened 2026-08-06
Let single-operation tools accept their own documented example without 'operation'

gasvn:
## Problem

148 tools across 41 classes rejected the most natural way to call them.

These tools dispatch internally on an `operation` argument, but each config declares exactly one operation in its `fields`. Calling such a tool with only its documented parameters — including the arguments from its **own `test_examples`** — failed validation with a missing-`operation` error. The caller had to restate a value the config already fixed and the schema often did not surface.

This is a systemic defect rather than 148 separate ones: the information needed to fill `operation` was already present in every affected config.

## Change

`_apply_operation_default()` runs in `execute_function` before validation. When `operation` is absent, it is filled from the tool's own `fields.operation`, else from the schema default. Tools that genuinely take multiple operations, and any call that passes `operation` explicitly, are unaffected.

The same defaulting is applied to the `DNA_*` tools so each defaults to its own operation.

## Verification

A guard test asserts that **every** single-operation tool accepts its own `test_examples` payload with `operation` omitted — so this cannot regress silently as tools are added.

### DIFF
--- a/src/tooluniverse/data/addgene_tools.json
+++ b/src/tooluniverse/data/addgene_tools.json
@@ -11,7 +11,8 @@
           "enum": [
             "search_plasmids"
           ],
-          "description": "Operation type"
+          "description": "Operation type",
+          "default": "search_plasmids"
         },
         "query": {
           "type": "string",
@@ -38,7 +39,6 @@
         }
       },
       "required": [
-        "operation",
         "query"
       ]
     },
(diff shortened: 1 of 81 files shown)


gasvn:
Superseded by #473, which consolidates this with the other five benchmark-driven fixes into a single branch. No changes dropped — the merged branch was re-verified (53 unit tests + 12 behavioural checks). Branch `pr/operation-default` is left in place for reference.

PR #468, opened 2026-08-06
Cut Type IIS enzymes at their real offsets, search both strands, and simulate Golden Gate assembly

gasvn:
> Stacked on #467 (`pr/operation-default`) — review that one first; this PR's diff against it is the DNA work alone.

## Problem

Type IIS restriction enzymes cut **outside** their recognition site at a fixed offset (BsaI `GGTCTC(N1)`, BbsI `GAAGAC(N2)`, Esp3I/BsmBI `CGTCTC(N1)`, SapI `GCTCTTC(N1)`). Two bugs made every Golden Gate simulation wrong:

1. **`_resolve_enzyme` read `enz.fst`, an attribute Biopython does not have.** The correct name is `fst5`. The lookup silently fell through, so every fallback enzyme was assigned a **midpoint** cut — meaning BsaI, BbsI, Esp3I and SapI all cut *inside* their own recognition site.
2. **Sites were searched on the forward strand only.** A Golden Gate donor carries its two Type IIS sites **inverted** around the insert, so the reverse-strand site was never found and a correct digest reported 1 fragment instead of 2.

The combination is quietly wrong rather than loudly broken: a digest returns *some* fragments, just not the right ones.

## Changes

- `_resolve_enzyme` uses `fst5`/`ovhg`, so each enzyme cuts at its true offset with the correct overhang polarity.
- `_biopython_cut_positions()` searches **both strands** via `enz.search()`.
- `_reverse_complement()` promoted to a module-level helper.
- **New `DNA_golden_gate_assemble` operation** — answers "what does this reaction produce?". It digests each input (circular by default), drops fragments that retain a recognition site (those are re-cut in a one-pot reaction and cannot persist), chains the rest by matching 4-bp overhangs, and returns `product_sequence`, `product_length` and `assembly_order` with the overhang at every junction.
- Molecular-cloning skill documents the working-backwards workflow and the diagnostic that a correct Golden Gate donor digest yields **2** fragments — getting 1 means the enzyme name or `circular` flag is wrong, not that the plasmid lacks sites.

## Verification

38 tests, including per-enzyme cut-offset assertions against known sites and a full assemble-and-annotate round trip.

## Honest limitation

Digestion and overhang-driven ligation are simulated faithfully; **reaction efficiency is not modelled** (no overhang ligation bias, partial digestion, or incorrect-but-possible junctions). A returned product is the intended assembly, not a yield prediction.


gasvn:
Superseded by #473, which consolidates this with the other five benchmark-driven fixes into a single branch. No changes dropped — the merged branch was re-verified (53 unit tests + 12 behavioural checks). Branch `pr/dna-type-iis` is left in place for reference.

PR #469, opened 2026-08-06
Route fact-lookup and patient-specific drug-choice questions to tools

gasvn:
## Problem

On clinical-vignette questions — "which drug for this patient" — the agent made **zero tool calls**, answering from memory alone. The `fact-lookup` skill never fired, because the router had no rule matching either shape.

Found by counting tool invocations per question rather than reading accuracy: the accuracy number looked unremarkable while the tools were not being consulted at all.

## Change

Two routing lines in the ToolUniverse router skill, sending fact-lookup questions and patient-specific drug-selection questions to the tools that can answer them.

## Verification

Explicit routing takes tool invocation on the affected CURE-Bench slice from **0 to 11**.

Small diff, but the failure it fixes is invisible in headline accuracy — the agent was guessing where it had a database available.


gasvn:
Superseded by #473, which consolidates this with the other five benchmark-driven fixes into a single branch. No changes dropped — the merged branch was re-verified (53 unit tests + 12 behavioural checks). Branch `pr/router-routing` is left in place for reference.

PR #470, opened 2026-08-06
Fix Round 3: GDC mutation-frequency inflation, ClinicalTrials/cBioPortal error-message fixes, PharmGKB drug filter

gasvn:
## Summary

Test-harness round 3 (post PR #463): internal regression pass + 2 fresh role-play personas (clinical pharmacogenomicist, cancer genomics/molecular oncologist) exercising domains not covered by prior rounds (pharmgkb/cpic/admetai dosing guidance; civic/cbioportal/opentarget/clinical_trials tumor-board workflow). Every reported issue was CLI-verified against the live upstream API before being queued; two of the persona's lower-confidence reports (PharmGKB unsupported-parameter schema, ADMETAI dependency-version noise) turned out to already be handled gracefully or to be local-environment artifacts, not code defects, and were left alone. A `/simplify` pass (4 parallel review agents: reuse, simplification, efficiency, altitude) found and fixed two sibling instances of the same bug pattern already living in the touched files (cBioPortal's CNA profile lookup, and the ClinicalTrials NCT-detail extraction tools), plus some redundant loop-state and duplicated-fallback cleanup.

- **GDC_get_mutation_frequency_by_project**: the numerator counted SSM-occurrence records instead of distinct cases, wildly inflating mutation frequency for some project/gene pairs (confirmed: KRAS in TCGA-TGCT reported 97% when only ~5% of cases actually carry the mutation, a ~20x error). Now derives the numerator from deduped `case_id`s across paginated `/ssm_occurrences` records.
- **search_clinical_trials / extract_clinical_trial_adverse_events (and siblings)**: a query that legitimately matched zero trials returned the same hardcoded error as a genuine API failure, making the two indistinguishable and breaking callers that only parse JSON on success. Both now return a structured empty success result when real data was fetched but the query (or field) is legitimately empty.
- **cBioPortal_get_mutations / get_copy_number_alterations**: an invalid `study_id` silently fell through to a guessed molecular-profile-id and only surfaced as a confusing raw 404 several steps later. Both now report the real problem immediately and point at `cBioPortal_get_cancer_studies`.
- **PharmGKB_get_drug_label_annotations**: listing mode had no way to filter by drug name, so finding one drug's label meant paging through up to 500 entries by hand. Added an optional `drug_name` filter.

## Test plan
- [x] CLI-reproduced each issue against the live upstream API before fixing (GDC ground-truth cross-checked against raw `/ssm_occurrences` and `/cases`/`/projects` data)
- [x] Existing/updated unit tests: `test_gdc_mutation_frequency.py`, `test_cancer_genomics_depth.py`, `test_ctg_tool.py`, `test_cbioportal_tool.py` — new regression tests added for each fixed behavior
- [x] Targeted regression sweep (12 test files covering all 4 changed modules plus schema-conformance and workflow tests): 117 passed, 0 failed
- [x] Live CLI re-verification of all 6 fixed code paths post-fix
- [x] Independent `tool-validator` agent gate: 11/11 checks pass

### DIFF
--- a/src/tooluniverse/cbioportal_tool.py
+++ b/src/tooluniverse/cbioportal_tool.py
@@ -1,5 +1,5 @@
 import requests
-from typing import Any, Dict
+from typing import Any, Callable, Dict, Optional
 from urllib.parse import quote
 from .base_tool import BaseTool
 from .tool_registry import register_tool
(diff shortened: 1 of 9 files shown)


gasvn:
Consolidated into #472 along with round 4.

PR #471, opened 2026-08-06
Fix Round 4: GWAS sort/alias/dedup, IntAct empty details, OmniPath filter

gasvn:
## Summary

Round 4 of the test-harness role-play cycle, covering GWAS Catalog, IntAct, and OmniPath.

- **GWASAssociationSearch**: default `sort=p_value` whenever no explicit sort is given (not just when a p-value filter is present), matching sibling tool `gwas_get_associations_for_trait` and ensuring a client-side p-value threshold actually sees the top hits (previously `disease_trait=SLE, p_value_threshold=1e-100` returned 0 hits despite real hits existing at p=2e-298).
- **GWASStudyByID**: accept `accession_id` as an alias for `study_id` -- every sibling tool and every association record uses `accession_id` for this same GCST identifier.
- **GWASSNPsForGene**: dedupe by `rsId` -- the v1 `findByGene` endpoint repeats identical SNP records (confirmed byte-for-byte duplicates), inflating apparent SNP counts.
- **intact_get_interaction_details**: this tool was missing from the EBI Search fallback's per-tool query-field config entirely, so every call -- including its own documented example -- silently returned empty. Also fixed its `return_schema` (described an array; the endpoint returns one object) and its test example (used a single-part ID that doesn't match the tool's own documented `EBI-XXXXXX-EBI-YYYYYY` format and never matched a real record).
- **OmniPath_get_dorothea_regulon**: `dorothea_level` wasn't in the requested API fields, so every `confidence_levels` filter silently matched nothing; the field is also a list of codes (an interaction can carry more than one level), not a scalar, so filtering/grouping needed set membership instead of equality.

## Verification

All CLI-verified live against the real upstream APIs (not just schema inspection):

- `gwas_search_associations` with `disease_trait=Systemic lupus erythematosus, p_value_threshold=1e-100` now returns the p=2e-298 hit instead of 0 results
- `gwas_get_study_by_id` with `accession_id=GCST000392` returns the study (previously would have errored "study_id is required")
- `gwas_get_snps_for_gene` for TP53: 22 duplicate records removed, 28 unique SNPs returned
- `intact_get_interaction_details` with a real interaction ID (`EBI-714308-EBI-366083`, fetched live from `intact_search_interactions`) now returns the actual record instead of an empty result
- `OmniPath_get_dorothea_regulon` with `confidence_levels=A` now returns matching interactions instead of an empty list
- `tu test` passes for all 6 affected tools (`gwas_search_associations`, `gwas_get_study_by_id`, `gwas_get_snps_for_gene`, `intact_get_interaction_details`, `OmniPath_get_dorothea_regulon`, `Reactome_get_complex`)

### DIFF
--- a/src/tooluniverse/data/gwas_tools.json
+++ b/src/tooluniverse/data/gwas_tools.json
@@ -649,18 +649,20 @@
   {
     "type": "GWASStudyByID",
     "name": "gwas_get_study_by_id",
-    "description": "Get a specific GWAS study by its unique identifier.",
+    "description": "Get a specific GWAS study by its unique identifier. Accepts either 'study_id' or 'accession_id' (the same GCST identifier field name used by GWASAssociationsForStudy and returned on every association record) so IDs can be chained directly between gwas_* tools without renaming.",
     "parameter": {
       "type": "object",
       "properties": {
         "study_id": {
           "type": "string",
-          "description": "GWAS study identifier"
+          "description": "GWAS study identifier (GCST accession, e.g. 'GCST000392')"
+        },
+        "accession_id": {
+          "type": "string",
+          "description": "Alias for study_id -- the same GCST accession identifier, accepted so results from other gwas_* tools (which use the field name 'accession_id') can be passed straight in."
         }
       },
-      "required": [
-        "study_id"
-      ]
+      "required": []
     },
     "label": [
       "GWAS",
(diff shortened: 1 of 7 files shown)


gasvn:
Consolidated into #472 along with round 3.

PR #472, opened 2026-08-06
Consolidated Rounds 3-4: GDC/ClinicalTrials/cBioPortal/PharmGKB + GWAS/IntAct/OmniPath fixes

gasvn:
## Summary

Consolidates rounds 3 and 4 of the test-harness role-play cycle into one PR (originally #470 and #471).

**Round 3** (cancer genomics / clinical pharmacogenomics personas):
- `GDC_get_mutation_frequency_by_project`: numerator was counting SSM-occurrence records instead of distinct cases, inflating frequencies up to ~20x (KRAS in TCGA-TGCT reported 97% vs. the real ~5%)
- `search_clinical_trials` / clinical-trial-detail extraction tools: a legitimate zero-match result was indistinguishable from an API failure
- `cBioPortal_get_mutations` / `get_copy_number_alterations`: an invalid `study_id` silently fell through to a guessed profile ID, surfacing only as a confusing raw 404 later
- `PharmGKB_get_drug_label_annotations`: added a `drug_name` filter to listing mode

**Round 4** (GWAS / molecular interaction personas):
- `GWASAssociationSearch`: default `sort=p_value` whenever no explicit sort is given, so a p-value threshold actually sees the top hits (previously returned 0 hits for a real association at p=2e-298)
- `GWASStudyByID`: accept `accession_id` as an alias for `study_id`
- `GWASSNPsForGene`: dedupe by `rsId` (the API repeats identical records)
- `intact_get_interaction_details`: was missing from a config map entirely, so every call -- including its own documented example -- silently returned empty; also fixed its `return_schema` (array -> object) and its test example (a fake ID that didn't match the tool's own documented format)
- `OmniPath_get_dorothea_regulon`: the field a confidence-level filter matched against was never requested from the API, so the filter always matched nothing; the field is also a list, not a scalar

## Verification

All fixes CLI-verified live against the real upstream APIs. `tu test` passes for every affected tool; full relevant unit-test files (`test_gdc_mutation_frequency.py`, `test_cancer_genomics_depth.py`, `test_cbioportal_tool.py`, `test_ctg_tool.py`) pass locally (26 tests). Full CI is currently blocked repo-wide by a GitHub Actions platform outage (githubstatus.com confirms `major_outage` on Actions) -- will pick up automatically once GitHub recovers.

Supersedes #470 and #471, which are being closed in favor of this single PR.

### DIFF
--- a/src/tooluniverse/cbioportal_tool.py
+++ b/src/tooluniverse/cbioportal_tool.py
@@ -1,5 +1,5 @@
 import requests
-from typing import Any, Dict
+from typing import Any, Callable, Dict, Optional
 from urllib.parse import quote
 from .base_tool import BaseTool
 from .tool_registry import register_tool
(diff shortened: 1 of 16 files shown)


gasvn:
CI failure was one pre-existing test written against the narrower, pre-fix behavior -- not a new issue. \`test_no_pvalue_filter_leaves_sort_unset\` asserted \`gwas_search_associations\` leaves \`sort\` unset without a p-value filter, but this PR's round-4 fix intentionally broadened the sort=p_value default to apply whenever no explicit sort is given at all. Verified that's the correct call: sibling tools \`GWASAssociationsForTrait\` and \`GWASVariantsForTrait\` already default to sort=p_value unconditionally (the latter's own comment, pre-existing on main, documents the exact same inconsistency for a different endpoint), so a plain \`gwas_search_associations\` call was the odd one out, returning an arbitrary API-order slice while its near-identical siblings didn't. Updated the test to assert the new, now-consistent default.

Full local \`pytest tests/unit\` (2600+ tests) is clean aside from expected environment skips (bcftools/torch/edgeR not installed here) and one test that's flaky only under full-suite thread contention (unrelated to this PR).

PR #473, opened 2026-08-07
Tool and skill fixes found by running ToolUniverse against biomedical benchmarks

gasvn:
Consolidates #464, #465, #466, #467, #468 and #469 into one branch. Every change here came from tracing which tools an agent actually invoked on benchmark questions and finding the tool broken, rather than from reading code.

That method matters, because **none of these were visible in accuracy numbers.** A tool that returns a plausible-but-wrong sequence, or reports `0 files (0 errors)`, looks like a model mistake.

## 1. UCSC coordinates — silently off by one base (was #464)

`UCSC_get_sequence` forwarded `start`/`end` straight to the UCSC REST API, which is **0-based half-open**. A locus as written everywhere else — `chr14:89000000-89000100` in the browser, IGV, `samtools`, papers — is **1-based inclusive** and spans 101 bases.

Found on a LAB-Bench2 item where the returned sequence was *exactly* the expected answer with its leading base missing. Confirmed against the live API:

```
start=89000000 end=89000100 -> 100 nt  matches expected: False
start=88999999 end=89000100 -> 101 nt  matches expected: True
```

Adds `region` (paste a locus verbatim) and `coordinate_system` for the explicit form, and echoes `region_1based` / `start_0based` so a mix-up is visible. **The default path is byte-identical to before.**

## 2. PhyKIT — five defects (was #465)

1. `saturation` passed the alignment positionally; phykit requires `-a`, so it exited 2 and could never run.
2. Errors were swallowed — stderr discarded, so a fully broken batch reported `0 files (0 errors)`, indistinguishable from "no input".
3. `treeness_over_rcv` was not in the supported list, though it is what "treeness/RCV" questions ask for.
4. The tree was passed only for `saturation`, so treeness/RCV died with `required: -t/--tree`.
5. **Batch ran strictly serially** — 249 alignments × ~2.2 s ≈ 9 min for one metric, past the 900 s budget. Six benchmark questions failed as timeouts with the tool working correctly.

Defects 3 and 4 were each found *by the error surfacing added for defect 2*. Fixing diagnostics first paid for itself repeatedly.

Batch now uses a thread pool: **544 s → 35.9 s (~15×)**, with median and error count identical to serial and to an independent reference.

**Skill correction.** The phylogenetics skill said to use the ClipKit-trimmed alignment for all tree-paired metrics. Right for saturation, wrong for treeness/RCV — RCV measures variability across alignment columns, so trimming changes it:

| | untrimmed `.faa.mafft` | trimmed `.clipkit` |
|---|---|---|
| median treeness/RCV | **0.2683** (reference 0.26) | 0.3050 |
| max, >70% gapped columns | **0.1866** (reference 0.21) | 0.4205 |

Plain `treeness` takes no alignment and reproduces exactly (median 0.0501), which isolated the alignment choice as the cause rather than the tool.

## 3. MSigDB — mouse collections unreachable (was #466)

Hardcoded `/human/`. MSigDB serves mouse (MH, M1–M8, including `MP_*` phenotype sets) from a separate endpoint, so every mouse lookup failed — and surfaced a raw `JSONDecodeError`, because MSigDB answers unknown names with an HTML page rather than a 404. Adds `species` (`auto` tries human then mouse), echoes which collection answered. 20/20 mouse-phenotype lookups validated.

## 4. `operation` parameter — 148 tools rejected their own examples (was #467)

Tools that dispatch on `operation` but declare exactly one in `fields` rejected calls that omitted it — including the arguments from their **own `test_examples`**. One systemic fix in `execute_function`, plus a guard test asserting every single-operation tool accepts its own example with `operation` omitted.

## 5. Type IIS enzymes cut in the wrong place (was #468)

`_resolve_enzyme` read `enz.fst`, **an attribute Biopython does not have** (it is `fst5`). The lookup fell through, so every fallback enzyme got a *midpoint* cut — meaning BsaI, BbsI, Esp3I and SapI all cut inside their own recognition site. Sites were also searched on the forward strand only, so the inverted site in a Golden Gate donor was never found and a correct digest reported 1 fragment instead of 2.

Adds `DNA_golden_gate_assemble` to answer what a reaction produces. 38 tests.

*Honest limitation:* digestion and overhang ligation are simulated faithfully; **reaction efficiency is not modelled**, so a returned product is the intended assembly, not a yield prediction.

## 6. Routing for fact-lookup and drug vignettes (was #469)

On clinical vignettes the agent made **zero tool calls**, answering from memory. Explicit routing takes tool invocation on that slice from **0 to 11**.

The vignette procedure lives in `tooluniverse-biomedical-fact-lookup`, keyed by **modifier → label section**, not by tool name — there are 76 `FDA_get_*_by_drug_name` tools, so any fixed list is arbitrary. Brand→generic resolution comes first, with its reason: two options are sometimes the same drug (verified: Sajazir and Firazyr both resolve to icatibant), which eliminates both.

## Verification on the merged branch

- 53 unit tests pass
- 12 behavioural checks re-run post-merge: UCSC gold match + legacy path unchanged, phykit aliases/tree-flag/parallelism, MSigDB species, operation defaulting, `fst5` and Golden Gate, and all three skill edits

Three config files conflicted with main's restyled `return_schema`. Resolved by taking main's schema and re-deriving the two semantic changes (drop `operation` from `required`, add `fields.operation`) from each tool's own `const` — so none of main's schema work is reverted.

## What is not claimed

Benchmark deltas above were measured before I discovered that this environment's editable install pointed at a different checkout, so I am not restating headline accuracy figures here. The tool-level facts — API responses, cut offsets, timings, test results — were all verified directly against the merged branch.


gasvn:
Merged `fix/consolidated-475-474` (#476) into this branch to pick up the GWAS sort-default test fix.

CI on the previous commit failed on exactly one test, `test_gwas_pvalue_filter_sort.py::test_no_pvalue_filter_leaves_sort_unset`, which is **not from this PR** — this branch changes no GWAS files, its copies of the test and tool are byte-identical to main, and a pristine `origin/main` worktree fails the same test. #476 is the designated fix for it.

The two PRs touch **no files in common** (93 vs 16), so the merge was clean with zero conflicts. #476 remains open and independently reviewable; git will reconcile the duplicated commits when it lands.

Verified after merging: 87 tests pass across both PRs' surfaces, and `ruff check` is clean.

PR #474, opened 2026-08-07
Fix Round 5: CTD/metabolite RENCI-backend migration, NHANES zero-corruption, LipidMaps and Metabolomics Workbench fixes

gasvn:
## Summary

Round 5 of the test-driven quality harness, this time targeting metabolomics, proteomics/mass-spec, and environmental toxicology/exposure -- domains not covered by rounds 1-4. Three researcher personas ran real research questions against these domains; every reported finding was independently reproduced via `tu run`/`tu test` before being queued for a fix (several persona claims were discarded as false positives after not reproducing). One issue (the CTD backend outage) was found independently via direct CLI sampling of existing test_examples, ahead of the persona reports, and a persona later reproduced it too.

- **CTD tools + `Metabolite_get_diseases`/`HMDB_get_diseases`**: the RENCI Automat mirror all five of these depended on has been fully decommissioned (its own registry no longer lists a `ctd` backend). Migrated the chemical<->disease direction to `mydisease.info` (confirmed live, and it carries real evidence-type data RENCI never had); the gene<->chemical direction now returns an honest "permanently unavailable" error (no free live replacement found) instead of a misleading generic HTTP error.
- **NHANES**: `NHANES_download_and_parse` was silently corrupting every legitimate zero value in numeric columns into `5.397605346934028e-79` (a pandas XPORT-reader artifact for all-zero IBM floats, verified by reproducing it from pandas' own conversion code). `nhanes_search_datasets` silently searched only cycle 2017-2018 when `year` was omitted; now searches the two most recent cycles, matching its sibling tool.
- **LipidMaps**: `LipidMaps_search_by_name`'s own documented examples (`'PC 16:0/18:1'`, `'SM 34:1'`) returned zero results. Added an automatic retry against the `abbrev_chains` field for acyl-chain-specific names, and corrected the sphingomyelin example (needs the `;O2` hydroxyl suffix).
- **Metabolomics Workbench**: `get_study` returned a transposed, confusing shape for every output type (fixed with a dedicated stanza parser); `find_studies_by_phenotype` leaked raw upstream HTML on slash-containing filters (now rejected upfront with guidance); a misleading RefMet example in the "no results" message was corrected; `search_by_mz`/`search_by_exact_mass`'s `Delta` field is now properly zero-padded.

## Test plan

- [x] `tu test` on every changed tool (`CTD_get_chemical_diseases`, `CTD_get_disease_chemicals`, `CTD_get_chemical_gene_interactions`, `CTD_get_gene_chemicals`, `Metabolite_get_diseases`, `HMDB_get_diseases`, `NHANES_download_and_parse`, `nhanes_search_datasets`, `LipidMaps_search_by_name`, `MetabolomicsWorkbench_get_study`, `MetabolomicsWorkbench_find_studies_by_phenotype`, `MetabolomicsWorkbench_get_refmet_info`, `MetabolomicsWorkbench_search_by_mz`, `MetabolomicsWorkbench_search_by_exact_mass`) -- all pass against live APIs
- [x] Rewrote and ran the 4 pre-existing unit test files that mocked the now-removed RENCI-specific internals (`tests/unit/test_ctd_tool.py`, `test_ctd_id_normalization.py`, `test_ctd_disease_predicate_caveat.py`, `test_metabolite_ctd_pubmed_ids.py`) -- 25/25 pass
- [x] Sibling/unaffected tools in each touched file re-verified for no regressions (`LipidMaps_search_by_formula`, `LipidMaps_get_compound_by_id`, `LipidMaps_get_compound_by_xref`, `nhanes_get_dataset_info`, `MetabolomicsWorkbench_search_compound_by_name`, `_get_compound_by_pubchem_cid`, `_get_gene_protein`, `Metabolite_get_info`, `Metabolite_search`, `CTD_get_gene_diseases`)

### DIFF
--- a/src/tooluniverse/ctd_tool.py
+++ b/src/tooluniverse/ctd_tool.py
@@ -1,13 +1,27 @@
-"""CTD (Comparative Toxicogenomics Database) tool — backed by the RENCI Automat
-mirror of CTD's knowledge graph.
+"""CTD (Comparative Toxicogenomics Database) tool.
 
-CTD's native batchQuery.go now requires an altcha proof-of-work CAPTCHA, so
-direct programmatic access via that endpoint is blocked. This tool uses the
-NIH-NCATS-Translator-funded mirror at https://automat.renci.org/ctd/ instead.
-That mirror is **chemical-centric** (June-2024 snapshot, 26k nodes / 166k
-edges, biolink-categorized predicates) — it covers chemical↔gene and
-chemical↔disease cleanly, but does NOT contain CTD's gene-disease inferred
-edges. Gene→disease queries return an honest error pointing at OpenTargets.
+History of this integration, for anyone tempted to "fix" the URL again:
+
+1. CTD's native batchQuery.go now requires an altcha proof-of-work CAPTCHA,
+   so direct programmatic access via that endpoint has been blocked for a
+   long time.
+2. This tool was rewritten to use the NIH-NCATS-Translator-funded RENCI
+   Automat mirror at https://automat.renci.org/ctd/ instead.
+3. That mirror has since been decommissioned outright: its own registry
+   (https://automat.renci.org/registry) no longer lists a "ctd" backend at
+   all (confirmed live -- only cam-kp, robokopkg, translatorkg-gandalf,
+   icees-kg, ubergraph, cohd, translatorkg, ohd, and robomousekg remain).
+   Every request now 404s, not intermittently but always.
+
+Current backend: the BioThings mydisease.info API, which independently
+indexes a cached copy of CTD's chemical-disease curation under
+`ctd.chemical_related_to_disease` on each disease document. This covers the
+chemical<->disease direction only. mychem.info and mygene.info were checked
+live and carry no CTD fields at all, so there is currently no free, live
+replacement for the gene<->chemical direction; those two tools return an
+honest "permanently unavailable" error pointing at DGIdb/OpenTargets instead
+of a raw HTTP error, matching the CTD_get_gene_diseases precedent already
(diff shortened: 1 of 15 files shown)


gasvn:
CI showed 3 failures. Two are real, now fixed here (pre-existing tests written against the old, pre-migration behavior):

- \`test_metabolite_dead_ctd_backend_is_an_error_not_zero_diseases\` / \`test_metabolite_unknown_term_still_returns_empty_not_an_error\`: both mocked \`requests.post\`, but \`_ctd_diseases\` now queries mydisease.info via \`requests.get\` (the decommissioned RENCI mirror it used to POST to). The stale mock target meant \`patch()\` never intercepted anything -- fixed both to patch \`requests.get\`, and updated the second test's mocked response body to mydisease.info's own \`{"hits": [...]}\` shape instead of RENCI's.
- \`test_tsv_body_parsed_into_row_dicts\`: asserted the raw, un-fixed \`"Delta": ".0000"\` value from the mocked upstream body -- the tool now restores the leading zero Metabolomics Workbench's own endpoint omits, so the correct value is \`"0.0000"\`. Updated the assertion.

The third failure (\`test_no_pvalue_filter_leaves_sort_unset\` in \`test_gwas_pvalue_filter_sort.py\`) is **not from this PR** -- this branch never touches GWAS. It's inherited from testing against current \`main\`, which has a known, already-fixed-in-#475 issue (round 4's GWAS sort-default change merged via #472 slightly before that test fix landed). This failure will clear on its own once #475 merges -- no action needed here.

gasvn:
Consolidated into #476 along with #475.

PR #475, opened 2026-08-07
Fix a test written against the pre-fix narrow sort-default scope

gasvn:
## Summary

Follow-up to #472, which merged before this specific fix commit was pushed -- `main`'s test suite currently fails `test_no_pvalue_filter_leaves_sort_unset`.

That test asserted `gwas_search_associations` leaves `sort` unset when no p-value filter is given -- true before #472, which intentionally broadened the `sort=p_value` default to apply whenever the caller gives no explicit sort at all (not only when a p-value filter is present). Verified that broadening was correct: sibling tools `GWASAssociationsForTrait` and `GWASVariantsForTrait` already default to `sort=p_value` unconditionally, so a plain `gwas_search_associations` call was the odd one out, returning an arbitrary API-order slice while its near-identical siblings didn't.

Updated the test to assert the new, now-consistent default instead of the old exception.

## Verification

- `pytest tests/unit/test_gwas_pvalue_filter_sort.py` -- 3/3 pass
- Confirmed the test currently fails on unmodified `main` (reproduced locally before this fix)

### DIFF
--- a/tests/unit/test_gwas_pvalue_filter_sort.py
+++ b/tests/unit/test_gwas_pvalue_filter_sort.py
@@ -1,11 +1,13 @@
-"""Unit test: gwas_search_associations p-value threshold fetches most-significant first.
+"""Unit test: gwas_search_associations defaults to most-significant-first.
 
 Regression: the p_value threshold is applied CLIENT-SIDE to the fetched page, but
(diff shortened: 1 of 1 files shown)


gasvn:
Consolidated into #476 along with #474.

PR #479, opened 2026-08-09
Fix Round 7: plant-genomics and microbiome data-quality issues

gasvn:
## Summary

Five small, independent fixes found this round via persona-driven testing across two domains not covered in prior rounds: model-organism genetics (ZFIN/WormBase/FlyBase/USDA-Plants/Plant-Reactome) and microbiome/metagenomics (MGnify). Every issue below was reproduced directly with `tu run`/`tu test` before being fixed, and every fix was re-verified the same way afterward.

- **USDA_plants_get_profile**: `scientific_name` leaked raw `<i>...</i>` HTML tags straight from the USDA API (e.g. `"<i>Abies balsamea</i> (L.) Mill."`), even though the same file already has a `_strip_html` helper that its sibling `_wetland`/`_search` methods use for the exact same field — `_profile` just never called it. Now fixed to match. Also corrected the tool description: the `PlantProfile` endpoint never actually returns per-state/county FIPS distribution codes (only a `HasDistributionData` boolean flag), so `fips_distribution` was promising data this endpoint structurally cannot provide.
- **PlantReactome_search_pathways**: pathway names leaked raw Reactome ContentService search-highlighting markup (`<span class="highlighting">Flavonoid</span>`), the identical problem the sibling `ReactomeContent_search` tool already solves with its own `_strip_html` helper — added the equivalent here. Also fixed handling of zero-match/invalid queries: this API signals "no results" via a bare HTTP 404 (confirmed live) rather than an empty 200, so a nonsense query previously surfaced as an opaque `"Plant Reactome API HTTP error: 404"`; now returns a clear "no results found for: `<query>`" message, matching the pattern the sibling human-Reactome tool already uses for its own 404s.
- **MGnify_get_study_detail**: `is_public` always came back `null` because the code read a nonexistent `is-public` JSON:API attribute — the real field is `is-private` and needs inverting (confirmed against the live API). `analyses_count`/`downloads_count` were also always `null`, since the `studies/{accession}` endpoint only ever returns relationship *links*, never a `meta.count`; replaced both with `samples_count`, which the endpoint does populate, and updated the description to point callers at `MGnify_list_analyses`/`MGnify_list_analysis_downloads` for exact counts instead of promising fields this endpoint can't fill.
- **MGnifyExpandedTool** (shared by all single-entity lookups: study/genome/analysis/sample): a 404 for a bad accession/ID surfaced as a bare `"MGnify API HTTP error: 404"` with no indication of which input was wrong. The shared error handler now echoes the queried identifier and suggests checking for a typo or a mismatched ID type — fixed once in the common `run()` handler so every endpoint benefits, not just one.

## Test plan

- [x] Every issue reproduced via `tu run`/`tu test` before fixing (see per-tool detail above)
- [x] Every fix re-verified via `tu run`/`tu test` with the original reproduction args after the change
- [x] Shipped `test_examples` still pass for all 5 affected tools (`USDA_plants_get_profile`, `PlantReactome_search_pathways`, `MGnify_get_study_detail`, `MGnify_get_genome`, and other MGnify endpoints spot-checked for the shared 404 fix)
- [x] Targeted regression suite green: `tests/unit/test_usda_plants_tool.py`, `tests/unit/test_plant_agriculture_depth.py`, `tests/unit/test_mgnify_params.py`, `tests/unit/test_epidemiology_public_health_depth.py` (26/26 passed)
- [x] Grepped `tests/` for assertions on the old (buggy) field values/behavior before removing them — none exist; the removed `analyses_count`/`downloads_count` fields and the old raw-HTML `scientific_name` were never asserted on by name, so nothing needed updating
- [x] Full `tests/` sweep (`--no-cov` for speed) shows no new failures attributable to this diff — the 9 pre-existing failures on this run are all in files this PR never touches (`test_claude_code_plugin.py` doc-sync checks, `test_ux_error_hints_r3.py` logging-capture tests, `test_http_api_thread_pool.py`), plus the same `numpy`/`numba`/`onnxruntime` ABI-mismatch collection errors already called out as pre-existing shared-environment issues in the round 6 PR

## Note

Persona testing also surfaced several lower-confidence/higher-scope findings deliberately left out of this PR to keep it focused: `FlyBase_get_gene_alleles`/`ZFIN_get_gene_alleles` returning structurally empty variant details, `WormBase_get_interactions` silently capping results at 150/category with no pagination param, raw 400/500 errors for bad gene IDs on ZFIN/WormBase/FlyBase (Alliance API backend), and the lack of a FlyBase gene-name search tool. Flagging these for a follow-up round rather than expanding this diff.

### DIFF
--- a/src/tooluniverse/data/mgnify_expanded_tools.json
+++ b/src/tooluniverse/data/mgnify_expanded_tools.json
@@ -229,7 +229,7 @@
   },
   {
     "name": "MGnify_get_study_detail",
-    "description": "Get detailed information about a specific MGnify metagenomics study. Returns study name, abstract, associated BioProject, centre, biomes, and counts of analyses/downloads. Complementary to MGnify_search_studies which lists studies. Example: MGYS00002008 is a Tara Oceans marine metagenomics study.",
+    "description": "Get detailed information about a specific MGnify metagenomics study. Returns study name, abstract, associated BioProject, centre, public/private status, sample count, and biomes. Complementary to MGnify_search_studies which lists studies; use MGnify_list_analyses / MGnify_list_analysis_downloads for exact analyses/downloads counts (this endpoint does not provide them). Example: MGYS00002008 is a Tara Oceans marine metagenomics study.",
     "parameter": {
       "type": "object",
       "properties": {
@@ -281,7 +281,13 @@
                 "type": "string"
               }
             },
-            "analyses_count": {
+            "is_public": {
+              "type": [
+                "boolean",
+                "null"
+              ]
+            },
+            "samples_count": {
               "type": [
                 "integer",
                 "null"
--- a/src/tooluniverse/data/usda_plants_tools.json
+++ b/src/tooluniverse/data/usda_plants_tools.json
@@ -1,7 +1,7 @@
 [
     {
         "name": "USDA_plants_get_profile",
-        "description": "Get a USDA PLANTS Database profile for a plant by its PLANTS symbol (e.g. 'ABBA' = balsam fir): scientific name, common name, taxonomic group/rank, duration (annual/perennial), growth habits (tree/shrub/forb), native status, synonyms, and US distribution FIPS codes. Authoritative US plant taxonomy. No API key required.",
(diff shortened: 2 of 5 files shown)


gasvn:

PR #480, opened 2026-08-09
Fix Round 8: OpenFDA API key wiring, HPO/Monarch silent-failure fixes

gasvn:
## Summary

Test-harness round 8: role-play personas in immunogenetics (HLA typing / TCR repertoire), rare-disease diagnosis (HPO / Orphanet / Monarch phenotype-driven differential diagnosis), and nutrition/food-safety science, followed by CLI verification of every reported issue before fixing. Four fixes, all reproduced live and verified after the fix:

- **All 12 `OpenFDA_*` tools never actually used `FDA_API_KEY`.** The config declared it as `optional_api_keys` (gating a "works without it, better with it" log message only), but no code path injected it into requests, so every call — including `OpenFDA_search_food_enforcement` / `OpenFDA_search_food_adverse_events` hit during this round — used the anonymous rate limit and failed with `OVER_RATE_LIMIT` even with a valid key configured. Reuses the existing `auth_param` query-param injection mechanism (already used by the WAQI air-quality tool) instead of adding a new one.

- **`get_joint_associated_diseases_by_HPO_ID_list` silently zeroed out on one bad ID.** It computes candidate diseases by intersecting per-HPO-ID disease sets. A single invalid/obsolete HPO ID in the input list returns zero diseases for that ID alone, and the previous AND-intersection logic let that single empty set collapse the *entire* result to `[]` with no signal that anything was wrong — a clinician entering an otherwise-correct phenotype panel with one typo'd ID would see "no candidate diseases" instead of a still-useful differential. Now excludes zero-hit IDs from the intersection and reports which ones were skipped.

- **`HPO_get_term` silently substituted a different term for obsolete IDs.** Confirmed live: querying `HP:0006887` (obsolete) returns the JAX ontology API's silent redirect target `HP:0001249`'s full record, with no obsolescence flag anywhere in the payload — the sibling tool `get_phenotype_by_HPO_ID` does flag `deprecated: true` for the same ID via a different endpoint. Detects the ID mismatch (`data.id != requested term_id`) and surfaces `requested_id`/`resolved_id`/a note in metadata instead of letting the substitution pass unnoticed.

- **`FDA_get_drug_names_by_food_safety_warnings` searched a nearly-empty field.** It only searched openFDA's `food_safety_warning` field, which — confirmed via `_exists_:food_safety_warning` — is populated on just 3 label records in the entire database. Realistic queries like `grapefruit` or `alcohol` returned real drug names but with the warning field always `null`, and the tool's own documented `test_example` (`field_info: "Instructions"`) returned `NOT_FOUND`. Broadened the search to OR across `food_safety_warning`, `drug_interactions`, `precautions`, `warnings_and_precautions` — the sections where food-drug interaction text (e.g. grapefruit-statin interactions) actually lives — using the same OR-group `search_fields` idiom already used by ~80 other tools in this file.

## Test plan

- [x] CLI-verified each issue live before fixing (`tu run <tool> '<args>'`), and again after
- [x] All 12 `OpenFDA_*` tools' shipped `test_examples` re-run via `tu test <tool>` — all pass, no regressions
- [x] `HPO_get_term`, `get_joint_associated_diseases_by_HPO_ID_list`, `FDA_get_drug_names_by_food_safety_warnings` shipped `test_examples` re-run — all pass, including a new obsolete-ID / invalid-ID test case added to each
- [x] Targeted existing unit tests re-run: `tests/unit/test_fda_label_envelope.py`, `test_openfda_field_value_projection.py`, `test_faers_multifield_or_parens.py`, `test_faers_unknown_param.py`, `test_hpo_phenotype_annotations.py`, `test_hpo_search_terms_limit.py`, `test_hpo_id_normalization.py`, `test_monarch_restful_curie_normalization.py`, `test_monarch_result_id_prefix.py`, `test_confidently_wrong_answers_r6.py`, `test_pharmacovigilance_faers_depth.py`, `test_disease_ontology_depth.py` — all pass
- [x] `/simplify` review pass on all changed files (2 parallel review agents); applied the one safe finding (redundant field lookup in `hpo_tool.py`), explicitly skipped a return-shape-consistency suggestion that would have widened the change's blast radius to the common (non-warning) case

### DIFF
--- a/src/tooluniverse/data/fda_drug_labeling_tools.json
+++ b/src/tooluniverse/data/fda_drug_labeling_tools.json
@@ -5619,13 +5619,13 @@
   },
   {
     "name": "FDA_get_drug_names_by_food_safety_warnings",
-    "description": "Retrieve drug names based on specific food safety warnings.",
+    "description": "Retrieve drug names whose label discusses a food-drug interaction or food safety warning (e.g. 'grapefruit', 'alcohol', 'dairy'). Searches across the label's food-interaction-relevant sections: 'food_safety_warning' (a real but very rarely populated openFDA section -- only a handful of labels use it), plus the much more commonly populated 'drug_interactions', 'precautions', and 'warnings_and_precautions' sections where food/beverage interactions are usually actually documented. Matching a single word like 'alcohol' can return many results since these are broad safety sections, not a dedicated food-warning field.",
     "parameter": {
       "type": "object",
       "properties": {
         "field_info": {
           "type": "string",
-          "description": "Information related to food safety warnings."
+          "description": "Food/beverage or interaction keyword to search for, e.g. 'grapefruit', 'alcohol', 'dairy', 'tyramine'."
         },
         "indication": {
           "type": "string",
@@ -5647,7 +5647,10 @@
     "fields": {
       "search_fields": {
         "field_info": [
-          "food_safety_warning"
+          "food_safety_warning",
+          "drug_interactions",
+          "precautions",
+          "warnings_and_precautions"
         ],
         "indication": [
           "indications_and_usage"
@@ -5666,8 +5669,7 @@
     ],
     "test_examples": [
       {
-        "field_info": "Instructions",
-        "indication": "Hypertension",
+        "field_info": "grapefruit",
         "limit": 1,
         "skip": 0
       }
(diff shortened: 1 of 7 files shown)


gasvn:

PR #481, opened 2026-08-09
Fix Round 9: Ensembl overlap-region/sequence-type defaults, ClinVar zero-result cases, MyVariant field paths, PubChem SMILES key

gasvn:
## Summary

Persona role-play testing this round covered cheminformatics/ADMET, structural-variant/CNV genomics, and comparative-genomics/phylogenetics research questions. Every reported issue was CLI-verified against the live upstream API before being queued as a fix (roughly half of the raw persona reports turned out to be false positives -- upstream data sparsity or service outages, not ToolUniverse bugs -- and were discarded).

Fixes in this PR (all confirmed via `tu run`/`tu test` against live APIs, not mocked):

- **`ensembl_get_structural_variants` / `ensembl_get_overlap_features`** -- both share the generic `EnsemblRESTTool` `/overlap/region` handler, which hardcoded a `"variation"` fallback for the `feature` query parameter regardless of what each tool's own JSON schema declared as its default. Both tools were silently returning dbSNP point variants instead of the structural variants / genes their names and descriptions promise -- confirmed live on the tools' own shipped `test_examples`. The fallback now reads each tool's own schema default.
- **`ensembl_get_sequence`** -- rejected the only enum value that could mean "protein" (`"peptide"`) with an upstream HTTP 400, because Ensembl's REST API actually expects `type=protein`. `"peptide"` is now normalized to `"protein"`, so protein sequence retrieval works via either name.
- **`ClinVar_search_variants`** -- a `condition` search restricted to ClinVar's own indexed disease name (`[dis]`) returned 0 hits for phrasing that's arguably the most natural way to refer to a condition (e.g. "MECP2 duplication syndrome"), even though ClinVar has 100+ matching records under a different indexed name ("Xq28 duplication syndrome"). Now retries unrestricted (`[All Fields]`) once before giving up -- this can only add matches the strict query missed, never drop ones it found. Also adds a `zero_result_hint` when a `gene` filter returns 0 results, since that's otherwise indistinguishable from "this gene truly has no ClinVar entries" whether the symbol was a typo or a protein/full name instead of the HGNC symbol.
- **`MyVariant_get_pathogenicity_scores`** -- requested several dbNSFP field paths that don't match MyVariant.info's actual schema (`dbnsfp.cadd.phred` should be the top-level `cadd.phred` field; `dbnsfp.polyphen2_hdiv.score` should be `dbnsfp.polyphen2.hdiv.score`; PhyloP/PhastCons need the same dotted nesting). CADD, PolyPhen-2, GERP, PhyloP, and PhastCons scores were silently absent from every response despite being named in the tool's own description as the whole point of the tool.
- **`PubChem_get_compound_properties_by_CID`** -- requested the deprecated `"CanonicalSMILES"` property name by default; PubChem's live API now returns that data under the key `"ConnectivitySMILES"` instead, so code reading the documented key name got nothing.
- **`EnsemblCompara_get_orthologues`/`_get_paralogues`/`_get_gene_tree`** -- HTTP error handling dropped the response body, leaving a bare `"HTTP error: 400"` with no indication an unresolved gene symbol was the cause. Now extracts Ensembl's JSON error message (matching the pattern already used in `ensembl_map_tool.py`/`ensembl_variation_ext_tool.py`) plus a targeted hint.

Two persona-reported issues were investigated and found **not** to be ToolUniverse bugs: SwissADME's CYP-inhibitor/iLOGP fields return `"n/d"` because the upstream `swissadme.ch` service itself returns that (confirmed by querying the raw service directly for an unrelated compound); ADMET-AI/RDKit-dependent tools crash in this dev environment due to a NumPy 2.x binary incompatibility with the installed `admet_ai`/`rdkit` wheels, which is an environment issue, not a code defect (the tool already has a lazy-import/graceful-error wrapper around this dependency).

## Test plan

- [x] `tu test` on every changed tool passes (own shipped `test_examples`)
- [x] `pytest tests/unit/test_ensembl_*.py tests/tools/test_ensembl_tool.py tests/unit/test_clinvar_*.py tests/unit/test_phylogenetics_orthology_depth.py tests/unit/test_opentargets_efoid_bridge_and_ensembl_version.py` -- all pass, no regressions
- [x] `ruff check` / `ruff format --check` clean on all changed Python files
- [x] Verified the ClinVar/Ensembl fixes against the live APIs with the exact args the failing tools originally used

### DIFF
--- a/src/tooluniverse/clinvar_tool.py
+++ b/src/tooluniverse/clinvar_tool.py
@@ -259,6 +259,7 @@ class ClinVarSearchVariants(ClinVarRESTTool):
                 gene_hyphen_variant = gene.replace("-", "")
             query_parts.append(f"{gene}[gene]")
 
+        condition_dis_index = None
         if "condition" in arguments:
             # Feature-70B-005: [disease/phenotype] is not a valid ClinVar eSearch field.
             # Use [dis] (disease/phenotype) field tag for condition searches.
@@ -266,6 +267,7 @@ class ClinVarSearchVariants(ClinVarRESTTool):
             condition = arguments["condition"].strip()
             if " " in condition and not condition.startswith('"'):
                 condition = f'"{condition}"'
+            condition_dis_index = len(query_parts)
             query_parts.append(f"{condition}[dis]")
 
         if arguments.get("rsid"):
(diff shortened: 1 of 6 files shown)


gasvn:

PR #482, opened 2026-08-09
Fix Round 10: Golden Gate round-trip, primer-design coverage, KEGG network genes, PyPI metadata

gasvn:
## Summary

Verified via CLI reproduction against issues surfaced by systems-biology (pathway/network modeling), neuroscience/psychiatric-genetics, and synthetic-biology (agricultural trait engineering) role-play sessions.

- **DNA_golden_gate_assemble**: overhangs from a reverse-oriented Type IIS recognition site were read literally off the top strand instead of reverse-complemented, so the tool could never correctly chain fragments produced by its own sibling tool `DNA_golden_gate_design` (explicitly documented as the inverse operation). Confirmed by round-tripping `DNA_golden_gate_design`'s own output through `DNA_golden_gate_assemble` and cross-checking the fix against Biopython's restriction-site cut geometry. Also fixes a `oneOf` `return_schema` ambiguity that made the tool's own `test_examples` fail (`tu test DNA_golden_gate_assemble`).
- **DNA_primer_design**: the forward/reverse primer search windows extended past the boundary they need to respect (`target_start`/`target_end`), so the single best-Tm-scoring candidate in-window could structurally be unable to cover the target, producing a spurious "does not cover the target" failure even when a covering (just slightly worse-scoring) primer existed in the same window.
- **KEGG_get_network**: `N#####`-style perturbed-network records (e.g. `N00151`, TNF-NFKB signaling) list their driving genes and source KEGG pathway map under `GENE`/`PATHWAY`, which the parser didn't recognize at all — `elements` (only meaningful for the other, `nt######`-style network family) stayed empty, silently dropping the gene list for exactly the network type callers need it most for.
- **PackageTool** (backs every `get_<package>_info` tool, e.g. `get_mne_info`, `get_nilearn_info`): `last_updated` used PyPI's internal monotonic change-counter (`last_serial`) as if it were a Unix timestamp, decoding to bogus 1970s dates; `python_versions` returned every PyPI trove classifier unfiltered (license, OS, audience, ...) instead of just Python version entries; `repository` only checked two of the several `project_urls` keys real projects use (e.g. mne-python only sets `Source Code`).
- **WikiPathways_get_pathway_genes / _get_pathway_metabolites**: now accept `wpid` as an alias for `pathway_id`, matching the parameter name used by the sibling `WikiPathways_get_pathway` tool, so switching between the two tools doesn't require renaming the same pathway identifier.
- **graphql_tool.execute_query**: GraphQL error responses were printed to stdout verbatim on failure, including upstream stack traces with internal server file paths (confirmed live via `OpenNeuro_get_dataset` on a nonexistent dataset ID, which surfaced `/srv/packages/openneuro-server/dist/graphql/permissions.js:...` in the CLI output). Now logged at debug level with just the sanitized error message.

## Test plan

- [x] `tu test` passes for every touched tool (`DNA_golden_gate_assemble`, `DNA_golden_gate_design`, `DNA_primer_design`, `DNA_gibson_design`, `WikiPathways_get_pathway_genes`, `WikiPathways_get_pathway_metabolites`, `KEGG_get_network`)
- [x] Live CLI reproduction of each original failure, confirmed fixed
- [x] Full targeted pytest run across all touched files: 116/116 pass
- [x] `tests/tools/test_return_schema_data_shapes.py`: 28/28 pass
- [x] New regression tests added: `test_graphql_tool_error_logging.py`, `test_package_tool_pypi_metadata.py`, plus additions to `test_dna_golden_gate_assemble.py`, `test_kegg_ext_tool_fixes.py`, `test_wikipathways_pathway_genes.py`
- [x] Existing `test_dna_golden_gate_assemble.py` fixture data was hand-crafted against the old (incorrect) overhang convention; updated to build biochemically-correct BsaI donor sequences and confirmed the new behavior against Biopython's own restriction-cut geometry before changing the assertions

### DIFF
--- a/src/tooluniverse/data/dna_tools.json
+++ b/src/tooluniverse/data/dna_tools.json
@@ -986,55 +986,56 @@
         {
           "type": "object",
           "properties": {
-            "status": {
+            "enzyme": {
               "type": "string"
             },
-            "data": {
-              "type": "object",
-              "properties": {
-                "enzyme": {
-                  "type": "string"
-                },
-                "product_length": {
-                  "type": "integer",
-                  "description": "Length of the assembled circular product in bp."
-                },
-                "num_fragments_assembled": {
-                  "type": "integer"
-                },
-                "assembly_order": {
-                  "type": "array",
-                  "description": "Fragments in assembly order, each with source, length and its two overhangs."
-                },
-                "unused_fragments": {
-                  "type": "array",
-                  "description": "Site-free fragments that did not enter the circular product."
-                },
-                "product_sequence": {
-                  "type": "string"
-                }
-              }
+            "product_length": {
+              "type": "integer",
+              "description": "Length of the assembled circular product in bp."
+            },
+            "num_fragments_assembled": {
+              "type": "integer"
+            },
+            "assembly_order": {
+              "type": "array",
+              "description": "Fragments in assembly order, each with source, length and its two overhangs."
+            },
+            "unused_fragments": {
+              "type": "array",
+              "description": "Site-free fragments that did not enter the circular product."
+            },
+            "product_sequence": {
+              "type": "string"
             }
-          }
+          },
+          "required": [
+            "enzyme",
+            "product_length",
+            "num_fragments_assembled",
+            "assembly_order",
+            "product_sequence"
+          ]
         },
         {
           "type": "object",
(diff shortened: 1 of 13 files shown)


gasvn:

PR #483, opened 2026-08-09
Fix Round 11: ClinicalTrials search, BV-BRC surveillance/AMR, WikiPathways, UniProt annotation

gasvn:
## Summary

Cardiology, dermatology, and outbreak-genomics researcher persona role-play surfaced several issues; each was CLI-verified live before being queued for a fix (most persona reports turned out to be agent error or working-as-intended, and were discarded).

- **`ClinicalTrials_search_studies` was completely broken**, including its own shipped test example (`tu test ClinicalTrials_search_studies` failed with `Additional properties are not allowed ('operation' was unexpected)`). Root cause: an earlier round's fix (`Fix-R8C-2`, see `tests/unit/test_clinicaltrials_search_studies_schema.py`) added `additionalProperties: false` to reject unrecognized params, but didn't account for `_apply_operation_default` injecting `"operation"` into arguments before validation for this single-operation tool. Fixed by declaring `operation` as a proper enum schema property, matching the pattern already used elsewhere (e.g. `DNA_find_orfs`).
- **`BVBRC_search_surveillance` looked like it returned duplicate rows** for a subtype/host query. Verified directly against the live BV-BRC API: the rows are genuinely distinct records (separate lab submissions for the same sample, each with its own `id`) that only *looked* identical because no distinguishing field was in `select_fields`. Added `sample_accession`/`id` so callers can tell them apart instead of assuming a pagination bug.
- **`annotate_variant_multi_source` always returned an empty `uniprot.function`** field. `UniProt_search`'s result entries never carry a function comment (only accession/id/protein_name/gene_names/organism/length) — confirmed live. Added a follow-up `UniProt_get_function_by_accession` call keyed on the resolved accession.
- **`WikiPathways_search` returned zero results for "IL-4 signaling"** even though the pathway exists (WP395, titled "IL4 signaling"). The search is a literal substring match and pathway titles use unhyphenated gene shorthand. Normalized hyphens out of both sides of the SPARQL `CONTAINS` comparison.
- **`FDA_get_drug_names_by_mechanism_of_action`'s shipped test example was broken**: it used `"E11.9"` (an ICD-10 diagnosis code) as a mechanism-of-action search string, so it never matched. Replaced with a working example and clarified the parameter description to say it searches mechanism text, not diagnosis codes.
- **`BVBRC_search_amr`'s description didn't scope itself to antibacterial resistance**, so a researcher looking for antiviral resistance mutations got a silent empty result with no pointer to the right tool (`Pathoplexus_get_mutations`).

## Test plan

- [x] `tu test <ToolName>` passes for all 6 affected tools (previously 2 were failing: `ClinicalTrials_search_studies`, `FDA_get_drug_names_by_mechanism_of_action`)
- [x] `tu test` passes for unmodified sibling tools sharing the same code paths (`ClinicalTrials_get_study`, `ClinicalTrials_search_by_intervention`, `ClinicalTrials_search_by_sponsor`, `ClinicalTrials_get_database_stats`, `ClinicalTrials_get_field_values`) — no regression
- [x] Existing pytest suite passes, including the pre-existing `Fix-R8C-2` schema regression guard (`tests/unit/test_clinicaltrials_search_studies_schema.py`) and the `BVBRC_search_amr` description guard (`tests/unit/test_bvbrc_amr_description.py`)
- [x] `ruff check` / `ruff format --check` clean on all modified `.py` files
- [x] Live-verified `BVBRC_search_surveillance`'s "duplicate" records against the real BV-BRC API before concluding it wasn't a pagination bug

### DIFF
--- a/src/tooluniverse/bvbrc_tool.py
+++ b/src/tooluniverse/bvbrc_tool.py
@@ -405,7 +405,17 @@ class BVBRCTool(BaseTool):
 
         limit = min(arguments.get("limit") or 25, 100)
 
+        # Fix-11C-1: BV-BRC's surveillance collection commonly holds multiple
+        # distinct records (e.g. separate lab submissions/panels) that share
+        # identical sample_identifier/collection_date/subtype values. Without
+        # a unique identifier in the response, these genuinely distinct
+        # records looked like a duplication bug (same visible fields
+        # repeated). Including sample_accession (human-readable submission
+        # accession) and id (the record's unique key) lets callers tell
+        # distinct records apart instead of appearing to be verbatim dupes.
         select_fields = [
+            "sample_accession",
+            "id",
             "sample_identifier",
             "collection_date",
             "geographic_group",
--- a/src/tooluniverse/compound_variant_tool.py
+++ b/src/tooluniverse/compound_variant_tool.py
@@ -308,7 +308,34 @@ class CompoundVariantAnnotationTool(BaseTool):
                 if error:
                     sources_failed.append(f"UniProt: {error[:100]}")
                 else:
-                    annotations["uniprot"] = self._parse_uniprot(r, gene_for_gnomad)
+                    parsed = self._parse_uniprot(r, gene_for_gnomad)
+                    # Fix-11B-1: UniProt_search's entries never carry a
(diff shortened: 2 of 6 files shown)


gasvn:

PR #484, opened 2026-08-09
Fix Round 12: geroscience/endocrinology/hematology data-quality issues

gasvn:
## Summary

Test-harness round driven by CLI-verified role-play across three domains not previously covered: geroscience/aging biology, endocrinology/metabolic-disease genetics, and hematology (inherited blood disorders).

- **AgingCohort_search**: relevance filtering matched almost the entire registry (29/30 cohorts) whenever a query contained just one near-universal word like "biomarker" or "study" -- a nonsense query was indistinguishable from a relevant one. Now requires a majority of meaningful query tokens to match.
- **OMIM_search**: returned full OMIM records (~700KB for one query) despite its own description promising a lightweight "MIM numbers, titles, and entry types" index. Now returns just that summary.
- **PanelApp_search_panels**: a clinician searching the singular disease name ("haemoglobinopathy") got zero results because the matching panel is named in the plural. Added a scoped fuzzy-prefix fallback (verified it does NOT over-match, e.g. "myopathy" no longer false-matches "cardiomyopathy"). Also improved description/empty-result guidance for disease names not covered by panel-level metadata.
- **CTD_get_gene_diseases**: its error-redirect suggestion pointed to a tool name that doesn't exist. Now points to `gather_gene_disease_associations`, a real working aggregator.
- **KEGG_get_drug_targets**: description claimed gene-symbol output and cited a wrong example; corrected to match actual behavior.
- **BiomarkerDiscoveryWorkflow**: every default-configured LLM call crashed with a temperature-unsupported error because the default model ("gpt-5") was missing from the reasoning-model classifier used for temperature normalization, even though the sibling max-tokens check already covered it -- consolidated both into one shared classifier. Also fixed the actual root cause of "no genes found for any disease": the HPA search step checked for a top-level `"genes"` key that `call_tool()` never returns at that level (it's nested under `"data"`), so it silently fell through to a hardcoded fallback of unrelated cancer genes with a false success indicator. Now it finds real genes for real disease queries, and reports an honest miss instead of fabricating data when there truly are none.

## Test plan
- [x] Reproduced every issue via `tu run`/`tu test` before fixing
- [x] 103 tests pass across all six touched modules (`tests/unit/test_panelapp_client_side_search.py`, `tests/test_aging_cohort_tool.py`, `tests/unit/test_ctd_tool.py`, `tests/unit/test_openai_compatible_client.py`, `tests/unit/test_biomarker_discovery_compose.py`, `tests/tools/test_omim_tool.py`, `tests/unit/test_gemini_client.py`), including new regression tests for each fix
- [x] `tu test` re-run on every changed tool (`PanelApp_search_panels`, `AgingCohort_search`, `OMIM_search`, `KEGG_get_drug_targets`, `CTD_get_gene_diseases`, `BiomarkerDiscoveryWorkflow`) to confirm no schema/behavior regressions
- [x] Broader unstaged regression pass (pytest across `tests/unit`, `tests/tools`, `tests/api`, minus pre-existing environment-level collection errors from an unrelated `numpy`/`onnxruntime` binary incompatibility affecting `markitdown`-dependent guideline tests) showed no failures attributable to this change

### DIFF
--- a/src/tooluniverse/aging_cohort_tool.py
+++ b/src/tooluniverse/aging_cohort_tool.py
@@ -8,6 +8,7 @@ No external API calls -- uses an internal curated database. Cohort metadata
 changes rarely (study design, variables, and access methods are stable).
 """
 
+import math
 from typing import Dict, Any, List
 from .base_tool import BaseTool
 from .tool_registry import register_tool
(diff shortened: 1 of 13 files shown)


gasvn:

PR #485, opened 2026-08-09
Fix Round 13: ClinicalTrials.gov intervention/sponsor relevance, DisGeNET noise/alias hints, GWAS trait note, DrugBank interaction bloat

gasvn:
## Summary

Role-play sessions this round covered nephrology, gastroenterology, and reproductive-medicine/fertility-genetics research questions. Every reported issue was CLI-verified against live APIs before being queued; several reported issues turned out to be false positives (agent misreadings or expected upstream behavior) and were discarded.

### Verified and fixed

- **`ClinicalTrials_search_by_intervention` / `ClinicalTrials_search_by_sponsor`** (`src/tooluniverse/ctg_tool.py`): CTG API v2's Essie search engine matched `query.intr`/`query.spons` against the whole study record, not just the intervention-name/lead-sponsor-name field. Confirmed live: searching `intervention="vedolizumab"` surfaced an unrelated salivary-gland-carcinoma oncology trial in the top 5 (215 unrestricted hits vs. 134 once restricted to the field, all correctly containing the drug); `sponsor="Takeda"` surfaced Neurocrine Biosciences- and Shire-sponsored trials (1977 vs. 1022 once restricted). Both now wrap the value in Essie's `AREA[...]` operator, mirroring an existing fix already applied to `query.cond`.
- **`DisGeNET_search_gene`** (`src/tooluniverse/disgenet_tool.py`): the API emits one boilerplate "parameter unspecified" warning per unset optional filter (~30 for a plain gene query), burying the one actionable warning (an academic-key CURATED-sources-only note). Filtered out. Also, a legacy/alias gene symbol (e.g. `CARD15`, the old official symbol for `NOD2`) silently returned an empty association list with no indication a symbol mismatch, not an absence of data, was the cause; added an explanatory note (we can't safely auto-resolve the alias without a maintained synonym table, so we flag the ambiguity instead).
- **`gwas_get_variants_for_trait`** (`src/tooluniverse/gwas_tool.py`): confirmed live that "premature ovarian insufficiency" resolves to `HP_0008209` (an HPO term with 0 tagged associations) while the older synonym "premature ovarian failure" resolves to `MONDO_0005387` (9 real associations, e.g. PMID 21989058) — because GWAS Catalog's own trait-label search isn't restricted to EFO and does no synonym expansion. The empty-result note now explains this and suggests trying an older/alternate clinical synonym.
- **`drugbank_get_drug_interactions_by_drug_name_or_id`** and 5 sibling XMLTool-family tools (`src/tooluniverse/xml_tool.py`): a drug with many documented interactions (e.g. azathioprine, ~1286) produced an 80k+ token response regardless of the `limit` parameter, because `limit` only bounds top-level matched records, not a nested list field inside one record. Nested list fields are now capped for display (25 items) with the true count preserved in a `<field>_total_count` key; the untruncated list is still used to build searchable text so query matching is unaffected. Also moved a `print()` (dataset-load status) to the logger — it was writing straight to stdout ahead of any JSON payload, breaking naive parsing of piped/raw tool output.

### Reviewed but not changed (documented reasoning, not silently dropped)

- `find_tools` semantic search not surfacing DDI tools for a natural-language "check drug interaction" query — real discoverability gap, but fixing it well would mean touching the embedding/index pipeline, which is out of scope for a small, verifiable fix this round.
- FAERS/openFDA tool family using different drug-name parameter names (`medicinalproduct` vs `drug_name` vs `drug1`/`drug2`) — each wraps a genuinely different upstream endpoint shape; adding aliases across the family risks the "aliases that grow forever" anti-pattern without fixing the actual abstraction.
- `drugbank_get_drug_interactions_by_drug_name_or_id`'s parameter being named `query` when the tool name says "by_drug_name_or_id" — cosmetic (LOW severity), the actual error message when the wrong name is guessed already names the correct parameter.

## Test plan

- [x] `tu test` re-run on every changed tool (`ClinicalTrials_search_by_intervention`, `ClinicalTrials_search_by_sponsor`, `DisGeNET_search_gene`, `gwas_get_variants_for_trait`, and all 6 XMLTool-family drugbank tools sharing `xml_tool.py`) — all pass
- [x] Live CLI re-verification of each original repro case (vedolizumab/ustekinumab intervention search, Takeda sponsor search, CARD15/NOD2 gene lookup, premature ovarian insufficiency trait lookup, azathioprine interaction payload size + stdout cleanliness)
- [x] `pytest tests/unit/test_ctg_phrase_quoting.py tests/unit/test_ctg_tool.py tests/unit/test_ctg_phase_gate_with_status.py tests/unit/test_xml_tools_required_defaults.py tests/tools/test_disgenet_tool.py tests/unit/test_gwas_empty_result_note.py` and other gwas_* unit tests — all pass; two pre-existing tests (`test_condition_and_intervention_are_quoted_in_live_request`, `test_sponsor_alias_maps_to_query_spons`) asserted the old unrestricted-value behavior and were updated to match the new, verified-correct AREA-restricted behavior
- [x] `ruff check` / `ruff format` clean on all changed files
- [x] Full `pytest tests/` regression run — no new failures beyond a handful of pre-existing/environment-flaky ones (confirmed identical on unmodified `main`: an MCP-plugin doc-sync test, a subprocess-timeout test, and a `--refresh` flag assertion, plus two SDK-integration tests that time out at the default 60s under concurrent load but pass at 180s)

### DIFF
--- a/src/tooluniverse/ctg_tool.py
+++ b/src/tooluniverse/ctg_tool.py
@@ -24,6 +24,28 @@ def _phrase_quote_if_plain(value: str) -> str:
     return value
 
 
+def _restrict_to_area(value: str, area: str) -> str:
+    """Fix-R13B-1: CTG API v2's Essie search engine treats a bare
+    query.intr/query.spons value as a match against the whole study
+    record (brief summaries, arm descriptions, eligibility text, etc.)
+    rather than restricting to the actual intervention-name or
+    lead-sponsor-name field, so a specific query like 'vedolizumab' or
+    'Takeda' surfaces completely unrelated trials -- confirmed live: an
+    oncology trial with no vedolizumab intervention ranked in the
+    query.intr top 5 (215 unrestricted hits vs. 134 once restricted, all
+    correctly containing the drug), and a Neurocrine/Shire-sponsored
+    trial ranked in the query.spons top 5 for 'Takeda' (1977 unrestricted
+    hits vs. 1022 once restricted, all correctly Takeda-sponsored).
+    Wrapping the value in Essie's AREA[<area>] operator (and
+    phrase-quoting multi-word values, same as _phrase_quote_if_plain)
+    fixes this -- but only for a plain value with no existing Essie
+    syntax, since advanced users may already supply their own
+    AREA/boolean operators."""
+    if _ESSIE_OPERATOR_RE.search(value):
+        return value
+    return f"AREA[{area}]{_phrase_quote_if_plain(value)}"
+
+
 @register_tool("ClinicalTrialsTool")
 class ClinicalTrialsTool(RESTfulTool):
     _BASE_URL = "https://clinicaltrials.gov/api/v2"
@@ -160,6 +182,13 @@ class ClinicalTrialsTool(RESTfulTool):
             return {"status": "success", **result}
         return result
 
+    # Essie fields that get restricted via AREA[<name>] rather than a plain
+    # value match (see _restrict_to_area). Keyed by the mapped API param name.
+    _AREA_RESTRICTED_FIELDS = {
+        "query.intr": "InterventionName",
+        "query.spons": "LeadSponsorName",
+    }
+
     def _run_search(self, arguments):
         """Handle search operations (search_studies, search_by_intervention, search_by_sponsor)."""
         import requests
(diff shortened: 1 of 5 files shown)


gasvn:

PR #486, opened 2026-08-09
Fix Round 14: silent parameter mismatches across ChEMBL, OpenTargets, DisGeNET, GWAS, Monarch, ClinicalTrials.gov

gasvn:
## Summary

Role-play sessions this round covered rheumatology/autoimmune genetics, sleep medicine, and dental/oral health genetics research questions. Every reported issue was CLI-verified against live APIs before being queued.

Root cause shared by several findings: a tool whose JSON schema has no required properties (a common "all filters optional" search endpoint) lets `jsonschema.validate()` silently accept a misnamed parameter, since nothing forces a required-property check to fail. The caller gets a plausible-looking `status=success` response built from the tool's default/unfiltered query instead of an error naming the mistake.

### Verified and fixed

- **`BaseTool.validate_parameters()`** (`src/tooluniverse/base_tool.py`): now catches this class of bug for all ~2600 tools built on `BaseTool`. Two checks, added conservatively to avoid new false positives: (1) a near-miss of an unset schema property is flagged as a "did you mean?" hint regardless of what else was supplied; (2) if literally nothing supplied matches a real property, the tool's actual accepted parameter list is named. Confirmed live: `ChEMBL_search_drugs({"drug_name": "baricitinib"})` previously returned 20 unrelated drugs (its default page) because the only real parameter is `query`; `OpenTargets_get_evidence_by_datasource` silently ignored a singular `datasourceId` (real param is plural `datasourceIds`) even when mixed with otherwise-valid parameters. Also extended the existing "did you mean?" hint to `oneOf`-alternative schemas (e.g. `gwas_get_associations_for_trait`'s `disease_trait`/`efo_uri`/`efo_id`/`efo_trait`), which previously gave only a bare "not valid under any of the given schemas".
- **`restful_tool.py` (`MonarchTool`)**: `subject`/`object` parameters documented as CURIEs (e.g. `Monarch_get_gene_diseases`, `Monarch_get_gene_phenotypes`) now reject a bare gene/disease symbol instead of silently returning a `status=success, total=0` page indistinguishable from a genuine "no associations" result. Confirmed live: `Monarch_get_gene_diseases({"subject": "IRF6"})` returned `total=0`, but the correct CURIE `HGNC:6121` returns 2 real associations.
- **`disgenet_tool.py`**: `DisGeNET_search_disease`'s "not a UMLS CUI" error always suggested the identical placeholder CUI `C0152200` regardless of the input disease name, positioned right after echoing the user's own query — reading as a resolved answer rather than a generic example (and `C0152200` is Achromatopsia, unrelated to any of the diseases tested). Replaced with a consistent, explicitly-labeled example matching the tool's own description (`C0006142` for breast cancer), shared via one helper instead of three copies of the same string.
- **`opentarget_tools.json`** (`OpenTargets_get_drug_mechanisms_of_action_by_chemblId`): two evidence rows for the same drug/target/mechanism looked like exact duplicates because the query never requested the `references` field that actually distinguishes them (confirmed live for pitolisant: one row cites PubMed only, the other cites FDA + PubMed).
- **`ctg_tool.py`** (`search_clinical_trials`): the tool's hardcoded phase>=2 filter excludes NA-phase/observational studies — the bucket most genetics/epidemiology trials fall into — and a resulting `total_count=0` was indistinguishable from "no matching trials exist". Confirmed live: `condition="orofacial cleft"` returns 0 here despite real NA-phase matches (e.g. NCT03065686) existing on ClinicalTrials.gov. Added a note whenever the filter is the likely reason for a zero-result page.

### Reviewed but intentionally not changed this round

- `DisGeNET_search_gene`'s per-request "parameter unspecified" warning noise and the ClinicalTrials.gov intervention/sponsor `AREA[...]` restriction are already fixed on unmerged sibling PRs #485 and #483 (main hasn't absorbed them yet) — re-fixing here would only conflict on merge.
- `drugbank` `exact_match` grouping combination products under their parent drug, and Essie phrase-quoting stacking with the phase filter to collapse "non-syndromic cleft lip and palate" from 15 to 0 results, are both existing, deliberate tradeoffs from earlier rounds; changing them needs more validation than this round's time budget allows.
- Extending the `references`-field fix to two sibling OpenTargets tools (`get_known_drugs_by_drug_chemblId`, `get_associated_targets_by_drug_chemblId`) surfaced a larger, pre-existing `return_schema`/`query_schema` field mismatch in both — out of scope for this fix, flagged for a dedicated round.

## Test plan

- [x] Live CLI re-verification of every original repro case (ChEMBL_search_drugs, OpenTargets_get_evidence_by_datasource, gwas_get_associations_for_trait, Monarch_get_gene_diseases, DisGeNET_search_disease, search_clinical_trials, OpenTargets_get_drug_mechanisms_of_action_by_chemblId)
- [x] Targeted regression: 194 tests across `test_ux_error_hints_r3`, `test_parameter_validation`, additional-properties/optional-operation suites, DisGeNET, Monarch (CURIE normalization + result-id-prefix), ClinicalTrials.gov, and OpenTargets suites — all pass
- [x] `ruff check` / `ruff format --check` clean on all changed Python files
- [x] Independent tool-validator agent re-ran `tu info`/`tu run`/`tu test` smoke checks on all changed tools plus sibling Monarch tools whose params are *not* CURIE-gated (to confirm no regression) — pass

### DIFF
--- a/src/tooluniverse/base_tool.py
+++ b/src/tooluniverse/base_tool.py
@@ -233,6 +233,34 @@ class BaseTool:
         )
         return norm_to_key[match[0]] if match else None
 
+    @classmethod
+    def _best_property_match(
+        cls, supplied_key: str, unset_props: list
+    ) -> Optional[str]:
+        """The inverse of ``_find_misspelled_key``: given one supplied key
+        that isn't a schema property, find which *unset* property it was
+        probably meant to be. Used when there are few unknown keys but many
+        optional schema properties (a filter-heavy search endpoint), so the
+        fuzzy match runs once per supplied key instead of once per schema
+        property."""
+        if not unset_props:
+            return None
+        target = supplied_key.lower()
+        for prop in unset_props:
+            if prop.lower() == target:
+                return prop
+        target_norm = cls._normalize_key(supplied_key)
+        for prop in unset_props:
+            if cls._normalize_key(prop) == target_norm:
+                return prop
+        import difflib
+
+        norm_to_prop = {cls._normalize_key(p): p for p in unset_props}
+        match = difflib.get_close_matches(
+            target_norm, list(norm_to_prop), n=1, cutoff=0.8
+        )
+        return norm_to_prop[match[0]] if match else None
+
     def validate_parameters(self, arguments: Dict[str, Any]) -> Optional[ToolError]:
         """
         Validate parameters against tool schema.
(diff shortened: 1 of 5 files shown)


gasvn:

PR #487, opened 2026-08-09
Fix Round 15: Ensembl species-default, Reactome/KEGG/NCBI non-human silent failures, FAERS rate-limit masking, PubMed/translate-id docs

gasvn:
## Summary

Role-play testing across new personas this round (forensic genetics, pain-management/anesthesiology pharmacogenomics, agricultural/livestock breeding genomics) surfaced several places where a tool returns a plausible-looking wrong answer instead of a clear error -- worse than crashing outright because it looks like real data.

- **src/tooluniverse/ensembl_tool.py** (high severity): `EnsemblRESTTool` auto-defaulted `species=homo_sapiens` into query params even when routing a stable-ID lookup (`/lookup/id/{gene_id}`), which doesn't need a species filter since stable IDs are already species-specific. This silently broke every non-human Ensembl gene lookup chained from a symbol resolver (confirmed with bovine DGAT1, `ENSBTAG00000026356`) with a misleading "ID not found" error. Now only forwards `species` when the caller explicitly supplied it.
- **src/tooluniverse/reactome_tool.py**: Reactome's `/data/mapping/` endpoints return HTTP 404 `NOT_FOUND` as their genuine "nothing found for this identifier" signal, which the tool treated as a hard error instead of an empty result -- breaking pathway/reaction lookups for non-human protein orthologs with no directly curated Reactome entries.
- **src/tooluniverse/kegg_ext_tool.py**: `KEGG_get_gene_pathways` silently returned an empty pathway list (indistinguishable from a real negative) when `gene_id` was passed as `"org:SYMBOL"` (e.g. `"bta:DGAT1"`) rather than KEGG's numeric ID form. Now resolves the symbol the same way the no-colon path already did.
- **src/tooluniverse/ncbi_datasets_tool.py**: `NCBIDatasets_get_gene_by_symbol` returned a silent empty success for taxon tokens NCBI itself flags as ambiguous (e.g. `"cattle"` resolves above species level), with no way to tell that apart from a genuine zero-match. Now surfaces NCBI's own warning message.
- **src/tooluniverse/faers_analytics_tool.py**: `FAERS_calculate_disproportionality`'s internal count fetcher swallowed all exceptions -- including HTTP 429 rate limits -- and returned 0, so a rate-limited window produced `"Insufficient data: a=0, b=0, c=0, d=0"` for well-known drug/event pairs instead of surfacing the real upstream failure.
- **src/tooluniverse/data/fda_drug_adverse_event_tools.json**: documented that `FAERS_count_additive_adverse_reactions` combines multiple `medicinalproducts` with OR (union), not AND (co-occurrence) -- intentional aggregation behavior, but easily misread as combination-risk counting since a sibling tool uses AND semantics for the same-shaped input.
- **src/tooluniverse/data/pubmed_tools.json**: documented that PubMed's esearch ANDs every keyword by default, so a realistic multi-concept query silently returns 0 results even when relevant literature exists (confirmed live: an 11-word query returned 0 on PubMed but multiple relevant hits on EuropePMC for the same string).
- **plugin/commands/translate-id.md**: the skill's own worked example used the wrong tool name casing (`mygene_query_genes`) and parameter name (`q`), so following its instructions literally failed immediately. Fixed to `MyGene_query_genes` / `query`.

Checked all 9 other currently-open round PRs (#478-486) for overlap before fixing anything; skipped one duplicate (DrugBank interaction-list bloat, already covered by round 13's PR #485).

## Test plan

- [x] Every fix reproduced live against the real upstream API (Ensembl, Reactome, KEGG, NCBI Datasets, openFDA, PubMed/EuropePMC), several cross-checked via raw `curl`
- [x] Shipped `test_examples` still pass for every touched tool (`tu test <tool>`)
- [x] Targeted regression suite covering Ensembl/Reactome/KEGG/NCBI-Datasets/FAERS/PubMed: 103 passed, 0 failed
- [x] Confirmed no regression on the human/default-case path for each fix (e.g. human Ensembl stable ID, human Reactome mapping, `hsa:` KEGG IDs, `taxon="human"`, non-rate-limited FAERS calls)

### DIFF
--- a/plugin/commands/translate-id.md
+++ b/plugin/commands/translate-id.md
@@ -37,7 +37,7 @@ or pick gene by default and note the assumption.
 ### 2. Pick the resolver
 
 For genes/proteins:
-- HGNC symbol → all → `tu run mygene_query_genes '{"q":"<symbol>","species":"human"}'`
+- HGNC symbol → all → `tu run MyGene_query_genes '{"query":"<symbol>","species":"human"}'`
   (returns Ensembl, UniProt, RefSeq, NCBI Gene, MGI in one call)
 - Ensembl ID → all → `tu run ensembl_lookup_gene '{"gene_id":"<id>"}'`
 - UniProt → gene/Ensembl → `tu run UniProt_search '{"query":"<accession>","limit":1}'`
--- a/src/tooluniverse/data/fda_drug_adverse_event_tools.json
+++ b/src/tooluniverse/data/fda_drug_adverse_event_tools.json
@@ -1300,7 +1300,7 @@
   {
     "type": "FDACountAdditiveReactionsTool",
     "name": "FAERS_count_additive_adverse_reactions",
-    "description": "Aggregate adverse reaction counts across specified medicinal products. Only medicinalproducts is required; all other filters (patientsex, patientagegroup, occurcountry, serious, seriousnessdeath) are optional. Use filters sparingly to avoid overly restrictive searches that return no results. Data source: FDA Adverse Event Reporting System (FAERS).",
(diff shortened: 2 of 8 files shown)


gasvn:

PR #489, opened 2026-08-09
Fix Round 17: FAERS rate-limit handling, OpenTargets search overconfidence, GraphQL/XML tool clarity

gasvn:
## Summary

Persona role-play across three domains not previously covered (transplant medicine/HLA-matching, urology, addiction-medicine pharmacogenomics) surfaced five CLI-verified issues, each reproduced live before being fixed:

- **FAERS_calculate_disproportionality**: silently swallowed a failed openFDA count query (commonly HTTP 429 rate limiting) and treated it as a genuine zero, producing a mathematically impossible negative contingency-table cell (`a=0, b=283544, c=0, d=-283544` on the tool's own documented example). `_get_faers_count` now returns `None` (not `0`) on request failure, and every FAERS analytics operation is wired with `FDA_API_KEY` + `request_with_retry` to reduce how often the rate limit is hit at all.
- **OpenTargets_get_disease_ids_by_name**: hardcoded search page size 1, so a single full-text-search hit was trusted blindly even when OpenTargets' own relevance ranking got it wrong (confirmed live: "localized prostate cancer" ranks *Ewing sarcoma* above prostate cancer, which doesn't appear in the top 5 at all). Now returns up to 5 ranked candidates with their relevance score, and the description tells callers to verify the `name` field before trusting `dbXRefs`.
- **GraphQLTool (shared base class)**: a search-shaped query with a genuine zero-hit result collapsed to an opaque `{"search": {}}` once its empty `hits` list was stripped by `remove_none_and_empty_values` — indistinguishable from a broken response. Now attaches an explanatory `metadata.note`, inherited by every tool built on this class.
- **XMLDatasetTool (drugbank_*/mesh_* tools, ~19 tools)**: all share one search parameter literally named `query`, regardless of what the tool's own name promises (e.g. `drugbank_get_pharmacology_by_drug_name_or_drugbank_id`). A natural first guess like `drug_name` errored with `"'query' is a required property"`. Aliases are now resolved inside `validate_parameters()` (before schema validation runs, not just inside `run()`).
- **FAERS_search_reports_by_drug_and_reaction**: the `reactionmeddrapt` parameter description now explains it requires an exact MedDRA Preferred Term and points to `FAERS_count_reactions_by_drug_event` for term discovery, since a non-matching term returns an empty result indistinguishable from a genuine zero-report finding.

A sixth finding (`gwas_search_snps` silently ignoring an unrecognized `gene` filter param) was confirmed live but not fixed here — it's already covered by the not-yet-merged systemic parameter-validation fix in PR #486, which flags any single unrecognized parameter against a tool's declared schema.

## Test plan

- [x] All 5 findings reproduced live via `tu run`/`tu test` before fixing
- [x] New regression tests: `tests/unit/test_faers_disproportionality_rate_limit.py`, `tests/unit/test_xml_tool_query_alias.py`, plus a new case in `tests/tools/test_graphql_empty_guard.py`
- [x] Existing test suites covering the changed files pass (FAERS analytics, OpenTargets/GraphQL, XML tools) — 77+ tests, 0 failures
- [x] Each fix re-verified against the original failing live call after the change
- [x] `ruff check` clean on all changed files

### DIFF
--- a/src/tooluniverse/data/fda_drug_adverse_event_detail_tools.json
+++ b/src/tooluniverse/data/fda_drug_adverse_event_detail_tools.json
@@ -263,7 +263,7 @@
         },
         "reactionmeddrapt": {
           "type": "string",
-          "description": "MedDRA preferred term for the adverse reaction (required). Example: 'INFUSION RELATED REACTION', 'DYSPNOEA'."
+          "description": "MedDRA preferred term for the adverse reaction (required). Example: 'INFUSION RELATED REACTION', 'DYSPNOEA'. Must be the exact MedDRA Preferred Term FAERS uses, not a clinical description or synonym -- e.g. 'NEPHROTOXICITY' matches nothing, the real terms are 'ACUTE KIDNEY INJURY' or 'RENAL IMPAIRMENT'. A non-matching term returns an empty result (not an error), which looks identical to a genuine zero-report finding. If unsure of the exact term, call FAERS_count_reactions_by_drug_event first (with just medicinalproduct) to list the actual MedDRA terms reported for that drug."
         },
         "limit": {
           "type": "integer",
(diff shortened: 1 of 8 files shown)


gasvn:

PR #493, opened 2026-08-10
Fix Round 20: EFO abbreviation lookups return the wrong disease; openFDA fallback results look like exact matches

gasvn:
Round 20 of the test-harness. Two independent data-quality fixes, both reproduced live against the real APIs before and after the change.

## 1. EFO abbreviation lookups return the wrong disease (`efo_tool.py`)

OLS4 ranks search hits on the full indexed text — including a term's free-text description — not just the term's own name. For an abbreviation, the real disease term loses to whatever unrelated term happens to mention the abbreviation in prose.

Reproduced directly against the OLS4 API:

```
GET /ols4/api/search?ontology=efo&q=PCOS&rows=1
  -> EFO_0005187  "C-peptide measurement"     (mentions PCOS only in its description)

rows=10 shows the term that should have won:
  -> MONDO_0008487 "polycystic ovary syndrome" (exact_synonyms includes "PCOS")
```

`EFOTool` passes `rows` straight through, so the default `rows=1` never even fetched the correct term. A confidently wrong EFO ID is worse than a lookup failure here: callers treat the ID as authoritative and chain further lookups onto it.

The fix fetches enough candidates to see past a mis-ranked top hit, then prefers any candidate whose own `label` or `exact_synonyms` matches the query exactly, falling back to OLS's own ordering otherwise.

## 2. openFDA fallback results look identical to exact matches (`openfda_tool.py`)

`search_openfda` has three fallback stages — broadened term-based match in the same fields, full label-text match, and closest-spelling match — plus a label-section substitution. Each weakens the match in a different way, but every one of them returned a payload shaped exactly like an exact hit. `used_generic_fallback` was tracked internally and never surfaced.

The practical consequence: a query that names no product at all still returns rows, and the caller has nothing to distinguish "this is the drug you asked for" from "this is the nearest thing we could find".

Each stage now attaches a `note` describing what was actually matched.

## Verification

CLI, against live APIs:

| Call | Result |
| --- | --- |
| `OSL_get_efo_id_by_disease_name {"disease":"PCOS"}` | `MONDO_0008487` polycystic ovary syndrome (was `EFO_0005187` C-peptide measurement) |
| `OSL_get_efo_id_by_disease_name {"disease":"asthma"}` | `MONDO_0004979` asthma — unchanged |
| `OSL_get_efo_id_by_disease_name {"disease":"breast carcinoma","rows":3}` | 3 ordered hits — unchanged |
| `FDA_get_active_ingredient_info_by_drug_name {"drug_name":"aspirn"}` | section-substitution note: `active_ingredient` absent, served from `spl_product_data_elements` |
| `..."magnesium stearate tablet"` | label-text broadening note |
| `..."ibuprofin"` | closest-spelling note naming `IBUPROFEN` |
| `..."EDTA"` | no note — matches at the primary stage, correctly unannotated |

Also checked:

- `tests/unit/test_openfda_field_value_projection.py`, `test_confidently_wrong_answers_r6.py`, `test_gap_efo_descendants.py` — 55 passed.
- No `return_schema` on any FDA/OpenFDA tool sets `additionalProperties: false`, so the added `note` key does not break envelope validation.

## Notes

This branch was originally produced by a session that was interrupted before it could verify or ship the work; the commit was preserved unverified. Everything above is verification done after the fact, and one code comment was corrected — the original cited an "EDTA" example that does not actually reproduce (that query matches at the primary stage), replaced with a case that does.

### DIFF
--- a/src/tooluniverse/efo_tool.py
+++ b/src/tooluniverse/efo_tool.py
@@ -29,7 +29,20 @@ class EFOTool(BaseTool):
         return self._search(disease, rows)
 
     def _search(self, disease, rows):
-        params = {"ontology": "efo", "q": disease, "rows": rows}
+        # Fix-20B-2: OLS4's default relevance ranking scores hits on the full
+        # indexed text (including free-text descriptions), not just the
+        # term's own name. For an abbreviation like "PCOS" this put a term
+        # whose *description* merely mentions "PCOS" in passing (EFO_0005187
+        # "C-peptide measurement") ahead of the actual term for the disease
+        # (MONDO_0008487 "polycystic ovary syndrome", which lists "PCOS" as
+        # an exact_synonym) -- confirmed live against the OLS4 API. Silently
+        # returning the wrong disease's EFO ID is worse than a fetch failure,
+        # since callers treat it as authoritative and chain further lookups
+        # on it. Always fetch enough candidates to see past a mis-ranked top
+        # hit, then prefer any candidate whose own label/exact_synonym is a
+        # case-insensitive exact match to the query over the raw OLS order.
+        fetch_rows = max(rows, 10)
+        params = {"ontology": "efo", "q": disease, "rows": fetch_rows}
         try:
             response = requests.get(self.base_url, params=params, timeout=20)
             response.raise_for_status()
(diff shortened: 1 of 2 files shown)


gasvn:

PR #494, opened 2026-08-10
Fix Round 21: queries that silently return nothing instead of answering

gasvn:
Round 21 of the role-play test harness. Three researcher personas (clinical immunologist / autoimmune genetics, structural biologist / drug discovery, clinical microbiologist / infectious disease) worked real multi-tool research questions end to end. They reported 15 issues; 5 were discarded as false positives after CLI verification, and the 10 below reproduced.

They share one failure mode: a lookup that had a good answer available returned an **empty-but-successful** result, or an error whose text gave the caller nothing to act on. Zero rows and "this does not exist" are indistinguishable to the caller, so each of these quietly produced a wrong scientific conclusion rather than a visible failure.

### 1. `UniProt_search` rejected every non-model organism (HIGH)
`organism` mapped 22 model organisms to taxids and passed anything else straight into `organism_id:`, which accepts numeric IDs only — so pathogens, plants and non-model species failed the **whole search** with HTTP 400. Non-numeric organisms now route to `organism_name:`, which accepts free text; the same rule applies to an `organism:` clause written inside `query`.

```
python -m tooluniverse.cli run UniProt_search '{"query":"gene:KPC-2","organism":"Klebsiella pneumoniae","limit":5}'
```
Before: `400 Client Error ... organism_id%3AKlebsiella+pneumoniae`. After: Q9F663 `BLKPC_KLEPN`, carbapenem-hydrolyzing beta-lactamase KPC-2. `organism:"human"` is unchanged (1887 results, same as before).

### 2. FAERS analytics found zero reports for any brand name
All six operations searched `openfda.generic_name` only, so a brand name looked like a drug with no reports and disproportionality refused to compute.

```
python -m tooluniverse.cli run FAERS_calculate_disproportionality '{"drug_name":"XELJANZ","adverse_event":"Pulmonary embolism"}'
```
Before: `Insufficient data: a=0, b=0, c=94648, d=20598042`. After: ROR 1.254 (95% CI 1.179–1.332), consistent with generic tofacitinib's 1.306 on the same event. Generic-name queries return identical numbers to before.

### 3. `OpenTargets_get_evidence_by_datasource` advertised a retired datasource ID
`ot_genetics_portal` was renamed upstream to `gwas_credible_sets`; the retired name matches nothing rather than erroring, so genetic-association evidence looked absent. The tool's own description still listed the retired name first. Both known retired IDs (`ot_genetics_portal`, `chembl`) are now remapped, the substitution is reported in `metadata.datasource_rename_note`, and the description is corrected.

```
python -m tooluniverse.cli run OpenTargets_get_evidence_by_datasource '{"gene_symbol":"IL23R","disease_name":"Crohn disease","datasourceIds":["ot_genetics_portal"],"size":2}'
```
Before: `evidences.count: 0`. After: 43 rows. Independently confirmed on TP53/cancer (0 → 51).

### 4. `GWAS_search_associations_by_gene` missed hyphenated gene spellings
`mapped_gene` matches the catalog symbol literally, so `IL-23R` returned nothing while 112 associations sit under `IL23R`. The unhyphenated retry fires **only after a miss**, so genuinely hyphenated symbols are untouched, and the response names the spelling that produced the results.

```
python -m tooluniverse.cli run GWAS_search_associations_by_gene '{"gene_name":"IL-23R","size":2}'
```
Before: `association_count: 0`. After: `total_found: 112` plus `gene_name_note` explaining the substitution. `HLA-A` still resolves on the first attempt (107 associations) and is not rewritten.

### 5. `BVBRC_get_taxonomy` turned a species name into a bare 404
`/taxonomy/{id}` is a taxid path lookup, so a name came back as `BV-BRC API HTTP error: 404` with no indication of why. The id is now validated as numeric up front, pointing at `BVBRC_search_taxonomy`, which resolves names.

```
python -m tooluniverse.cli run BVBRC_get_taxonomy '{"taxon_id":"Klebsiella pneumoniae"}'
```
Numeric ids (`573`) are unaffected.

### 6. `EBITaxonomy_get_by_scientific_name` returned empty success for abbreviated genera
ENA indexes full binomials, so the form clinicians write (`K. pneumoniae`) returned `status: success, data: []` — identical to a species genuinely absent from the database. The genus letter cannot be expanded unambiguously (`K.` is Klebsiella, Klebsormidium, Kluyveromyces…), so it is rejected with the expansion rule rather than guessed.

```
python -m tooluniverse.cli run EBITaxonomy_get_by_scientific_name '{"scientific_name":"K. pneumoniae"}'
```
Full binomials are unaffected (taxid 573).

### 7. `BindingDB_get_ligands_by_pdb` dumped a Tomcat HTML error page
The upstream PDB endpoint answers HTTP 500 for **every** structure — verified on 6OIM and 2RH1, with and without the cutoff parameter, while `getLigandsByUniprots` answers 200 and the singular `getLigandsByPDB` is a 404. The tool echoed the first 200 characters of the HTML error page. HTML bodies are now summarised, and a 5xx from this route names the PDB→UniProt path that does work.

```
python -m tooluniverse.cli run BindingDB_get_ligands_by_pdb '{"pdb_ids":"6OIM"}'
```
Before: `<!doctype html><html lang="en"><head><title>HTTP Status 500 …<style type="text/css">body {font-family:Tahoma…`. After: a one-line error naming `PDBeSIFTS_get_pdb_to_uniprot` + `BindingDB_get_ligands_by_uniprot`.

### 8–9. Chained calls rejected the identifier name their own sibling emits
`ChEMBL_get_molecule_targets` refused `chembl_id` (the name `ChEMBL_get_molecule` uses), and the three `alphafold_*` tools refused `accession` (the name the `UniProt_*_by_accession` tools use) — so feeding one call's output into the next failed validation on an id that had just been looked up. The alphafold *code* already accepted `accession`; only its schema did not, so validation rejected it before the handler ran.

```
python -m tooluniverse.cli run ChEMBL_get_molecule_targets '{"chembl_id":"CHEMBL4535757"}'
python -m tooluniverse.cli run alphafold_get_summary '{"accession":"P01116"}'
```
Both previously `Unrecognized parameter(s)`; existing parameter names still work.

### 10. `PubChem_get_compound_properties_by_CID` promised a property name PubChem no longer returns
PUG-REST renamed `CanonicalSMILES` → `ConnectivitySMILES` (and `IsomericSMILES` → `SMILES`). A request for the legacy name returns the value under the *current* key, so a caller reading back the key it asked for found nothing. Description-only change documenting the current names and both renames.

## Discarded as false positives (CLI-verified clean)

- **`ChEMBL_search_targets` "tyrosine kinase 2" surfaces PTK2B, not TYK2** — `pref_name__contains` is a substring filter and "Protein-tyrosine kinase 2-beta" genuinely contains that substring; TYK2's ChEMBL name is "Non-receptor tyrosine-protein kinase TYK2". `pref_name__contains: "TYK2"` returns CHEMBL3553 correctly. The tool honoured its documented semantics.
- **`get_mutation_annotations_by_pdb_id` omits G12C for 6OIM and reports 4 mutations while listing 3** — reproduced, but this is RCSB's own deposited data: `data.rcsb.org/rest/v1/core/polymer_entity/6OIM/1` returns exactly `rcsb_mutation_count: 4` with `pdbx_mutation: "C51S,C80L,C118S"`. The tool relays upstream faithfully.
- **OpenTargets `efoId` field returns MONDO identifiers** — upstream field naming; MONDO ids are accepted directly by the downstream `efoId` tools, as the persona confirmed.
- **Misspelling "Chrohns disease" returns no hits** — genuine zero-hit search, and the existing metadata note already says so clearly.
- **`ClinicalTrials_search_by_intervention` exposes both `status` and `filter_status`** — documented intentional alias.

## Verification

- Full `tests/unit` suite green before and after (exit 0, failures 0; only environment skips for absent rdkit/torch/bcftools/Rscript).
- `tu test` passes for every touched tool, including `EBITaxonomy_get_by_scientific_name`, `alphafold_get_summary` and `GWAS_search_associations_by_gene`.
- Each fix re-run through the CLI with the persona's original arguments, plus an unchanged-behaviour control alongside it (shown above).
- `test_graphql_tool_round26_fixes.py`, which asserts no `metadata.note` is added for non-IntOGen datasource queries, still passes — the new disclosure uses a distinct `datasource_rename_note` key.
- Generated wrappers for the four tools whose schemas changed were hand-edited to match the committed style; the full regenerator was deliberately not run, as its output currently reformats all 2600 wrappers.
- No overlap with open PR #493 (`efo_tool.py`, `openfda_tool.py`) — this branch touches neither. FAERS analytics lives in `faers_analytics_tool.py`, a different module. Diffs of every other open PR were checked for file overlap.

### DIFF
--- a/src/tooluniverse/bindingdb_tool.py
+++ b/src/tooluniverse/bindingdb_tool.py
@@ -12,6 +12,7 @@ single id or a semicolon-delimited list — so we route every operation
 through the plural endpoint.
 """
 
+import re
 from typing import Any, Dict, List
 
 import requests
@@ -22,6 +23,21 @@ from .tool_registry import register_tool
 
 BASE_URL = "https://www.bindingdb.org/rest"
 DEFAULT_TIMEOUT = 30
+
+
+def _error_detail(resp: "requests.Response") -> str:
+    """Readable one-line summary of an error response body.
+
+    BindingDB serves Tomcat's HTML error page on failures, and echoing its
+    first 200 characters put a doctype and a stylesheet fragment in front of
+    the caller instead of anything diagnostic.
+    """
+    body = (resp.text or "").strip()
+    if "<html" in body[:200].lower() or body[:20].lower().startswith("<!doctype"):
+        title = re.search(r"<title>(.*?)</title>", body, re.IGNORECASE | re.DOTALL)
+        summary = title.group(1).strip() if title else "HTML error page"
+        return f"upstream returned an HTML error page ({summary})"
+    return body[:200] or "empty response body"
 
 
 def _http_get(
@@ -45,7 +61,7 @@ def _http_get(
     except requests.exceptions.ConnectionError as e:
         return {"_err": f"BindingDB connection failed: {e}"}
     if resp.status_code != 200:
-        return {"_err": f"BindingDB HTTP {resp.status_code}: {resp.text[:200]}"}
+        return {"_err": f"BindingDB HTTP {resp.status_code}: {_error_detail(resp)}"}
     try:
         return resp.json()
     except ValueError as e:
@@ -154,7 +170,24 @@ class BindingDBTool(BaseTool):
             timeout=self.timeout,
         )
         if "_err" in result:
-            return {"status": "error", "error": result["_err"]}
+            error = result["_err"]
+            # getLigandsByPDBs is currently answering 500 for every input, valid
+            # ids included (confirmed for 6OIM and 2RH1, with and without the
+            # cutoff parameter; the singular getLigandsByPDB is a 404 while
+            # getLigandsByUniprots answers 200). A bare relay of the upstream
+            # status leaves the caller unable to tell "this structure has no
+            # binding data" from "this whole lookup route is down", so name the
+            # route that does work.
+            if "HTTP 5" in error:
+                error += (
+                    ". BindingDB's PDB lookup endpoint is failing for all "
+                    "structures, not just this one. Map the PDB entry to a "
+                    "UniProt accession with PDBeSIFTS_get_pdb_to_uniprot and "
+                    "query BindingDB_get_ligands_by_uniprot instead, or use "
+                    "get_binding_affinity_by_pdb_id for RCSB's own curated "
+                    "affinity data."
+                )
+            return {"status": "error", "error": error}
         return {
             "status": "success",
             "data": {"pdbs": ids, "cutoff": cutoff, "affinities": _affinities(result)},
--- a/src/tooluniverse/bvbrc_tool.py
+++ b/src/tooluniverse/bvbrc_tool.py
@@ -584,6 +584,21 @@ class BVBRCTool(BaseTool):
         if not taxon_id:
             return {"status": "error", "error": "taxon_id parameter is required"}
 
+        # /taxonomy/{id} is an NCBI-taxid lookup, so a species name reaches
+        # BV-BRC as a nonexistent path and comes back as a bare "HTTP error:
+        # 404" that says nothing about why. Reject the name here and name the
+        # tool that turns it into an id.
+        if not str(taxon_id).strip().isdigit():
+            return {
+                "status": "error",
+                "error": (
+                    f"taxon_id must be a numeric NCBI taxonomy ID (e.g. 573 for "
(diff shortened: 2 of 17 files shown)


gasvn:

PR #495, opened 2026-08-10
Fix Round 22: lookups that report "no data" when the data is there

gasvn:
Round 22 of the role-play test harness. Three researcher personas (computational neuroscience / tauopathies, environmental toxicology / PFAS, clinical biochemical genetics / metabolomics) worked real research questions end-to-end; every reported issue was re-run through `tu run` before being queued, and most were discarded.

The theme of what survived verification: **a tool answers a well-formed question with a successful-looking empty or wrong result.** That reads to the caller as "this evidence does not exist" rather than "you asked the wrong way" — it ends the investigation instead of correcting it.

### 1. `HPA_get_rna_expression_in_specific_tissues` reported "No data" for tissues HPA does publish

Read HPA's tissue-*enrichment* summary (`RNA tissue specific nTPM`), which contains only the few tissues a gene is classified as enriched in — not the ~50-tissue panel. Every other tissue came back as `expression_level: "No data"`.

```
tu run HPA_get_rna_expression_in_specific_tissues '{"ensembl_id":"ENSG00000186868","tissue_names":["cerebral cortex","cerebellum","liver"]}'
```
- Before: `cerebral cortex -> matched_tissue "Not found", expression_level "No data"`, while HPA holds 147.9 nTPM for it (the sibling `HPA_get_rna_expression_by_source` returns it). `available_tissues_sample` listed only `["brain","skeletal muscle"]`.
- After: `cerebral cortex 147.9 (Very high)`, `cerebellum 139.1`, `liver 0.6`, each with the `source_field` it came from.

Same change also fixes a hard failure: `{"ensembl_id":"ENSG00000135823","tissue_names":["cerebral cortex","basal ganglia","midbrain","cerebellum"]}` (STX6, a PSP GWAS locus) previously errored with `"No RNA tissue expression data available for this gene"`; it now returns all four brain regions.

Now queries the per-tissue `t_RNA_<tissue>` columns — the same correction already applied to the sibling `HPAGetRnaExpressionBySourceTool` in Fix-R4A-2. An unmatched tissue now says the *name* was unrecognized rather than implying the gene is not expressed.

### 2. `disease_target_score` returned zero targets for a retired Open Targets datasource ID

Open Targets renamed `ot_genetics_portal` to `gwas_credible_sets` without keeping an alias, so the stale ID matched nothing and produced a successful empty result.

```
tu run disease_target_score '{"efoId":"MONDO_0019037","datasourceId":"ot_genetics_portal","pageSize":100}'
```
- Before: `total_targets_with_scores: 0`, empty `target_scores` — reading as "PSP has no genetic evidence".
- After: 23 targets, `datasource: "gwas_credible_sets"`, strongest first (MOBP 0.69, MAPT 0.62, STX6 0.57), plus a `datasource_rename_note` disclosing the substitution.

Three related changes to the same tool (and its nine datasource variants):
- The retired `chembl` -> `clinical_precedence` rename is handled the same way (already documented as confirmed-live in `opentarget_tools.json`).
- **Results are now sorted by score, strongest first.** The tool paginates under a 25s wall-clock budget, so a truncated run previously returned an arbitrary unordered subset — `tu run expression_atlas_disease_target_score '{"efoId":"MONDO_0019037","pageSize":15}'` returned 124 rows in API order with the top scorer buried; it now returns them sorted, so "the strongest targets" is actually obtainable.
- When a result is genuinely empty, the response now lists the datasource IDs the disease *does* have. Those IDs were already in the rows just scanned, so it costs nothing. `tu run disease_target_score '{"efoId":"MONDO_0019037","datasourceId":"not_a_real_source","pageSize":100}'` now returns `available_datasources: [clinical_precedence, europepmc, eva, expression_atlas, genomics_england, gwas_credible_sets, impc, uniprot_literature, uniprot_variants]` instead of a bare empty success.

### 3. RefMet name lookups failed on standard clinical metabolite names

`refmet/name/` is exact-match only, so names used in every clinical and HMDB context returned empty even when RefMet knows the compound under its own shorthand.

```
tu run MetabolomicsWorkbench_search_compound_by_name '{"input_value":"Octanoylcarnitine"}'
```
- Before: `{"status":"success","data":[],"message":"No results found. RefMet requires exact metabolite names..."}` — for the primary biomarker of MCAD deficiency.
- After: the full `CAR 8:0` record (CID 11953814, `RM0154093`, exact mass 287.209659) plus a `name_resolution_note` disclosing the resolution.

Falls back to RefMet's own `refmet/match/` synonym endpoint. Class-level terms still return the existing guidance message (`{"input_value":"ceramide"}` is unchanged), and the resolution is reported rather than applied silently.

### 4. `metabolights_search_studies` ignored its own query

The MetaboLights API accepts a `query` parameter and silently discards it, so every search returned the same first N accessions by ID.

```
tu run metabolights_search_studies '{"query":"acylcarnitine","size":5}'
```
- Before: `["MTBLS1","MTBLS2","MTBLS3",...]` — the numerically-first studies, identical for every query, with `status: "success"`.
- After: `["MTBLS12655","MTBLS14407","MTBLS13382","MTBLS13730","MTBLS10509"]`, `total_hits: 113`.

Routed through EBI Search, which indexes the same studies and honours the query. A no-hit search now says so and surfaces EBI's spelling suggestion. The tool's description previously documented the broken behaviour instead of fixing it; the now-dead pass-through paths in `_build_url`/`_build_params` and the client-side pagination branch were removed.

### 5. `AOPWiki_list_aops` had no way to search

The tool took no parameters and returned all ~600 AOPs, so finding the pathways for a chemical or endpoint meant pulling the whole list and grepping client-side.

```
tu run AOPWiki_list_aops '{"search":"thyroid"}'
```
- Before: no `search` parameter existed; the call returned every AOP.
- After: 20 thyroid AOPs including the two that matter for PFAS — 152 (transthyretin interference) and 8 (hepatic nuclear receptor induced TH catabolism).

Added to both `aopwiki_tool.py` and `data/aopwiki_tools.json` — the loader reads the JSON, so updating only the decorator config would have left the parameter undiscoverable in `tu info` even though it worked at runtime.

### 6. `openalex_literature_search` wrote non-JSON to stdout

A bare `print()` on the success path emitted `[OpenAlex] Retrieved N papers for keywords: '...'` ahead of the JSON, breaking `json.loads` for any programmatic consumer. No sibling literature tool does this. Routed to the module logger, along with two other stray prints in the same file.

## Test changes

`tests/unit/test_disease_target_score_pagination.py` used `chembl` as its fixture datasource — an ID retired upstream in favour of `clinical_precedence`, which the new remap correctly rewrites, so the fixture matched nothing. Moved to the current ID (the rename is already documented as confirmed-live in `opentarget_tools.json`) and added regression tests for the remap, the sort order, and the available-datasources hint.

## Verification

- `tu test` passes on every changed tool: `HPA_get_rna_expression_in_specific_tissues`, `disease_target_score`, `expression_atlas_disease_target_score`, `MetabolomicsWorkbench_search_compound_by_name`, `MetabolomicsWorkbench_get_refmet_info`, `metabolights_search_studies`, `metabolights_list_studies`, `AOPWiki_list_aops`, `AOPWiki_get_aop`, `openalex_literature_search` (15/15 examples).
- 213 targeted unit + integration tests covering every changed module pass.
- `ruff check` under the project config: clean. `ruff format --check`: no changed line is misformatted.

### DIFF
--- a/src/tooluniverse/aopwiki_tool.py
+++ b/src/tooluniverse/aopwiki_tool.py
@@ -42,11 +42,23 @@ def _aopwiki_not_found(result: Any) -> bool:
             "List Adverse Outcome Pathways (AOPs) from AOPWiki. Returns AOP IDs, "
             "titles, and short names. AOPs describe causal chains from molecular "
             "initiating events through key events to adverse outcomes, used in "
-            "toxicology and risk assessment."
+            "toxicology and risk assessment. Pass `search` to keep only AOPs whose "
+            "title or short name contains a keyword (e.g. 'thyroid', 'PPAR', "
+            "'steatosis') -- AOPWiki publishes ~600 AOPs, so an unfiltered list is "
+            "rarely what you want. Use AOPWiki_get_aop on a returned ID for its key "
+            "events, stressor chemicals, and causal relationships."
         ),
         "parameter": {
             "type": "object",
-            "properties": {},
+            "properties": {
+                "search": {
+                    "type": "string",
+                    "description": (
+                        "Case-insensitive keyword to filter AOPs by title or short "
+                        "name. Omit to return every AOP."
+                    ),
+                },
+            },
         },
         "settings": {"base_url": _BASE, "timeout": 30},
     },
@@ -85,7 +97,33 @@ class AOPWikiListTool:
         except Exception as e:
             return {"status": "error", "error": f"AOPWiki API error: {e}"}
 
-        return {"status": "success", "data": aops}
+        # AOPWiki's /aops.json has no server-side keyword filter, so filter the
+        # fetched list here rather than making callers pull ~600 AOPs and grep.
+        search = str(arguments.get("search") or "").strip().lower()
+        if not search:
+            return {"status": "success", "data": aops}
+
+        matches = [
+            a
+            for a in aops
+            if search in str(a.get("title") or "").lower()
+            or search in str(a.get("short_name") or "").lower()
+        ]
+        result = {
+            "status": "success",
+            "data": matches,
+            "count": len(matches),
+            "total_aops": len(aops),
+            "search": arguments.get("search"),
+        }
+        if not matches:
+            result["note"] = (
+                f"No AOP title or short name contains '{arguments.get('search')}'. "
+                f"Searched all {len(aops)} AOPs. Try a broader keyword (AOPWiki "
+                "titles use terms like 'thyroid', 'PPAR', 'steatosis', "
+                "'neurotoxicity'), or omit `search` to list every AOP."
+            )
+        return result
 
 
 @register_tool(
--- a/src/tooluniverse/data/aopwiki_tools.json
+++ b/src/tooluniverse/data/aopwiki_tools.json
@@ -2,10 +2,15 @@
   {
     "name": "AOPWiki_list_aops",
     "type": "AOPWikiListTool",
-    "description": "List Adverse Outcome Pathways (AOPs) from AOPWiki. Returns AOP IDs, titles, and short names. AOPs describe causal chains from molecular initiating events through key events to adverse outcomes, used in toxicology and risk assessment.",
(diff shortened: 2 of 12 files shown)


gasvn:

PR #496, opened 2026-08-10
Fix Round 23: results that look complete but silently are not

gasvn:
Round 23 of the role-play test harness. Three personas — cardiovascular pharmacoepidemiology, cancer immunotherapy clinical trials, and protein-interaction/pathway systems biology — reported 26 issues; 7 survived CLI verification and are fixed here.

All three personas independently hit the same shape of failure, and it is a different one from the last two rounds. Rounds 21 and 22 were about lookups that return *nothing* when the data is there. This round is about calls that return **something plausible while silently dropping part of the answer**. An empty result at least prompts the caller to look closer. A short-but-populated one does not — it gets used as if it were the whole answer.

## ClinicalTrials.gov (`ctg_tool.py`)

**Brand-name intervention searches nearly vanished.** Queries were restricted to `AREA[InterventionName]`, but registrants record brand names in the separate `InterventionOtherName` synonym field. Live counts:

| query | InterventionName only | both name fields | unrestricted |
|---|---|---|---|
| `Eliquis` + atrial fibrillation | 5 | **43** | 52 |
| `Opdualag` + melanoma | 2 | **19** | 22 |
| `vedolizumab` | 134 | **142** | 215 |
| `apixaban` + atrial fibrillation | 96 | **108** | 189 |

Searching both name fields recovers the brand-name hits without reopening the imprecision that motivated the original restriction — that noise came from description/arm-label text, which neither name field contains.

Before: `ClinicalTrials_search_by_intervention {"intervention":"Eliquis","condition":"atrial fibrillation","filter_phase":"PHASE3"}` → `total_count: 0`.
After: `total_count: 10`, including ARISTOTLE (NCT00412984).

**A failed per-study fetch was dropped without a trace.** `execute_RESTful_query` returns a falsy value for HTTP errors, timeouts and undecodable JSON alike, and the details loop discarded those silently. A transient blip surfaced either as a quietly short result set or as `"No relevant information found for the given NCT IDs."` — which reads as "this trial has no eligibility criteria" for a trial that demonstrably has them. Fetch failures and genuinely absent data no longer share a response; partial failures are now named in `metadata.failed_nct_ids`.

**Intervention lists were sliced to 5 with no count and no flag.** NCT03337698 is a real tiragolumab trial with 18 interventions, none of which survived the slice — it read as a false positive. Now reports `intervention_count` and `interventions_truncated`.

**`filter_status` / `filter_phase` were unvalidated.** The natural spellings `"recruiting"` and `"Phase 3"` produced a bare upstream HTTP 400 whose remediation advice — *retry, check service status, report if persistent* — is actively wrong for a typo. Both are now normalized case- and separator-insensitively, and a genuinely unknown value is rejected up front with the legal values named.

**The default phase-2-and-above gate was disclosed only on zero results**, so a caller with a healthy-looking list never learned that phase-1 and observational studies had been filtered out of it. A non-empty result is just as filtered as an empty one.

## Reactome interactors (`reactome_interactors_tool.py`)

`page_size` was clamped to 100, and `total_interactors` was read from the `/details` response's `count` field — which reports how many records *that page* returned, not the size of the full set. So `page_size=300` came back reporting `total_interactors: 100` with nothing indicating truncation, and the "total" changed with page size, breaking any pagination loop. Live: `pageSize=100` → `count=100`; `pageSize=300` → `count=249`.

The clamp is removed (Reactome honours larger page sizes) and the true total now comes from the dedicated `/summary` endpoint, with an explicit `truncated` flag and a note naming the `page_size` that would fetch everything. If `/summary` is unreachable, the page count is no longer passed off as a total (`total_is_exact: false`).

Before: `{"accession":"P04637","page_size":300}` → `total_interactors: 100`.
After: `total_interactors: 249, returned_interactors: 249, truncated: false` — matching the 249 the tool's own description advertises.

## STRING identifier mapping (`string_tool.py`)

Identifiers STRING could not map were simply absent from the response — no placeholder, no warning — so a batch of 500 gene symbols could return 480 records as a plain `success`, discoverable only by diffing `queryIndex` values by hand. Silently shrinking a gene set skews every downstream enrichment, and validating that identifiers exist in STRING is this tool's stated purpose. The submitted list is now reconciled against what came back, omissions are named in `metadata`, and mapping *nothing* is an error rather than an empty success.

Worth flagging, because the obvious fix here is wrong: absence does not prove failure. STRING also de-duplicates by resolved protein, keeping only the **last** identifier of each group — `["TP53","P04637"]` legitimately returns a single record at `queryIndex 1`. A naive check would have reported "TP53 could not be mapped", a confidently wrong claim about a real gene and worse than the original silent drop. Absent identifiers are therefore re-queried on their own (bounded, and only when something is actually missing) to separate collapsed duplicates from genuine misses.

## Verification

- `tests/unit` (the CI-blocking suite) green.
- Every fix re-run through the CLI with the persona's original arguments; before/after captured above.
- `tu test` passes for `ClinicalTrials_search_studies`, `ClinicalTrials_search_by_intervention`, `search_clinical_trials`, `get_clinical_trial_eligibility_criteria`, `ReactomeInteractors_get_protein_interactors`, `STRING_map_identifiers`.
- Two pre-existing test files asserted the old behaviour and were updated deliberately: `test_ctg_phrase_quoting.py` (the `AREA[]` construction) and `test_reactome_interactors_pagination.py` (which mocked a single HTTP call and is now split per endpoint, with added coverage for the true-total and degraded-`/summary` paths).
- Regression-checked that the STRING change does not touch the other tools sharing `STRINGRESTTool`.

19 of 26 reported issues were discarded as false positives or out of scope (73%) — including several where the tool was behaving correctly and the persona misread an honest empty result, and one where Reactome's static IntAct snapshot genuinely lacks the expected interactors (an upstream data-coverage matter, not a code defect).

### DIFF
--- a/src/tooluniverse/ctg_tool.py
+++ b/src/tooluniverse/ctg_tool.py
@@ -24,7 +24,7 @@ def _phrase_quote_if_plain(value: str) -> str:
     return value
 
 
-def _restrict_to_area(value: str, area: str) -> str:
+def _restrict_to_area(value: str, areas: tuple) -> str:
     """Fix-R13B-1: CTG API v2's Essie search engine treats a bare
     query.intr/query.spons value as a match against the whole study
     record (brief summaries, arm descriptions, eligibility text, etc.)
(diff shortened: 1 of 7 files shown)


gasvn:

PR #497, opened 2026-08-10
Fix Round 24: results that name the wrong entity while looking correct

gasvn:
Round 24 of the test harness. Three researcher personas (plant/agricultural genomics, nutrition & dietary exposure, rare-disease diagnostics) reported 30 issues; each was re-run through the CLI against a clean `origin/main` checkout before being queued.

The theme this round is different from #493-#496: not "the lookup came back empty", but **the lookup came back full, well-formed, and about a different entity**. Nothing in these responses gave a caller a reason to doubt them.

### `Orphanet_get_genes` answered MPS I with the MPS X gene
`{"orpha_code":"579"}` (Mucopolysaccharidosis type 1) returned `ARSK` at 5q15 — the MPS type X gene — with `subtype_sources: [{orpha_code: 662216, name: "Mucopolysaccharidosis type 10"}]`. Orphanet curates gene associations on subtype codes, so the tool falls back to scanning Orphadata's orphacode list for terms containing the parent's name, using a bare substring test. Orphanet's terms are numbered, and `"Mucopolysaccharidosis type 1"` is a substring of `"Mucopolysaccharidosis type 10"`.

The match now has to sit on a token boundary. Measured over all 4,245 gene-associated orphacodes, this removes matches for 43 parent names, and every one is the same false positive: `"Familial hyperaldosteronism type I"` → `type II/III/IV`, `"Mucolipidosis type II"` → `type III`, `"Achondroplasia"` → `"Pseudoachondroplasia"`, `"Neuroblastoma"` → `"Ganglioneuroblastoma"`. Real subtypes are all retained — `"Mucopolysaccharidosis type 4A"`, `"Mucopolysaccharidosis type 2, severe form"`, `"Neonatal Marfan syndrome"`, all 16 Ehlers-Danlos subtypes, all 71 Charcot-Marie-Tooth subtypes.

- Before: `ORPHA:579 → ["ARSK"]`
- After: `ORPHA:579 → []`; 580 → `IDS`, 582 → `GALNS, GLB1`, 583 → `ARSB`, 558 → `TGFBR1, TGFBR2, FBN1` all unchanged

### `gather_disease_profile` merged three diseases into one identifier block
`{"disease":"mucopolysaccharidosis type I"}` returned `orphanet: 583` (MPS VI), `mesh: D009085` (MPS IV) and `ordo: ORDO:217085` (MPS II severe form) together under one disease name, with `per_source.orphanet.name` literally reading `"Mucopolysaccharidosis type 6"`. Every source is a ranked search and the code took each one's first hit; Orphanet ranks MPS VI first for that query and the real MPS I fourth.

Cross-references are now only promoted to `identifiers` when their own label names the queried disease, folding Roman and Arabic subtype numerals together so `"type I"` still matches Orphanet's `"type 1"`. Non-matching hits stay visible in `per_source` with a `match` flag rather than being presented as the disease's IDs.

- Before: `{"orphanet":"583","efo":"MONDO_0001586","ncit":"NCIT:C85053","mesh":"mesh:D009085","ordo":"ORDO:217085"}`
- After: `{"orphanet":"579","efo":"MONDO_0001586","ncit":"NCIT:C85053"}`
- `Hurler syndrome`, `cystic fibrosis` and `Marfan syndrome` keep every identifier they had (and gain a few) — no recall loss

### `proteins_api_search` dropped the caller's organism on retry
`{"query":"FT1","organism":"Triticum aestivum","size":2}` returned `A0A023EZA4_TRIIF` — *Triatoma infestans*, an assassin bug — as a success, with the note *"No human matches found; the search was automatically widened to all organisms. Pass an explicit 'organism' … to scope it."* The caller had passed one. When the first search comes back empty the tool retries under the other field, and the retry parameters were rebuilt from scratch keeping only `format`/`size`/query, so `organism=`/`taxid=` never survived. `{"query":"FT1","taxid":"4565"}` and `{"query":"FT1","taxid":"9606"}` returned byte-identical results.

The retry now carries the original parameters forward and changes only the field being retried. The human-default widening (Feature-69A-007) is unaffected — it may still drop the `taxid=9606` this tool injects itself, and only that.

- Before: 2 hits, first is *Triatoma infestans*, echoed URL `?format=json&size=2&protein=FT1`
- After: `count: 0`, no note, echoed URL `?gene=FT1&size=2&organism=Triticum+aestivum&format=json`
- `{"query":"blaKPC","size":3}` still widens and still emits the note; `{"query":"CYP2D6","size":2}` still human-first

### KEGG cross-references handed out a substance ID labelled as PubChem
`KEGG_get_compound {"compound_id":"C05443"}` returned `dblinks.PubChem: "7805"`. KEGG deposits its records into PubChem as *substances*, so that is SID 7805 — and CID 7805 is 1-bromo-4-methylbenzene, C7H7Br. Feeding it to `PubChem_get_compound_properties_by_CID` returns that molecule as a clean success. Same for acrylamide: C01659 links `4808`, and CID 4808 is an unrelated C36 compound.

Verified against PubChem PUG-REST: `/substance/sid/7805/cids` → `5280795` (cholecalciferol) and `/substance/sid/4808/cids` → `6579` (acrylamide). The key is now `PubChem_SID`, and the schema documents how to resolve it.

### `NCBIVariation_rsid_lookup` reported every variant one base off
`grch38_placements[].position` was dbSNP's 0-based interbase SPDI offset, emitted under a description promising GRCh38 genomic coordinates. `rs121965019` (IDUA c.1205G>A, p.Trp402Ter) came back as chr4:1002746 where `ClinVar_get_variant_details`, `EnsemblVEP_annotate_hgvs` and `MyVariant_query_variants` all say 1002747; the tool's own documented example `rs429358` reported 44908683 rather than 44908684.

`position` is now 1-based, `coordinate_system` states the convention, and the raw offset is kept as `spdi_position`. The per-allele duplicate gene rows (IDUA listed twice for a single-gene variant) are collapsed.

### The integration suite rewrote 2,604 committed files
`tests/integration/test_coding_api_integration.py` called `generate_tools()` with no `output_dir`. That default is the committed `src/tooluniverse/tools/` tree, so simply running the integration suite rewrote 2,604 tracked wrapper files — including 119 that lose their `dict[str, Any]` return annotation — leaving a diff indistinguishable from real work. Every call site already allocates a temp dir and is commented *"Generate tools in temp directory"*; it just was not passed. Guarded with an AST test so it cannot silently regress.

## Verification

- New: `tests/unit/test_orphanet_subtype_name_boundary.py`, `test_disease_profile_identifier_match.py`, `test_ncbi_variation_placement_coordinates.py`, `test_kegg_dblinks_pubchem_namespace.py`, `test_coding_api_generation_is_sandboxed.py`, plus a case added to `test_proteins_api_organism_widening.py` — 29 tests, all passing
- Pre-existing suites re-run green: `test_orphanet_gene_fetch_failure_visibility`, `test_orphanet_classification_hierarchy`, `test_orphanet_invalid_code_validation`, `test_orphanet_gene_name_slash_encoding`, `test_proteins_api_accession_routing`, `test_proteins_api_organism_widening`, `test_genomic_variation_archives_depth`, `tests/tools/test_orphanet_tool.py`
- `tests/unit` was green on `origin/main` before the change and stays green
- Every fix re-verified live through `tu run` with the original failing arguments

No overlap with #493, #494, #495 or #496 — none of them touch `orphanet_tool.py`, `compound_disease_tool.py`, `kegg_ext_tool.py`, `ncbi_variation_tool.py` or `proteins_api_tool.py`.

### DIFF
--- a/src/tooluniverse/compound_disease_tool.py
+++ b/src/tooluniverse/compound_disease_tool.py
@@ -5,10 +5,42 @@ returning identifiers, associated genes, phenotypes, and prevalence data
 in a single call.
 """
 
+import re
 from typing import Any, Dict, List
 
 from .base_tool import BaseTool
 from .tool_registry import register_tool
+
+# Disease nomenclature mixes Roman and Arabic subtype numbers freely --
+# "Mucopolysaccharidosis type I" (query), "Mucopolysaccharidosis Type I"
+# (NCIT) and "Mucopolysaccharidosis type 1" (Orphanet) are one disease.
+_ROMAN_TO_ARABIC = {
+    "i": "1",
+    "ii": "2",
+    "iii": "3",
+    "iv": "4",
+    "v": "5",
+    "vi": "6",
+    "vii": "7",
+    "viii": "8",
+    "ix": "9",
+    "x": "10",
+}
+
+
+def _normalize_disease_label(text: str) -> str:
+    """Fold a disease name to a comparable form: lowercase, punctuation-free,
+    with Roman subtype numerals rewritten as Arabic."""
+    tokens = re.split(r"[^0-9a-z]+", (text or "").lower())
+    return " ".join(_ROMAN_TO_ARABIC.get(t, t) for t in tokens if t)
+
+
+def _labels_match(query: str, label: str) -> bool:
+    """True when `label` names the same disease the caller asked for."""
+    normalized_query = _normalize_disease_label(query)
+    return (
+        bool(normalized_query) and _normalize_disease_label(label) == normalized_query
+    )
 
 
 def _truncate_msg(msg: str, limit: int = 240) -> str:
@@ -61,7 +93,7 @@ class CompoundDiseaseProfileTool(BaseTool):
         r = _try_tool(
             "Orphanet", "Orphanet_search_diseases", {"query": disease, "language": "en"}
         )
-        sections["orphanet"] = self._parse_orphanet(r)
+        sections["orphanet"] = self._parse_orphanet(r, disease)
 
         # 2. OMIM
         r = _try_tool("OMIM", "OMIM_search", {"query": disease, "limit": 5})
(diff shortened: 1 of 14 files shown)


gasvn:

PR #498, opened 2026-08-10
Fix Round 25: results that look complete or point at the wrong culprit

gasvn:
Seven tools that returned a well-formed, successful-looking response while quietly omitting — or misattributing — the thing that was asked for. Every one was reproduced through the CLI before any code was touched, and every root cause was confirmed against real upstream data rather than inferred.

Three came from persona sweeps in domains this harness had not covered before (veterinary/comparative biology, reproductive & developmental genetics, microbiome/metagenomics); four were carry-forwards re-verified from scratch.

---

## 1. DrugBank chemistry lookups returned no formula or weight

`drugbank_get_drug_chemistry_by_drug_name_or_drugbank_id` advertises "molecular formula, weight, and structure" and returned `""` for both on **every small molecule**, sitting next to a correctly populated `melting_point` and `water_solubility` — so the blanks read as "DrugBank has no value" rather than "we looked in the wrong element".

DrugBank files Molecular Formula/Weight under `<calculated-properties>` for small molecules, and under `<experimental-properties>` only for biotech entries. The config mapped both fields at the experimental section alone. Dumping DB00602 from the cached dataset confirms it: `calculated-properties` holds `Molecular Formula = C95H146O28`, while `experimental-properties` holds only Water Solubility and Melting Point.

Swapping one hard-coded section for the other would just move the blanks onto the biologics, so instead a flat `field_mappings` value may now be a **list of XPath expressions, tried in order, first non-empty wins**. `ElementTree.findall()` does not accept `|` unions, which is why a list is the right shape. Single-string mappings and the nested `parent_path` branch are byte-identical.

```
$ tu run drugbank_get_drug_chemistry_by_drug_name_or_drugbank_id '{"drug_name": "Ivermectin"}'
before: "molecular_formula": "",           "molecular_weight": ""
after:  "molecular_formula": "C95H146O28", "molecular_weight": "1736.185"
```
Aspirin (DB00945) `C9H8O4` / `180.1574`; the biotech entry Lepirudin still resolves `C287H440N80O111S6` through the second list element, exercising the fallback.

## 2. `ChEBI_search` silently capped every result set at 15

The tool accepted `limit` up to 100, then returned at most 15 compounds regardless, reporting `result_count: 15` with nothing marking the answer as partial. The EBI `advanced_search` endpoint paginates at 15 hits/page and reports the true size in `total`/`number_pages`; the tool issued one POST and discarded both. A survey of "flavonoid" therefore looked like 15 hits when ChEBI holds 116 — an 8x undercount that reads as a complete answer.

It now walks successive pages until `limit` is satisfied or the set is exhausted, and surfaces the API's count as **`total_matches`**, so a caller can tell "10 returned of 116 matching" from "only 10 exist". `result_count` keeps its existing meaning. A failure fetching page N > 1 degrades to fewer results instead of discarding pages already in hand; absent page metadata degrades to the previous single-page behaviour rather than looping.

```
$ tu run ChEBI_search '{"query": "flavonoid", "limit": 40}'
before: result_count 15
after:  result_count 40, total_matches 116, 3 requests
```
`limit=5` still costs one request; the default of 10 is unchanged; `limit=100` returns 100 unique accessions in 7 requests.

## 3. `visualize_molecule_2d` could not resolve any molecule name

`_resolve_molecule_name()` returned `None` for every input, making the `molecule_name` parameter inert — the tool was drivable only by explicit SMILES/InChI.

PubChem PUG-REST renamed its SMILES properties (`IsomericSMILES` → `SMILES`, `CanonicalSMILES` → `ConnectivitySMILES`). Requesting the legacy name still works, but the response comes back keyed by the **new** name, and the code tested for the legacy key before reading it, so the lookup always fell through:

```
$ curl ".../compound/name/aspirin/property/IsomericSMILES/JSON"
{"PropertyTable":{"Properties":[{"CID":2244,"SMILES":"CC(=O)OC1=CC=CC=C1C(=O)O"}]}}
```

Now requests the current name and reads either key, matching the convention `metabolite_tool.py` already uses. aspirin and caffeine resolve; an unknown name still returns `None`.

## 4. PubMed results dropped the DOI for pre-2008 literature

`PubMed_search_articles` returned `"doi": null` for essentially every article published before ~2008 — a null that reads as "no DOI was assigned" rather than "we only looked in one place". For the canine MDR1 question that stripped DOIs from the entire foundational Mealey/Geyer-era literature while modern papers looked fine.

The esummary parser read the DOI only from `elocationid`, then iterated `articleids` matching just `idtype == "pmc"`, never `idtype == "doi"`. NCBI leaves `elocationid` empty for older records but always lists the DOI in `articleids`:

```
15720521 | eloc=''                        | articleids doi = 10.1111/j.1365-2885.2004.00625.x
17354009 | eloc=''                        | articleids doi = 10.1007/s00204-007-0193-6
41837342 | eloc='doi: 10.1002/jat.70166'  | articleids doi = 10.1002/jat.70166
```

The structured `articleids` entry is now primary and `elocationid` the fallback: it is populated for both eras, its value is bare rather than the mixed `pii: … doi: …` free text, and it needs no parsing heuristic. The same defect in the efetch XML path is fixed alongside it — PMID 15720521's efetch record has no `<ELocationID>` at all but does carry `<ArticleId IdType="doi">`.

Six previously-null PMIDs now resolve (back to a 1969 record); a modern record is unchanged with no `doi: ` prefix leaking in; an article genuinely without a DOI still returns `None`.

## 5. Every Bgee call failed with HTTP 403, including its own example

All three Bgee tools were unusable — `Bgee_get_gene_expression` returned `HTTP error: 403` even for the exact gene its description advertises.

**The obvious hypothesis — a blocked `python-requests` user agent — is wrong.** `BGEE_BASE_URL` lacked a trailing slash, so requests built `https://www.bgee.org/api?page=…`, and bgee.org's Cloudflare edge challenges that path before it ever reaches the application. Bisected against the tool's exact URL and params:

| Request | python-requests UA | curl UA | Chrome UA |
|---|---|---|---|
| `/api?page=…` | 403 | 403 | 403 |
| `/api/?page=…` | 200 | 200 | — |

The 403 carries `cf-mitigated: challenge` / `server: cloudflare`; a bare `GET /api/` returns Bgee's *own* JSON 404 envelope, proving the slashed path reaches the app and the slashless one does not. One character repairs all three tools, which shared the constant. HTTP failures now also surface Bgee's own JSON message, so an invalid gene ID explains itself.

## 6. OpenTargets expression returned 25 of ~1400 rows as if complete

`OpenTargets_get_target_expression_by_ensemblID` reported `count: 1409` and returned 25 rows, with no pagination parameter at all — the other 1384 unreachable. The GraphQL query asked for `baselineExpression` without a `page` argument, so the API's default page size of 25 applied silently. The description even advised filtering client-side by `datasourceId`, which is impossible when 98% of rows never arrive.

The 25 are not a representative sample. For NLRP7 they held exactly one GTEx tissue (pituitary gland, median 0.01) and omitted **testis at 7.53 — the gene's highest-expressing GTEx tissue**. The result was a confident, well-formed expression profile that inverts the answer.

Now passes an explicit page, exposes `size` (default 250) and `index`, and makes partial-ness explicit rather than leaving the caller to compare `count` against `len(rows)`: the response carries `returned`, `truncated` and the `page` actually applied, plus a metadata note when rows remain. Because rows from one datasource are interleaved across the whole result, that note says so — a partial page is not a complete view of any single source.

Live-probed rather than assumed: max `size` is 3000 (`size=5000` → *"the size must be between 0 and 3000"*), and `$index: Int!` with a default is rejected by this server, so the query uses nullable variables. `size=3000` returns all 1409 with `truncated: false`; testis is reachable at the default 250.

## 7. ClinVar blamed the gene when the significance filter was invalid

`ClinVar_search_variants` pasted an unrecognised `clinical_significance` straight into the Entrez term, returned `status: success` with zero results, and then told the user to go re-verify their gene symbol:

```
$ tu run ClinVar_search_variants '{"gene":"TP53","max_results":3,"clinical_significance":"Banana"}'
"total_count": 0,
"query_translation": "TP53[gene] AND (\"clinsig banana\"[Filter] OR clinsig_banana[prop])",
"zero_result_hint": "No ClinVar records matched gene 'TP53'. …verify the symbol…"
```

TP53 is about as valid as a gene symbol gets. The hint pointed at the one input that was never the problem, while the real culprit was visible only to someone who thought to read `query_translation`.

Validated at the input instead of diagnosed after the fact: an unrecognised class is now rejected before any search runs, with the accepted list in the message. A silently-wrong filter reporting success is the actual defect — rewording the hint would have left it in place. The genuine gene-symbol hint still fires unchanged for a real unknown symbol, and valid filters are untouched: NLRP7 still returns 434 unfiltered and 34 for `Pathogenic`, matching NCBI's own esearch count.

---

## Verification

- `pytest tests/unit` — green before and after, **0 regressions**. New tests added for six of the seven fixes (56 new assertions across five files), all network-free.
- `ruff check` on every modified module: finding counts identical to their `HEAD` baselines — **zero new lint findings**. `ruff format --check` clean on all changed files.
- `git status --short src/tooluniverse/tools/` — **0**. The bulk generator was never run; the two affected wrappers were produced per-tool and match generator output byte-for-byte, including the `[:77]`/`[:97]` docstring truncations.
- Every fix was re-run through the CLI with its original failing arguments.

## Overlap with open PRs #493–#497

Checked before queuing anything. Deliberately **not** fixed here because a sibling PR already owns the file and the symptom:

- UniProt `organism` rejecting non-model species with HTTP 400 — #494 rewrites exactly this path (routes names to `organism_name:`).
- HPA `..._in_specific_tissues` reading the tissue-*enrichment* summary instead of the full panel — #495 fixes precisely this.
- openFDA `limit` not clamped to its documented max — #493 owns `openfda_tool.py`.

The one shared file is `data/opentarget_tools.json`, which #494 also edits — but only its `OpenTargets_get_evidence_by_datasource` entry. This PR's diff there is two hunks, both inside the expression tool's entry; the two do not overlap. `graphql_tool.py` is touched by #494/#495 as well; the change here is one new constant plus a name-gated block, so a rebase conflict would be small and mechanical if those merge first.

### DIFF
--- a/src/tooluniverse/bgee_tool.py
+++ b/src/tooluniverse/bgee_tool.py
@@ -16,7 +16,12 @@ from typing import Dict, Any
 from .base_tool import BaseTool
 from .tool_registry import register_tool
 
-BGEE_BASE_URL = "https://www.bgee.org/api"
+# NOTE: the trailing slash is REQUIRED. Requests to "https://www.bgee.org/api"
+# (no slash) never reach the Bgee application -- they are intercepted by the
+# site's Cloudflare WAF and answered with an HTTP 403 challenge page
+# ("cf-mitigated: challenge"). Only "https://www.bgee.org/api/?..." is routed
+# to the API. This is independent of User-Agent and of any request header.
+BGEE_BASE_URL = "https://www.bgee.org/api/"
 
 
 @register_tool("BgeeTool")
@@ -54,13 +59,42 @@ class BgeeTool(BaseTool):
         except requests.exceptions.HTTPError as e:
             return {
                 "status": "error",
-                "error": f"Bgee API HTTP error: {e.response.status_code}",
+                "error": self._http_error_message(e),
             }
         except Exception as e:
             return {
                 "status": "error",
                 "error": f"Unexpected error querying Bgee: {str(e)}",
             }
+
+    def _http_error_message(self, exc: requests.exceptions.HTTPError) -> str:
+        """Build an actionable message from a failed Bgee HTTP response.
+
+        Bgee answers bad requests with a JSON body that carries a human
+        readable ``message`` (e.g. an unknown gene ID yields HTTP 404 with
+        ``"Page not found."``), so surface that instead of a bare status code.
+        """
+        response = exc.response
+        status = getattr(response, "status_code", None)
+        api_message = ""
+        try:
+            body = response.json()
+            if isinstance(body, dict):
+                api_message = str(body.get("message", "")).strip()
+        except (ValueError, AttributeError):
+            # non-JSON body (HTML error page) or no response attached
+            api_message = ""
+
+        message = f"Bgee API HTTP error: {status}"
+        if api_message:
+            message += f" - {api_message}"
+        if status == 404:
+            message += (
+                " (check that the gene ID is a valid Ensembl ID present in Bgee"
+                " for the requested species, e.g. gene_id=ENSG00000141510 with"
+                " species_id=9606)"
+            )
+        return message
 
     def _query(self, arguments: Dict[str, Any]) -> Dict[str, Any]:
         """Route to appropriate Bgee endpoint."""
--- a/src/tooluniverse/chebi_tool.py
+++ b/src/tooluniverse/chebi_tool.py
@@ -20,6 +20,15 @@ from .tool_registry import register_tool
 
 CHEBI_BASE_URL = "https://www.ebi.ac.uk/chebi/backend/api/public"
 
+# advanced_search is server-side paginated at 15 hits per page and exposes the
+# page number as a 1-indexed query-string parameter (`?page=1` is the first
+# page and is byte-identical to omitting the parameter; `?page=0` and any page
+# beyond `number_pages` return a non-JSON error body). `limit` is clamped to
+# 100, so at most ceil(100 / 15) == 7 pages are ever needed; the cap below is
+# a hard stop that protects against a server that stops honouring `page`.
+_SEARCH_PAGE_SIZE = 15
+_SEARCH_MAX_PAGES = 8
+
 _HTML_TAG_RE = re.compile(r"<[^>]+>")
 
 
@@ -202,7 +211,11 @@ class ChEBITool(BaseTool):
 
         if limit is None:
             limit = 10
-        limit = min(limit, 100)
+        try:
+            limit = int(limit)
+        except (TypeError, ValueError):
+            limit = 10
+        limit = max(0, min(limit, 100))
 
         # Use advanced_search endpoint for better relevance
         url = f"{CHEBI_BASE_URL}/advanced_search/"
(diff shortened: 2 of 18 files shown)


gasvn:

PR #500, opened 2026-08-10
Fix Round 26: answers that look complete or name the wrong record

gasvn:
Six issues surfaced by three researcher personas (neuro-oncology, single-cell atlases/aging biology, transplant immunology), each CLI-verified against live APIs before any fix was written.

The shape this round hunted: a well-formed response that quietly stands in for a larger or different answer, leaving the caller no way to tell it is incomplete.

## Fixes

**CIViC — a gene-prefixed variant name returns the wrong, empty record** (`civic_tool.py`)
CIViC filters variant and molecular-profile names by substring and stores variants under a bare designation (`VIII`, profile `EGFR VIII`), but clinical writing joins gene and designation into one token (`EGFRvIII`). A joined token can never be a substring of the separated record, so:
- `civic_search_variants{gene_name:"EGFR", variant_name:"EGFRvIII"}` returned only an empty duplicate stub (id 1516, `molecularProfileScore 0.0`, no HGVS) and dropped the curated record (id 312, score 69.0).
- `civic_search_evidence_items{molecular_profile:"EGFRvIII"}` returned 1 of 7 reachable evidence items, losing the Rindopepimut (971), gefitinib/erlotinib (1128) and EGFR-TKI class (772) glioblastoma records.

Gene-prefixed names are now split and unioned with the bare designation, both spellings disclosed via the existing `normalization_note`. Entity-agnostic — the designation patterns require a character after the digits, so `ERBB2`/`TP53` never split. `molecular_profile:"BRAFV600E"` goes from 0 items to the full `BRAF V600E` set.

**OLS — `exact_match` did nothing and reported that it had** (`ols_tool.py`)
OLS4's `exact` flag selects *how* the query matches, not *what* it matches against; the default field set includes `description`, so "exact" matched nearly everything: `fibroblast`/cl returned 167 of 168 terms, `T cell`/cl returned 6803 including `mesenchymal stem cell`. The response then echoed `exact_match: true`, asserting a filter that had removed one result in 168. Constraining `queryFields` is what makes the flag restrictive (`label,synonym`; `obo_id,short_form,iri` for CURIE/IRI queries so identifier lookups keep working). `filters.exact_match_fields` now makes the claim checkable. Non-exact queries are byte-for-byte unchanged.

**FDA labels — every section silently cut at 2000 characters** (`fda_label_tool.py`)
A hard-coded, undocumented, non-overridable slice applied to all eleven prose sections. Prograf's `clinical_pharmacology` is 19,517 chars and its CYP3A5 expresser/non-expresser dosing guidance — the single most important tacrolimus dosing fact — sits ~6,600 chars in, so the tool reported in fluent prose that the tacrolimus label says nothing about CYP3A5, while its own description promised "all clinical sections". Truncation is now disclosed at the top level (`truncated`/`truncated_fields`/`truncation_note`) and the budget is caller-controllable via `max_section_chars`.

**FAERS — 100 rows described as "all"** (`openfda_adv_tool.py`)
Count queries sent no `limit`, so openFDA returned exactly its default 100 rows. Progressive multifocal leukoencephalopathy (561 reports) ranks 202nd for tacrolimus and was therefore absent — a false negative in a pharmacovigilance review. All 18 count tools now accept `limit` (default 100, max 1000) and report `result_count`/`truncated`; the fetch requests `limit+1` so an exactly-full page is not misreported as clipped. openFDA returns no grand total for count aggregations, so none is invented. Success shape changes from a bare list to an object to carry the disclosure.

**CPIC — advertised provenance URL did not reproduce the result** (`cpic_search_pairs_tool.py`)
PostgREST filters travelled in query params while the reported `url` was the bare endpoint template, so the 13-row filtered answer and the 635-row unfiltered answer cited byte-identical URLs. The data was always correct; only the citation was wrong. The reported URL is now taken from the request actually issued, so the two cannot drift.

**Parameter validation — the framework blamed the caller for its own injected argument** (`base_tool.py`, `execute_function.py`)
`_apply_operation_default` injects an `operation` argument before validation; the validator then counted it as caller-supplied. `Pharos_get_target_expression{"target":"EGFR"}` answered `Unrecognized parameter(s): 'target', 'operation'` — naming a parameter the caller never sent. The injected value is now excluded from the report only when it matches exactly what the config would supply, so a caller-supplied bogus `operation` is still reported.

This also restores the unknown-parameter guard for the 348 configs that *declare* `operation`, where the injected key had been padding the `len(unknown) == len(arguments)` denominator and silently suppressing it. A differential sweep across all 2,573 tool configs and 4,356 test examples: **6 tools fixed** (no-arg calls like `PanglaoDB_list_cell_types` that previously errored), **152 all-bogus calls newly rejected** (the intended tightening), **0 tools newly broken**.

## Verification

Every fix re-run through the CLI with its original failing arguments; each now returns the complete or correctly-identified answer.

- `tests/unit`: 3816 collected, **0 failed**.
- `tests/tools` + `tests/examples`: 5 failures, **reproduced identically on a clean `origin/main` worktree** — pre-existing and unrelated (BRENDA credentials, Complex Portal 503, Ensembl homology timeout, CLI rendering, embedding ranking).
- Validator sweep over the changed surface: all 4 JSON configs valid; all 18 rewritten FAERS `return_schema`s match the new shape with zero double-wrapped envelopes; 21 hand-edited wrappers consistent with their configs; no new lint findings in the project's enforced rule set.
- Grep for tests asserting the old behaviour was run per-fix and at round level. Two test files legitimately encoded the old FAERS list shape and are updated; nothing else did.

Deliberately avoided every file owned by open PR #499 to keep this conflict-free.

### DIFF
--- a/src/tooluniverse/base_tool.py
+++ b/src/tooluniverse/base_tool.py
@@ -16,6 +16,28 @@ import hashlib
 import inspect
 
 
+def resolve_configured_operation(tool_config: Any) -> Optional[str]:
+    """Return the ``operation`` a tool's own config implies, if any.
+
+    Many multi-operation tool classes are registered once per operation and
+    still read ``arguments["operation"]``, so ToolUniverse fills that key in
+    from the tool's own config (``fields.operation``, else the schema default)
+    before the tool runs -- see
+    ``ToolUniverse._apply_operation_default``. Both the injection and the
+    validation-side recognition of the injected key resolve the value through
+    this single function, so the two cannot drift apart.
+    """
+    if not isinstance(tool_config, dict):
+        return None
+    operation = (tool_config.get("fields") or {}).get("operation")
+    if not operation:
+        schema = tool_config.get("parameter") or {}
+        prop = (schema.get("properties") or {}).get("operation")
+        if isinstance(prop, dict):
+            operation = prop.get("default")
+    return operation if isinstance(operation, str) and operation else None
+
+
 class BaseTool:
     STATIC_CACHE_VERSION = "1"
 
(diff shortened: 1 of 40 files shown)


gasvn:

PR #501, opened 2026-08-10
Fix Round 27: answers that confidently name the wrong entity

gasvn:
Five verified issues from role-play rounds across ophthalmology / inherited retinal disease, psychiatry & neuropharmacology, and hepatology. Every issue below was reproduced via `tu run` before any code was written, and re-run after.

Common shape this round: the tool returns `status: success` with a plausible, well-formed payload that is about a different entity than the caller asked for — or reports a number that does not mean what it appears to mean.

## Fixes

**`xml_tool.py` — DrugBank/MeSH search answered "Aspirin" with caffeine's chemistry.**
`XMLTool._search()` truncated to `limit` while scanning in document order and never considered match quality, even though every config already declares `search_fields` in priority order. 24 records matched "Aspirin"; the 10 returned were all combination products whose *brand names* contain the word, and DB00945 `Acetylsalicylic acid` — an exact synonym hit — was pushed off the page entirely. Now ranks before truncating on (whole-value match before substring, then declared field priority), stable by document order. Generic across all 19 tools of this class; no output field changes shape.

**`marrvel_tool.py` — a previous gene symbol returned a bare empty success.**
MARRVEL indexes only current approved HGNC symbols and answers an unknown one with HTTP 200 and an empty body, which was passed through with no note — indistinguishable from "this gene does not exist". `RP20` and `LCA2` are the previous symbol and an alias of `RPE65`, and the tool lists both in its own RPE65 record. Now resolves once against HGNC's public previous-symbol/alias search and retries once, **disclosing the substitution** so the caller is never silently answered about a different symbol. Ambiguous aliases are reported, never guessed. Any HGNC failure degrades to the previous empty result, now with an actionable note.

**`uniprot_idmapping_tools.json` — a mouse question was answered with human proteins.**
The schema left `additionalProperties` open, so `organism: "mouse"` was accepted and dropped while `tax_id` silently defaulted to 9606 — a cross-species wrong answer, not a null. The framework's unknown-key check deliberately tolerates a valid key mixed with an unrecognised one, so this is closed at the tool's own schema, as `search_clinical_trials` already does. Applied to all four tools in the file after confirming none accepts pass-through keys and none receives an injected `operation`. The `tax_id` description also claimed "Required." while the schema did not require it.

**`ctg_tool.py` — the primary trial-search tool matched study prose, not interventions.**
`search_clinical_trials` passed `intervention` straight through as `query.intr`, which Essie matches against the whole study record. Asking for Luxturna returned a single result: an **RDH12** gene-therapy trial whose only tie to the query was "using a method similar to approved gene therapies like Luxturna" in its brief summary, while all five real voretigene neparvovec trials were missed. The `AREA[InterventionName]`/`[InterventionOtherName]` restriction already existed in this file but had only been wired into the sibling `_run_search` path. Now reuses the same helper and the same shared mapping so the two cannot drift apart again. This also removes false positives from ordinary queries: `lung cancer` + `pembrolizumab` drops 504 → 474 and stops returning studies with no pembrolizumab intervention.
Restricting to name fields exposes a second gap — a brand name recorded in neither field now returns a clean zero. Confirmed against the raw API that Luxturna appears in neither while the INN returns three studies. An empty result now says the search matched registered names and synonyms only, suggests the generic name, and names whichever filter actually applied rather than blaming the phase gate for an intervention miss.

**`clinvar_tool.py` — a free-text count was reported as though disease-filtered.**
`RPE65` + `"Lebers congenital amaurosis"` — one apostrophe from the indexed name — returned `total_count: 1` with `status: success` and no warning, against 996 for the canonical spelling. This is *not* Entrez rewriting the field tag: live queries confirm `[dis]` is preserved verbatim and simply matches nothing, and the `[All Fields]` number comes from this module's own untagged retry, which adopted the free-text count and presented it identically to a real one. The response now states at top level whether the disease filter was actually applied, and when it was not, that the count includes records merely mentioning the text anywhere. Detection reads the query translation the tool already receives — no extra request.
Relatedly, an empty result never considered the condition as the culprit even when it was the filter that excluded everything. Zero results now probe the condition on its own and report whether it is a recognized disease term, keeping the existing gene-symbol guidance for cases where the gene really is the problem.

## Verification

- `pytest tests/unit` — full suite green after all five fixes (0 failures).
- 5 new unit test files, all with the HTTP layer mocked (no live network, no large-dataset dependency): `test_xml_search_ranks_by_match_quality.py`, `test_marrvel_previous_symbol_resolution.py`, `test_uniprot_idmap_rejects_unknown_species_param.py`, `test_ctg_search_intervention_area_restricted.py`, `test_clinvar_condition_filter_not_applied.py`.
- Each fix was re-run live through the CLI with the original failing arguments, plus a control call that must keep working.
- Grepped `tests/` for assertions on the old ordering, response shape, and hint text for every changed symbol; nothing required updating.
- `src/tooluniverse/tools/` is untouched apart from one hand-edited wrapper docstring (`UniProtIDMap_gene_to_uniprot.py`); the generator was never run.

## Not included

Two findings reproduced but were already fixed by #499 once it merged, and were re-checked against the merged tree rather than re-fixed: DrugBank `molecular_formula`/`molecular_weight` returning empty, and intervention search missing trials that register a drug under its development code (resmetirom/MGL-3196, `NCT03900429`).


gasvn:

PR #502, opened 2026-08-10
Fix Round 28: answers that quietly stand in for the question asked

gasvn:
Four tools returned a well-formed, confident response to a question the caller did not ask, with nothing in the payload to reveal the substitution. All four were reproduced live via the CLI before any code was written, and each root cause was confirmed against the upstream API rather than inferred.

## 1. `PubMed_search_articles` — a phrase that matched nothing was silently dropped

```
tu run PubMed_search_articles '{"query":"\"JAK2 V617F thrombosis unicornwombatxyz\"","limit":3}'
tu run PubMed_search_articles '{"query":"JAK2 V617F thrombosis","limit":3}'
```
Both returned `total: 540` and the identical top three PMIDs. The token `unicornwombatxyz` appears in no indexed article and had zero effect.

NCBI esearch drops quoted phrases it cannot match, answers a broader query, and reports exactly that in `esearchresult.warninglist` (`quotedphrasesnotfound`, `phrasesignored`) plus `querytranslation`. The tool read only `idlist` and `count`, so the disclosure never reached the caller.

Now propagated into `metadata` at both search return points. Additive only — queries NCBI does not warn about keep their existing metadata shape byte for byte.

## 2. BV-BRC — a page of rows reported as the complete match set

`metadata.total_results` was `len(results)`, so it always equalled the requested `limit`:
```
tu run BVBRC_search_genome_features '{"gene":"blaOXA-48","limit":3}'   # total_results: 3
tu run BVBRC_search_genome_features '{"gene":"blaOXA-48","limit":10}'  # total_results: 10
```
A surveillance user counting how far a resistance gene has spread could not tell truncation from completeness. BV-BRC already sends the true total in the `Content-Range` header (`items 0-3/13` for the query the tool actually issues), and exposes it via CORS.

Now parsed and reported as `returned_results` (this page) vs `total_results` (upstream), with explicit `truncated` + `truncation_note`. Both limits now report 13. Routed through one shared helper rather than ten copies of the same metadata block; header parsing is fully defensive and degrades to the old page-count behaviour when the header is absent or malformed.

## 3. `gather_gene_disease_associations` — answered about a different gene

```
tu run gather_gene_disease_associations '{"gene":"THP"}'
```
THP is the clinical alias for uromodulin. The tool returned 25 associations headed by "holoprosencephaly 9" (0.80) — **GLI2's** diseases — shaped exactly like a correct answer.

The OpenTargets target lookup is a fuzzy free-text search, so an unrecognised symbol still returns a ranked hit list; the code fell back to `hits[0]`. The hits carry only `id`/`name`/`description` with no synonym fields, so the alias is genuinely unresolvable here. Rather than guess, the tool now reports the symbol unresolved and names the candidates returned (GLI2, UMOD, THPO, PROS1, ...) so the caller can re-query. No alias table was added — a growing alias list is the anti-pattern this avoids. `{"gene":"UMOD"}` still returns its 27 kidney associations unchanged.

## 4. `proteins_api_search` — two paths to a silent empty for data that exists

**Asymmetric retry.** The empty-result retry swapped the search field only in the gene→protein direction. The routing heuristic sends anything over 10 characters to `protein=`, so `blaCTX-M-15` (the most common ESBL gene worldwide, 11 chars) was never retried as `gene=`; with a caller-supplied `organism` the auto-widen branch was skipped too, leaving no retry at all. Confirmed upstream that `gene=blaCTX-M-15&organism=Klebsiella pneumoniae` returns A0A077JMT3 while `protein=` returns `[]`. The swap is now symmetric.

**Numeric organism.** `organism` takes a name and `taxid` takes a number; `{"gene":"JAK2","organism":"9606"}` matched nothing and returned an empty success. A purely numeric value is unambiguously a taxonomy id, so it is routed to `taxid=` and the correction is disclosed in `normalization_note` — silent input mutation is the same class of defect this PR is about.

## Verification

- Full `tests/unit` suite: passes (exit 0, skips only for absent optional deps).
- All 49 pre-existing tests covering the touched areas pass unchanged; no pre-existing assertion was modified.
- Four new focused unit tests, all fully mocked (no network): `test_pubmed_unmatched_phrase_disclosed.py`, `test_bvbrc_total_reflects_upstream_count.py`, `test_gene_disease_no_wrong_target_substitution.py`, `test_proteins_api_search_no_silent_empty.py`.
- Every fix and every regression guard re-run live through the CLI.
- No files touched that are owned by the open round-26 (#500) or round-27 (#501) branches; `base_tool.py` and `execute_function.py` untouched.
- Tracked wrapper tree (`src/tooluniverse/tools/`, `_lazy_registry_static.py`) verified clean; no generator was run.

Two known-genuine defects in #500's territory (`ClinicalTrials_search_by_intervention` ignoring `phase`/`max_results`, `FDA_search_drug_labels` dropping `query`) were recognised and deliberately left alone.

### DIFF
--- a/src/tooluniverse/bvbrc_tool.py
+++ b/src/tooluniverse/bvbrc_tool.py
@@ -117,13 +117,113 @@ class BVBRCTool(BaseTool):
 
         return "&".join(parts)
 
-    def _make_request(self, endpoint: str, query: str) -> Any:
-        """Make a request to BV-BRC API."""
+    @staticmethod
+    def _parse_content_range_total(header: Any) -> Optional[int]:
+        """Extract the upstream match count from a ``Content-Range`` header.
+
+        BV-BRC answers list queries with a header shaped like
+        ``items 0-9/2431``, where the value after the slash is the number of
+        records matching the query upstream -- independent of how many rows
+        the current page returned.
+
+        Returns ``None`` whenever the total cannot be determined (header
+        absent, malformed, or reporting an unknown total as ``*``). This
+        never raises: an unparseable header simply means "total unknown".
+        """
+        if not isinstance(header, str):
+            return None
+        try:
+            total_part = header.rsplit("/", 1)[-1].strip()
+            if not total_part or total_part == "*":
+                return None
+            total = int(total_part)
+        except (ValueError, TypeError):
+            return None
+        return total if total >= 0 else None
+
+    def _make_request_with_total(self, endpoint: str, query: str) -> Any:
+        """Make a request to BV-BRC API, returning ``(payload, upstream_total)``.
+
+        ``upstream_total`` is the number of records matching the query on the
+        server (from the ``Content-Range`` header), or ``None`` when BV-BRC
+        did not report one.
+        """
         url = f"{BVBRC_BASE_URL}/{endpoint}/?{query}"
         headers = {"Accept": "application/json"}
         response = requests.get(url, headers=headers, timeout=self.timeout)
         response.raise_for_status()
-        return response.json()
+
+        response_headers = getattr(response, "headers", None)
+        try:
+            content_range = response_headers.get(
+                "Content-Range"
+            ) or response_headers.get("X-Content-Range")
+        except (AttributeError, TypeError):
+            content_range = None
+
+        return response.json(), self._parse_content_range_total(content_range)
+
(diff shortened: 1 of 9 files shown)


gasvn:

PR #503, opened 2026-08-10
Fix Round 29: partial answers presented as complete ones

gasvn:
Seven verified defects, all found by researcher personas in three domains not previously exercised (vaccine immunoinformatics, epidemiology/global health surveillance, medical imaging/radiogenomics) and each re-verified via the CLI before any fix was written. Root cause for every one was established against the upstream API directly, not inferred from tool behaviour.

The theme this round is **a partial or wrong answer presented as a complete, correct one**.

### 1. World Bank and WHO GHO returned an arbitrary slice as the whole series

`WorldBank_get_indicator` for four countries over 2010-2024 returned 50 of 60 rows with `status: success`, ending part-way through Myanmar — **Myanmar 2010-2019 vanished silently**. `pages: 2` was buried in the upstream metadata at `data[0]`; `count: 2` was the length of the `[metadata, rows]` envelope, not a row count; and no `page`/`per_page` parameter existed, so page 2 was unreachable through the tool.

`WHOGHO_get_indicator_data` sent `$top=20` with no `$orderby`, returning 20 rows in the API's internal `Id` order — an unpredictable, non-chronological sample of a 25-year series (2006, 2013, 2017, 2018, 2021 dropped). A trend line built from it has invisible holes.

Verified upstream: `per_page=100` returns all 60 records (Myanmar 2010 = 1010.53); `per_page` accepts up to 32500 and 400s at 40000. GHO applies `$orderby` before `$top`, and `$count=true` returns `@odata.count: 25`.

Both now report the true total, an explicit `truncated` flag, and a note naming the next step; WHO GHO results are ordered so a capped result is a contiguous prefix; both echo the fully-resolved request URL so a suspected dropped filter can be audited. The two `BaseRESTTool` hooks are opt-in `fields.*` keys and are no-ops for every tool that does not set them.

### 2. Nextstrain reported a total that counted only what it had truncated

`Nextstrain_list_datasets` capped each pathogen's `datasets` array at 10, then summed those capped arrays into `metadata.total_datasets` — reporting **113 for a catalogue of 295**, wrong by 2.6x, next to per-pathogen `dataset_count` fields that told the truth and contradicted it (avian-flu 66, ncov 86). Nothing disclosed the cap and no parameter could page past it.

Verified against `nextstrain.org/charon/getAvailable`: 295 datasets across 23 pathogens. The total is now real, what was returned moved to `listed_datasets`, and the cap is disclosed and controllable. The sibling `get_dataset` cut `available_colorings` at 15 where a real build has 20; it now returns all.

### 3. A GTDB species query returned a different species and the genus count

`GTDB_search_genomes` for `Mycobacterium tuberculosis` returned **Mycobacterium leprae** among the top five and reported `total_results: 13381`.

Root cause proven, not inferred: appending a term GTDB cannot match (`Mycobacterium xyzzyplop`) leaves the count and rows byte-identical to a bare `Mycobacterium` search — the second term is discarded. And 13381 exceeds the 13027 genomes in the genus, so the figure was `Mycobacterium ∪ tuberculosis`. The endpoint ORs query terms.

The documented `filterText` parameter applies an AND substring test; probing established it is case-insensitive but order- and whitespace-literal, so queries are normalized first. The precise lookup runs first, the loose search is a disclosed fallback, and the same query now returns 7869 with all five results *M. tuberculosis*. A genus query still matches the genus.

### 4. PDC promised disease-name search but only matched study titles

`PDC_search_studies` advertises "search by disease name (e.g. 'Breast', 'Lung', 'CCRCC')". Searching `Lung Adenocarcinoma` returned **zero** studies; `LUAD` returned 18. It called `studySearch(name:)`, which compares against the title alone — the documented `Breast` example worked only because CPTAC breast titles contain the word.

Verified upstream: `getPaginatedUIStudy(disease_type: "Lung Adenocarcinoma")` returns `total: 22`, including PDC000153 whose `disease_type` is `"Lung Adenocarcinoma;Other"`. A researcher looking for the matched molecular partner of the CPTAC-LUAD imaging collection was told it did not exist.

Search now covers the curated metadata PDC publishes and reports which field matched, so a title-only hit is never mistaken for curated evidence. Two further defects surfaced and are fixed: the old path silently capped at 100 results, and the description advertised HTAN, which is not a PDC program.

### 5. cBioPortal and Europe PMC hid the totals that reveal a result is partial

`cBioPortal_get_cancer_studies` returned 20 of 539 with `count: 20` and no total — a silent default, not an API limit (`limit: 1000` returns all 539). A persona concluded the NSCLC cohorts they needed were absent from cBioPortal when they were merely past position 20. There was no `offset`, so paging was impossible.

`EuropePMC_search_articles` parsed a payload containing `hitCount: 103` and reported only `metadata.count: 6`.

Both now report the true total under an unambiguous name, keep the returned count separate, and disclose truncation. cBioPortal reports its total only via a `projection=META` probe with paging params absent; the probe is skipped for unpaged endpoints and short-circuits on a short page, so it costs a request only when the answer is genuinely unknown.

### 6. A misspelled TCIA collection was indistinguishable from an empty one

`TCIA_get_modality_values '{"Collection":"NSCLC-Radiogenomics"}'` returned `status: success` with `data: ""`. That collection does not exist — the real name has a space — but every neighbour is hyphenated, so it is the natural typo, and a genuine filter miss on a *valid* collection produced the byte-identical payload.

Root cause: the NBIA v1 API answers every zero-result query with HTTP 200 and a completely empty body, which fell through to the plain-text branch and became `""` — also making `data` a string on the miss path and a list on the hit path, with no `count` at all.

An unknown collection is now an error naming the closest real names; a valid collection with no matches returns an honest empty list with `count: 0` and a note. The collection list is fetched only on the empty-result path and memoized on success, so a transient outage degrades to a caveat rather than a false "does not exist".

### 7. IEDB epitope prediction hung for two minutes and then gave up

`IEDB_predict_mhci_binding`, `IEDB_predict_mhcii_binding` and `IEDB_predict_antigen_processing` each stalled for the full 120 s client timeout and returned "IEDB prediction timed out" — **including on their own registered test examples**. A nine-residue peptide cannot legitimately take two minutes, so the message blamed the clock for something that was never going to finish. The whole T-cell epitope prediction capability was unusable, which for vaccine design is the load-bearing step.

The cause is upstream and total. Probed directly:

```
GET  tools-cluster-interface.iedb.org/tools_api/mhci/        -> 405 in 0.40 s   (host up)
POST tools-cluster-interface.iedb.org/tools_api/mhci/        -> 0 bytes, killed at 90 s
POST tools-cluster-interface.iedb.org/tools_api/processing/  -> 0 bytes, killed at 120 s
POST tools-cluster-interface.iedb.org/tools_api/bcell/       -> HTTP 200, real TSV, 66.8 s
```

The legacy host accepts the POST and never answers. `bcell/` still works, which is why B-cell prediction kept functioning and masked how broad the outage was.

These three tools now run on IEDB's next-generation asynchronous pipeline API (submit + bounded poll, per-request timeouts). `bcell` deliberately stays on the legacy route because that route still works.

Three contract details were established against the live API rather than assumed, each of which would have been a defect if guessed:
- **A failed stage leaves `status: "pending"` indefinitely while populating `data.errors`** — polling on `status` alone reproduces the original hang. Observed for 180 s.
- Validation failures return **HTTP 200 with `errors` and no `result_id`**; an invalid *method* returns an opaque HTML 500.
- Alleles are spelled **`H2-Kd`, not `H-2-Kd`** — which is the MHC-I tool's own first `test_example`, so it needed translating.

| | before | after |
|---|---|---|
| `IEDB_predict_mhci_binding` (own example) | timeout after **2m 01.8s** | real prediction, **16.1 s** |
| `IEDB_predict_mhcii_binding` (own example) | timeout after **2m 01.7s** | real prediction, **16.0 s** |
| `IEDB_predict_antigen_processing` | same 120 s hang | 23 scored peptides, **22.6 s** |
| `IEDB_predict_bcell_epitopes` | working | unchanged |

The MHC-I result for GILGFVFTL / HLA-A\*02:01 is `netmhcpan_el_score 0.789948`, `percentile 0.09` — correct for the influenza M1 immunodominant epitope.

The processing tool's output columns necessarily change: the legacy proteasome/TAP/total triple does not exist in the new API, and NetCTLpan's cleavage/TAP/MHC/combined scores are the closest true equivalent. The description and `return_schema` describe what is actually returned rather than renaming columns to imitate the retired shape.

Three pre-existing test files asserted the retired TSV contract and are rewritten against the new transport, preserving each original guard's intent — most importantly that an upstream failure is never reported as a prediction.

---

**Verification.** Every fix re-verified by re-running the original failing CLI command. `tests/unit` + `tests/integration` pass with zero failures both before and after (86 new assertions across 8 new regression test files, all mocking the HTTP layer). Each worker grepped `tests/` for assertions on the old behaviour; four pre-existing test files needed changes: `test_clinical_imaging_archives_depth.py` (three TCIA fixtures constructed `BaseRESTTool` directly, repointed to `TCIATool` with identical payloads) and the three IEDB files that asserted the retired TSV contract.

**Scope.** Deliberately avoids every file owned by open PRs #500, #501 and #502. Two verified defects were dropped for that reason and are noted for a later round: `proteins_api_get_epitopes` returns an empty success for P03420 while the upstream EBI epitope endpoint holds 47 features, and `BVBRC_search_surveillance` swallows an unknown parameter and blames the caller for the resulting omission.

### DIFF
--- a/src/tooluniverse/base_rest_tool.py
+++ b/src/tooluniverse/base_rest_tool.py
@@ -237,6 +237,87 @@ class BaseRESTTool(BaseTool):
 
         return result
 
+    @staticmethod
+    def _resolve_path(data: Any, path) -> Any:
+        """
+        Walk a list of dict keys / list indices into a decoded JSON payload.
+
+        Returns None if any step is missing, so callers can treat "not found"
+        and "explicitly null" alike (neither is usable as a total/row list).
+        """
+        current = data
+        for step in path or []:
+            if isinstance(step, int) and not isinstance(step, bool):
+                if not isinstance(current, (list, tuple)) or not (
+                    -len(current) <= step < len(current)
+                ):
+                    return None
+                current = current[step]
+            else:
+                if not isinstance(current, dict) or step not in current:
+                    return None
+                current = current[step]
+        return current
+
+    def _apply_pagination_disclosure(self, result: Dict[str, Any]) -> None:
+        """
+        Surface upstream truncation at the top level of the response.
+
+        Opt-in via ``fields.pagination_disclosure``; tools without the key are
+        untouched. Many APIs return a server-side page of a larger result set
+        and report the real total somewhere inside the payload (World Bank puts
+        ``{"page","pages","per_page","total"}`` at ``data[0]``; OData services
+        return ``@odata.count``). Left buried there, a partial answer is
+        indistinguishable from a complete one -- a country panel can lose a
+        whole country and still report ``status: success``.
+
+        Config keys (all paths are lists of dict keys / list indices into
+        ``result["data"]``)::
+
+            {
+              "total_path": [0, "total"],   # int: rows matching upstream
+              "rows_path": [1],             # list: rows actually returned
+              "page_path": [0, "page"],     # optional int
+              "pages_path": [0, "pages"],   # optional int
+              "more_hint": "How to fetch the rest."
+            }
+
+        Sets ``count`` to the real row count (not the length of an envelope),
+        plus ``total_available`` and an explicit ``truncated`` boolean so
+        "complete" and "partial" are never conflated. When truncated, adds a
+        top-level ``truncation_note`` naming the true total.
+        """
+        config = self.tool_config.get("fields", {}).get("pagination_disclosure")
+        if not config:
+            return
+
+        data = result.get("data")
+        total = self._resolve_path(data, config.get("total_path"))
+        rows = self._resolve_path(data, config.get("rows_path"))
+        if not isinstance(total, int) or isinstance(total, bool):
+            return
+        if not isinstance(rows, list):
+            return
+
+        returned = len(rows)
+        result["count"] = returned
+        result["total_available"] = total
+        result["truncated"] = returned < total
+        if returned >= total:
+            return
+
+        page = self._resolve_path(data, config.get("page_path"))
+        pages = self._resolve_path(data, config.get("pages_path"))
+        page_part = ""
+        if isinstance(page, int) and isinstance(pages, int):
+            page_part = f" This is page {page} of {pages}."
+        hint = config.get("more_hint", "")
+        result["truncation_note"] = (
+            f"Partial result: {returned} of {total} matching records were "
+            f"returned; {total - returned} are missing.{page_part} "
+            f"{hint}"
+        ).strip()
+
     def _handle_special_endpoint(
         self, url: str, response: requests.Response, arguments: Dict[str, Any]
     ) -> Optional[Dict[str, Any]]:
@@ -339,6 +420,15 @@ class BaseRESTTool(BaseTool):
                 max_attempts=3,
             )
 
+            # The echoed `url` is the endpoint only -- query params live in
+            # `params` and never appear in it, so a caller cannot tell what was
+            # actually requested (which page? which filter? which $top?) and
+            # cannot reproduce the call. `fields.echo_request_url` opts a tool
+            # into echoing the fully-resolved request URI instead. Guarded with
+            # getattr so stubbed responses in tests stay valid.
+            if self.tool_config.get("fields", {}).get("echo_request_url"):
+                url = getattr(response, "url", None) or url
+
             # Check for errors (accept any 2xx success status)
             if not (200 <= response.status_code < 300):
                 return {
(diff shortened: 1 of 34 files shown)


gasvn:

PR #504, opened 2026-08-10
Fix Round 30: tools that discarded the evidence they already had

gasvn:
Seven tools each returned an answer that looked complete and was not. In every
case the tool had the correcting evidence in hand — in the payload it had
already fetched, in the record it was already serving, or in a field of the
upstream response it chose to drop.

Three researcher personas (rheumatology/autoimmune, endocrinology/metabolic,
toxicogenomics/occupational health) ran ~90 live tool calls. Eight issues were
reported; every one was re-run through the CLI before being queued, and each
root cause below was established against the upstream service directly rather
than inferred from tool behaviour.

## Fixes

**UniProt returned no protein for a species-level bacterial taxid.**
`UniProt_search` scoped numeric taxids with `organism_id`, which UniProt's field
configuration documents as "Organism [OS]" — the entry's own organism node only.
`taxonomy_id` ("Taxonomy [OC]") matches the lineage. Nearly every bacterial entry
is annotated at strain level, so `gene:gyrA AND organism_id:1423 AND reviewed:true`
returned 0 while P05653 (GYRA_BACSU) plainly exists. Measured live against
rest.uniprot.org: 0 vs 1 for B. subtilis, 9 vs 30 for E. coli recA, 105 vs 157 for
M. tuberculosis katG, and 114 vs 114 for human TP53 — single-node taxa are
unaffected, so this is strictly an increase in correctness. The tool's own
description told callers to write `organism_id:`, so those clauses are rewritten
too, and the widening is disclosed rather than applied silently.

**DGIdb dropped gene symbols it could not resolve.** All four DGIdb tools removed
unmatched names with no trace while reporting success; a single unrecognised
symbol was byte-indistinguishable from a real gene with no recorded interactions.
DGIdb's GraphQL API returns HTTP 200 with no `errors` key and simply omits what it
cannot resolve, so the omission has to be detected by comparing request to
response. Probing established that genes match exactly and drugs match by prefix,
so `erlotinib` legitimately returning `ERLOTINIB HYDROCHLORIDE` is not flagged.

**CTD threw away the records it had just fetched.** `CTD_get_chemical_diseases`
returned success with zero rows for "chromium, hexavalent". mydisease.info matched
55 documents for that phrase; the tool discarded all of them because its
client-side post-filter demanded full-string equality while the query itself was
an Elasticsearch phrase match, and CTD stores the chemical as "chromium hexavalent
ion". Strict equality remains the preferred match — "chromium" still returns its
own 56 records rather than widening — but when it leaves nothing and the backend
did return documents, the canonical values present in those documents are read off
the response already held. "chromate" had the same failure and now names the four
distinct chromate salts it matched.

**ChEBI answered about a different compound than the one requested.**
`ChEBI_get_compound` with id 1 returned CHEBI:18357, (R)-noradrenaline, labelled
success, with nothing indicating a substitution. ChEBI merges obsolete accessions
into a primary entry and serves the primary — correct upstream behaviour — and the
response carries `secondary_ids` listing CHEBI:1 among them. The tool dropped that
field. The disclosure is built from it, so the claim made is the one the response
supports; where the requested accession is absent from `secondary_ids`, or the
endpoint omits them, the note says exactly that instead.

**HGNC reported no such gene for symbols it had itself retired.**
`HGNC_fetch_gene_by_symbol` answered "CYP2E" with an empty record, contradicting
HGNC's own CYP2E1 entry which carries "CYP2E" in `prev_symbol`. `fetch/symbol` only
matches current approved symbols; HGNC's fetch endpoint works on any searchable
field, and the record returned via `prev_symbol` is byte-identical to a direct hit.
A current-symbol hit still resolves first and always wins; a retired symbol shared
by several genes (alias "p65" — RELA, SYT1, GORASP1) names all of them.

**Semantic Scholar reported one page as the whole literature.** `metadata.total`
was `len(papers)`, so it always equalled the caller's limit — 10, 5 and 3 for the
same query. The corpus match count arrives on every response and was discarded.

**GtoPdb accepted a ligand type filter it never applied.** Searching semaglutide
with type "NotARealType", "Synthetic", or nothing at all returned identical
results. GtoPdb drops `?type=` whenever `?name=` is present, confirmed directly for
both the ligands and targets endpoints. The published vocabulary was wrong in both
directions: three of six documented target values 404 upstream, while
`?type=Peptide` 404s despite 2452 records carrying that type. Filtering now runs
against each record's own type field, and unrecognised values are rejected before
any request rather than quietly returning everything. The old test example
(dopamine, "Endogenous peptide") asserted nothing — dopamine is not an endogenous
peptide, and it passed only because the filter was ignored.

## Contract changes needing human review

- **`SemanticScholar_search_papers` — breaking.** `metadata.total` is re-meaned
  from page size to corpus match count, and may now be `null`. Callers reading it
  as "how many rows did I get" must read `metadata.returned`.
- **`GtoPdb_search_ligands` / `GtoPdb_search_targets` — behaviour change.** Name +
  valid type now returns a subset (dopamine+Metabolite 4→2; serotonin+GPCR 19→13).
  An unrecognised type is now an error instead of unfiltered success — including
  "Endogenous peptide", which was in our own published enum and test example.
  `type` or `approved` used alone returns different rows (Peptide 0→2452;
  approved=true 13856→2218).
- **`CTD_get_chemical_diseases` — status flip.** Two states that were success with
  an empty array are now errors carrying the candidate list. Anything treating
  `data: []` as "CTD has nothing" will see an error instead.
- **`UniProt_search` — default change.** Numeric taxon scoping now matches the
  lineage. Result sets are supersets, never smaller; exact-node semantics are no
  longer reachable through this tool.
- **Additive optional keys** (absent on the unchanged happy path, so direct hits
  stay byte-identical): `data.normalization_note` (UniProt);
  `metadata.genes_requested`/`unmatched_genes`/`drugs_requested`/`unmatched_drugs`
  (DGIdb); `metadata.resolved_chemical_name`/`normalization_note`/`candidates`
  (CTD); `data.requested_chebi_id`/`redirect_note`/`secondary_ids` (ChEBI);
  `metadata.resolved_from`/`resolved_symbol`/`symbol_resolution_note`/
  `no_results_note` (HGNC); `type_filter`/`approved_filter` (GtoPdb). No
  `return_schema` in scope set `additionalProperties: false`; schemas were updated
  where they describe the payload.

## Verification

`tests/unit` + `tests/integration` (the CI-gated suites) pass with zero failures,
against a baseline captured before any change that was also clean. Seven new
regression tests, all mocked, no network: 20 UniProt, 13 DGIdb, 8 CTD, 14 ChEBI,
12 HGNC, 26 Semantic Scholar, 31 GtoPdb. Every fixed tool re-run through the CLI
with its original failing arguments, and `tu test` re-run across the affected tool
families.

`src/tooluniverse/tools/` was never regenerated and is unmodified.

**Not fixed, deliberately.** STITCH's `/json/interactions` and `/json/network`
endpoints genuinely 404 upstream; the tool already returns an honest error naming
the dead endpoint and pointing at the working sibling, which a prior round added.
Semantic Scholar 429s were reproduced by direct curl to the upstream API and are
rate limiting, not a defect. The post-fix Semantic Scholar envelope could not be
observed live for the same reason and is verified only through mocked tests.


gasvn:

PR #505, opened 2026-08-10
Fix Round 31: numbers that were confidently wrong, and tools that blamed the caller

gasvn:
Round 31 explored three domains untouched by rounds 20-30 — tropical/neglected disease, dermatology/cutaneous genetics, and sports/exercise physiology — plus an orchestrator probe of population genetics and identifier cross-referencing. 46 issues were reported across the three personas; 13 survived CLI verification, sibling-PR deduplication and root-cause confirmation at the upstream API.

Every fix below was reproduced by the orchestrator with `tu run` before and after, and every root cause was established against the upstream service (direct `curl`, upstream OpenAPI documents, or upstream source), not inferred from tool behaviour.

## The wrong-number fixes

**DrugCentral reported a −log10 potency under an `IC50` label.** `DrugCentral_get_targets '{"chem_id":"CHEMBL976"}'` returned `activity_type: "IC50", activity_value: "4.01"` for praziquantel/BSEP. ChEMBL's record for the same pair is `IC50 96800 nM, pChEMBL 4.01` — so `4.01` is the p-scale value and the tool was asserting a concentration roughly five orders of magnitude off. 143 sampled live rows all fell in 4.05–12.301, confirming the scale is universal. `activity_value` and `activity_type` are unchanged; new sibling keys (`activity_type_reported`, `activity_value_scale`, `activity_value_nM`) make the scale explicit.

**SIDER invented a placebo rate for every side effect.** `placebo_frequency` equalled the drug's own frequency on 43/43 dapsone rows and 233/233 acitretin rows — every side effect reading as occurring at exactly the placebo rate. SIDER's placebo cell is empty for those labels; the parser scanned the whole table row for percentages and took the second match, which is a per-label tooltip echoing the drug's own rate. Values are now read per cell; an empty placebo cell yields `null` (already permitted by the schema). Two further row-wide-regex defects fell out: a qualitative frequency (`frequent`) overwritten by a percentage found inside the row's MedDRA prose, and ranges written `1-10%` truncated to their upper bound.

**A missing-value sentinel outranked real genome-wide hits.** `GWASSumStats_get_region_associations` at `p_upper=5e-8` returned ten rows, eight of them `p_value: -99.0`, ranked above the two genuine hits (5.31e-24, 5.71e-23). `-99` is the GWAS Catalog missing-value sentinel and `-99 <= 5e-8` is numerically true; our own ascending sort then lifted every sentinel above every real result. Rows are dropped only when a threshold was requested, and the excluded count is disclosed.

**HPA labelled single-cell values with the wrong unit** (`nTPM` where HPA reports `nCPM`), and `tissues_with_expression` reported 184 against HPA's ~44 consensus tissues — it was a row count including 135 rows with no level, 9 explicitly "not detected", and per-antibody duplicates.

## The tools that blamed the caller

**HPA blamed the caller for column codes HPA never had.** Asking for skin expression returned "'skin' did not match an HPA tissue column (t_RNA_skin). This means the tissue name is unrecognized" — but HPA's column is `t_RNA_skin_1`, and HPA silently omits unknown codes, so a wrong code and a missing value look identical. The same defect hid behind `stomach`, `endometrium` and `brain`; four further names map to no RNA column at all. It was masked for skin-enriched genes, which fell through to the tissue-specificity summary column and answered plausibly — so the tool worked for exactly the genes you would already have guessed. `source_type` values `single_cell` and `blood` were dead for every gene: `rnascm`/`rnablm` are not HPA column codes.

**GWAS blamed the trait term for its own filter's zero result.** BMI associations at `p<5e-8` returned an empty list plus advice that the EFO term might be wrong — while the same payload reported 23,608 total elements for that term. The `p_value` filter is client-side over one page, and with `direction=desc` page 0 holds the least significant rows. The v2 OpenAPI document declares no p-value parameter and 15 candidate names were probed, so the filter must stay client-side; the tool now fetches the significant end and attributes an empty result to the filter.

**Orphanet called a valid disease code nonexistent.** `Orphanet_get_phenotypes '{"orpha_code":"1247"}'` answered "Disease not found. Use Orphanet_search_diseases to find a valid ORPHA code" — a false statement whose advice loops. Four sibling tools resolve that code. The `rd-phenotypes` 404 body is byte-identical for a valid-but-unannotated code and a made-up one, so the RDcode Name endpoint is now used to discriminate.

**MedlinePlus threw away documents it had already fetched.** Two of three documented `rettype` values fetched HTTP 200 with valid documents and then reported "Failed to parse health topic information". Every call also printed ten lines of diagnostics plus a raw structure dump to stdout ahead of the payload.

**InterPro reported a zero-hit search as an internal error.** The EBI API answers a zero-hit query with HTTP 204 and an empty body; `raise_for_status()` treats 204 as success, so `.json()` ran on an empty string. A caller could not distinguish "no such family" from "the tool is broken".

## The answers that were unreachable or incomplete

**Monarch could not be asked which genes cause a disease.** The implementation already read an `object` argument, but the registered schema exposed only `subject` and marked it required, so the disease-anchored direction of a directed Biolink category was unreachable. Upstream answers with three genes for MONDO:0006559.

**WoRMS search could not reach the records its siblings serve.** `WoRMS_search_species` returned a confident empty success for *Schistosoma mansoni* while `WoRMS_get_classification` and `WoRMS_get_distribution` both serve AphiaID 1599676 — and their docs say to obtain that ID from the search tool. An optional `marine_only` parameter is added rather than the default flipped.

**VEP dropped the allele each consequence belongs to.** For rs1815739 (ACTN3 R577X) the tool emitted two rows for the same transcript and protein position with opposite verdicts and nothing to tell them apart. Which allele is the null one is the entire clinical content of R577X.

**UniChem discarded the InChIKey it was already given.** The `/compounds` response carries `standardInchiKey` at the top level, but the field only ever echoed the caller's own query, so the common direction — "I have CHEMBL25, what is its InChIKey?" — always returned null.

**TogoID answered an unroutable conversion with a URL dump.** TogoID reports the actionable reason with HTTP 400, and the tool already had a branch written to turn that message into guidance — but it sat behind `raise_for_status()` and could never run.

## Contract changes flagged for review

Most of this round is additive. Six changes alter what a caller receives, and every one replaces a value that was wrong:

1. **SIDER `placebo_frequency` → `null`** for most rows, and `frequency` changes shape for qualitative/range labels. An additive disclosure was rejected: the old number was synthesised, not published, and the specific claim it made inverts the clinical conclusion.
2. **GWASSumStats row set** shrinks when `p_upper` is set (10 → 2 for the repro). Additive disclosure alone was rejected because `-99` was passing a significance filter and ranking first.
3. **InterPro zero-hit: error → success.** The safe direction; no caller could depend on an unconditional error.
4. **HPA `tissues_with_expression` re-meant** (184 → 40). This is the one key re-meaning in the round. The raw count is preserved additively as `tissue_expression_rows`; 184 was impossible under the name it carried.
5. **HPA tissue/source values** appear where `"N/A"` was returned, and single-cell `expression_unit` changes `nTPM` → `nCPM`.
6. **GWAS fetch direction** is overridden when a p-value threshold is present; rows are returned in the caller's requested direction and the override is disclosed in metadata.

MedlinePlus's stdout removal is a behaviour change in the benign direction — the diagnostics move to `logger.debug`, recoverable via `TOOLUNIVERSE_LOG_LEVEL=DEBUG`.

## Not touched

Files owned by open PRs #500-#504 were treated as a do-not-touch list, fetched per-PR at verification time. Six verified findings were skipped as sibling-PR duplicates or declined on conflict risk, and the InterPro guard was kept as a small per-module helper rather than moved into `base_rest_tool.py`, which #503 owns. The tracked wrapper tree under `src/tooluniverse/tools/` was not regenerated.

### DIFF
--- a/src/tooluniverse/data/drugcentral_tools.json
+++ b/src/tooluniverse/data/drugcentral_tools.json
@@ -54,16 +54,37 @@
             "description": "SMILES chemical structure"
           },
           "approval": {
-            "type": ["array", "null"],
-            "description": "Regulatory approval records",
-            "items": {
-              "type": "object",
-              "properties": {
-                "agency": {"type": ["string", "null"]},
-                "company": {"type": ["string", "null"]},
-                "date": {"type": ["string", "null"]}
-              }
-            }
+            "description": "Regulatory approval record(s), passed through from DrugCentral unchanged. Upstream returns a single object when the drug has exactly one approval record (e.g. praziquantel) and an array when it has several (e.g. metformin: FDA, EMA, PMDA), so both shapes occur.",
+            "oneOf": [
+              {
+                "type": "object",
+                "properties": {
+                  "agency": {"type": ["string", "null"]},
+                  "company": {"type": ["string", "null"]},
+                  "date": {"type": ["string", "null"]},
+                  "orphan": {
+                    "type": ["string", "null"],
+                    "description": "Orphan-drug designation flag, reported upstream as the string 'True' or 'False'."
+                  }
+                }
+              },
+              {
+                "type": "array",
+                "items": {
+                  "type": "object",
+                  "properties": {
+                    "agency": {"type": ["string", "null"]},
+                    "company": {"type": ["string", "null"]},
+                    "date": {"type": ["string", "null"]},
+                    "orphan": {
+                      "type": ["string", "null"],
+                      "description": "Orphan-drug designation flag, reported upstream as the string 'True' or 'False'."
+                    }
+                  }
+                }
+              },
+              {"type": "null"}
+            ]
           },
           "drugbank_id": {
             "type": ["string", "null"],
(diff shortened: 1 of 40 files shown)


gasvn:

PR #506, opened 2026-08-10
Fix Round 32: numbers and emptiness that a reader would have acted on

gasvn:
Six verified defects where a tool returned a number or an emptiness that a
reader would act on, and that was wrong. Each was reproduced through the CLI
and traced to its source — five of the six against the live upstream API —
before any code was changed.

This round was deliberately kept **additive**: a new response key, a new
accepted input, a restored field. Exactly one contract change is included and
is called out below, because the additive alternative would have left callers
holding a number that inverts the conclusion.

---

### 1. InterPro member-database listing reported a valid zero-result as an error

`InterPro_list_member_signatures` with `member_database=antifam, entry_type=domain`
failed with `InterPro API HTTP error: 204`. AntiFam has no signatures of type
`domain` — they are all families — so zero is the correct answer.

Established upstream: `GET /entry/antifam/?type=domain` returns **204**, while
`GET /entry/antifam/` returns **200** with results; `GET /entry/sfld/?type=repeat`
also returns 204. The EBI API uses 204 to mean "valid query, zero matches", and
the shared `_get()` mapped every non-200/non-404 to an error.

List endpoints now read 204 as an empty result set, keeping `total_count: 0`,
`returned: 0`, `entries: []` and adding a `note` that echoes the filter that
narrowed it. Detail endpoints still fail loudly on 204, so an empty body can
never become a record of all-null fields. `rcsb_search`, `worms`, `bridgedb`
and `pfam` already follow this convention.

**Additive.**

---

### 2. Sex-dependent clinical calculators silently assumed male

`ClinicalCalc_eGFR_CKD_EPI`, `ClinicalCalc_CHA2DS2_VASc` and
`ClinicalCalc_ASCVD_risk` accept sex as a `female` boolean but report sex back
in their output. A caller passing the natural `sex: "female"` had it dropped —
the configs set no `additionalProperties`, so unknown keys pass validation —
fell through to `female=False`, and received a male result carrying a field
that looked like it echoed their input:

```
run ClinicalCalc_eGFR_CKD_EPI '{"creatinine":1.2,"age":55,"sex":"female"}'
before: score 71.4, "CKD stage G2",  components.sex "male"
after:  score 53.5, "CKD stage G3a", components.sex "female"
```

A 33% overstatement across a CKD stage boundary. CHA2DS2-VASc lost the sex
point (3 → 2 for a 68-year-old woman with hypertension); ASCVD selected the
wrong Pooled Cohort coefficient set (`black_female` → `black_male`).

`sex` is now a documented parameter on all three, resolved by one shared
helper. The `female` boolean is untouched and the no-sex default stays male,
so no existing call changes its answer. A conflicting pair, or an
unrecognised sex string, is now rejected explicitly instead of quietly meaning
male. Where a result does rest on an unstated default — no sex anywhere, or no
race on ASCVD — the response carries an `assumptions` entry saying so.

CKD-EPI 2021 coefficients were checked against NIDDK's published equation and
are unchanged.

**Additive** (new input key; existing key, default and outputs preserved).

> Note for maintainers: the generated convenience wrappers in
> `src/tooluniverse/tools/ClinicalCalc_*.py` have a fixed signature and will not
> expose `sex` as a keyword argument until that tree is regenerated. The CLI,
> MCP and JSON-schema paths already accept it. The tree was intentionally not
> regenerated here.

---

### 3. ClinGen allele coordinates came back without their assembly

`ClinGenAllele_lookup_hgvs` returns one row per genome build, so the same
variant appears three or four times with different coordinates — for
rs1426654: 48134287 (GRCh38), 48426484 (GRCh37), 46213775 (NCBI36). Every row
reported `"reference_genome": null`.

Established upstream: the registry does supply the label, on the parent
`genomicAlleles[i]` object. The shared formatter read it from the inner
`coordinates[j]` dict, which contains only `allele`, `start`, `end`,
`referenceAllele`. Reading it from the parent populates the field for both
operations. The registry's RefSeqGene row genuinely has no build and still
reports `null` rather than a guess.

**Additive** (restores a field that was always null; row count unchanged).

---

### 4. PopGen_fst returned a number driven by sample size — the one contract change

```
run PopGen_fst '{"operation":"fst","p1":0.4,"p2":0.45,"n1":50,"n2":50}'
  before: Fst 0.0    "Little genetic differentiation"
run PopGen_fst '{"operation":"fst","p1":0.4,"p2":0.45,"n1":10000,"n2":10000}'
  before: Fst 0.9617 "Very great genetic differentiation"
```

Identical allele frequencies, opposite conclusions. At p=0.2 vs 0.8 with n=1e6
it saturated at exactly 1.0000, where the true value is about 0.53.

The among-population mean square was correct; the denominator was not. Weir &
Cockerham (1984) theta is `(MSP - MSG) / (MSP + (n_c - 1) * MSG)` with
`n_c = (n1 + n2) - (n1² + n2²)/(n1 + n2)` for two populations. The code used
`MSP + MSG`, under-weighting MSG by a factor of about n. After: 0.005 at both
n=50 and n=10000, and 0.5294 for the p=0.2/0.8 case — matching an independent
hand-derivation and cross-checking against Nei's Gst for order of magnitude.

The value is corrected rather than merely flagged because an ancestry or
forensic comparison of 1000 Genomes or gnomAD superpopulations, where sample
sizes run to the thousands, would otherwise read "very great differentiation"
at essentially every locus. An `estimator` key now names the estimator and its
correction. Interpretation thresholds and all other keys are unchanged.

**Contract change** — the returned `Fst` value changes for a given input. This
is the round's single budgeted contract change.

---

### 5. Expression Atlas answered "no data" from 0.3% of the experiment

`GxA_get_experiment_expression` reported `total_gene_profiles: 0` for TP53 in
E-MTAB-2836 — the 122-tissue human baseline experiment, and the tool's own
documented example.

An earlier round had already found that the endpoint ignores gene filters and
added a client-side filter, but missed why that is not enough: the endpoint
returns only a small default sample. Measured live, **29 rows out of a
`searchResultTotal` of 9570**, so the client-side filter searches 0.3% of the
data.

A controlled negative probe confirms the filter is genuinely unavailable
rather than mis-spelled by us — no params, `geneQuery=TP53`,
`geneQuery=[{"value":"TP53"}]`, `geneQuery=[{"value":"ENSG00000141510"}]` and
`&specific=true` all return the identical 29 rows; the `/genes` subpath returns
HTTP 400. No working per-gene endpoint was found.

Since the right answer cannot be obtained here, the response now says so.
Existing keys keep their names and meanings; three siblings are added:
`profiles_returned_by_upstream`, `profiles_available_upstream`, and a
`coverage_warning` that fires only when a gene filter matched nothing, stating
that an empty result is not evidence the gene is unexpressed. The counts also
stop `total_gene_profiles` from being mistaken for the experiment's gene count.
`searchResultTotal` arrives as a string upstream and is cast defensively.

**Additive.**

---

### 6. The baseline listing still steered callers into that broken check

`ExpressionAtlas_get_baseline`'s response `note` told the caller to "use
`GxA_get_experiment_expression` ... and `gene_id=<gene>` to check whether that
specific experiment has data for the gene" — the check defect 5 shows cannot
work. Updating the JSON description was not enough: this sentence is built in
Python and shipped in the response body, which is what the caller reads.

The note now states the listing cannot tell you whether the gene was measured,
warns against the `gene_id` check and why, and points to
`GTEx_get_expression_summary` and `HPA_get_rna_expression_in_specific_tissues`,
both of which were run and confirmed to return real per-gene values. A nearby
comment claiming the Atlas "does support per-gene filtering" is corrected.

The honesty test previously asserted only that the note *mentioned*
`GxA_get_experiment_expression`, which pinned the unsafe wording and would have
passed for any framing. It now checks that the note still discloses the tool
does not filter by gene, and that any mention of the other tool appears as a
warning rather than a recommendation.

**Additive** (text only; counts, ordering and all keys unchanged).

---

## Verification

- `tests/unit` — full suite green, no failures, same skips as the pre-change
  baseline. This is the CI gate.
- 29 new unit tests across six files covering each fix and its
  non-regression counterpart.
- Every fix re-run through the CLI with the original failing arguments.
- `tests/tools` failures (SemanticScholar, ClinVar, GTEx, IntAct) were
  reproduced identically against source *without* these changes, confirming
  they are inherited from `main` rather than introduced here.
- `src/tooluniverse/tools/` untouched; the generated tree was not regenerated.

## Deliberately not included

Two further defects in `popgen_tool.py` were verified but left alone, because
each would change every returned value and this round's contract-change budget
is spent on `PopGen_fst`:

- `PopGen_inbreeding` uses `F_t = base_F + (1-base_F)·F_{t-1}` for every
  pedigree. Continued full-sib mating needs Wright's second-order recurrence
  `F_t = ¼(1 + 2F_{t-1} + F_{t-2})`. Observed `[0.25, 0.4375, 0.578125]`
  against the textbook `[0.25, 0.375, 0.5]`. The tool does document the
  recurrence it uses, so it is transparent about the approximation.
- `BindingDB`'s `getLigandsByUniprots` route returns `affinities: []` for both
  P35372 and P00533; the emptiness is upstream, and distinguishing a dead route
  from a genuinely empty one would flip a success to a warning.

### DIFF
--- a/src/tooluniverse/clingen_allele_tool.py
+++ b/src/tooluniverse/clingen_allele_tool.py
@@ -103,6 +103,9 @@ class ClinGenAlleleTool(BaseTool):
         # Extract genomic coordinates
         genomic = []
         for ga in data.get("genomicAlleles") or []:
+            # `referenceGenome` (and `chromosome`) live on the parent
+            # genomicAlleles[i] object, not on the inner coordinates[j]
+            # dict -- the inner dict only has start/end/allele/referenceAllele.
             for coord in ga.get("coordinates") or []:
                 genomic.append(
                     {
@@ -111,7 +114,7 @@ class ClinGenAlleleTool(BaseTool):
                         "end": coord.get("end"),
                         "allele": coord.get("allele"),
                         "reference_allele": coord.get("referenceAllele"),
-                        "reference_genome": coord.get("referenceGenome"),
+                        "reference_genome": ga.get("referenceGenome"),
                     }
                 )
 
--- a/src/tooluniverse/clinical_calculators_tool.py
+++ b/src/tooluniverse/clinical_calculators_tool.py
@@ -37,6 +37,68 @@ def _req_number(args: Dict[str, Any], key: str) -> float:
         raise ValueError(f"'{key}' must be a number, got {val!r}")
 
 
+# Accepted string values for the `sex` parameter (case-insensitive).
+_SEX_STRING_MAP = {"female": True, "f": True, "male": False, "m": False}
+
+
+def _resolve_sex(a: Dict[str, Any]) -> "tuple[bool, Any]":
+    """Resolve biological sex from the `female` boolean and/or `sex` string args.
+
+    Both are accepted for backward compatibility: `female` is the original
+    parameter, `sex` is the natural-language clinical term this tool's own
+    output already speaks in terms of ("components": {"sex": ...}). Accepting
+    it as input too makes the interface symmetric.
+
+    Returns (is_female, assumption_note). `assumption_note` is None unless
+    neither `female` nor `sex` was supplied, in which case the *existing*
+    default (male) is still used, but a note is returned so the caller can
+    disclose that the default was silently assumed.
+
+    Raises ValueError if:
+      - `sex` is supplied but is not a recognised value (female/f/male/m,
+        case-insensitive), or
+      - both `female` and `sex` are supplied and they disagree.
+    """
+    has_female = "female" in a and a.get("female") is not None
+    has_sex = "sex" in a and a.get("sex") is not None
+
+    sex_is_female = None
+    if has_sex:
+        raw = str(a["sex"]).strip().lower()
+        if raw not in _SEX_STRING_MAP:
+            raise ValueError(
+                "'sex' must be one of "
+                f"{sorted(_SEX_STRING_MAP)} (case-insensitive), got {a['sex']!r}"
+            )
+        sex_is_female = _SEX_STRING_MAP[raw]
+
+    female_is_female = _truthy(a["female"]) if has_female else None
+
+    if has_sex and has_female:
+        if sex_is_female != female_is_female:
+            raise ValueError(
+                "conflicting sex inputs: "
+                f"'female'={a['female']!r} implies "
+                f"{'female' if female_is_female else 'male'}, but "
+                f"'sex'={a['sex']!r} implies "
+                f"{'female' if sex_is_female else 'male'}. "
+                "Provide consistent values (or only one of the two)."
+            )
+        return female_is_female, None
+
+    if has_sex:
+        return sex_is_female, None
+    if has_female:
+        return female_is_female, None
+
+    return False, (
+        "Neither 'female' nor 'sex' was supplied; male coefficients/points "
+        "were assumed by default. This choice materially changes the "
+        "result — supply 'sex' (or 'female') explicitly for an accurate "
+        "calculation."
+    )
+
+
 def _ok(score, interpretation, components, **extra) -> Dict[str, Any]:
     data = {"score": score, "interpretation": interpretation, "components": components}
     data.update(extra)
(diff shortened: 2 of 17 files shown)


gasvn:

PR #508, opened 2026-08-10
Fix Round 33: tools that answered without the data they were asked for

gasvn:
Six verified defects, each one a tool that answered confidently while dropping, ignoring, or never requesting the thing the caller asked for. Every issue was reproduced from the CLI and, where the cause was upstream, established with a controlled probe rather than inferred.

## BindingDB returned nothing for every target

`BindingDB_get_ligands_by_uniprot` reported `status: success` with `affinities: []` for HCRTR1, HCRTR2 and EGFR. The cause was the query-parameter name: BindingDB's REST spec documents singular `uniprot=`/`pdb=` even on the plural endpoints, and multiple ids join with a comma, not a semicolon. The endpoints do not reject the wrong spelling — `uniprots=` answers HTTP 200 with an empty list and `pdbs=` answers HTTP 500.

Controlled probe, identical URLs except the parameter name:

| request | result |
|---|---|
| `?uniprots=O43613` | HTTP 200, `affinities: []` |
| `?uniprot=O43613` | HTTP 200, 10,460 records |
| `?uniprot=O43613;O43614` | 0 records |
| `?uniprot=O43613,O43614` | 3,229 records at cutoff=10 |
| `?pdbs=4S0V` | HTTP 500 |
| `?pdb=4S0V` | HTTP 200 with data |

After: O43613 → 10,460, O43614 → 12,846, P00533 (EGFR) → 29,169.

The wrong parameter had also produced a false diagnosis that outlived it. The module docstring, an inline comment, a user-facing error and `broken_apis/bindingdb_rest.json` all asserted the REST tier was down, or that the PDB route was "failing for all structures", and routed callers to substitute databases for an outage that does not exist. Only the singular `getLigandsByUniprot` genuinely hangs. `getTargetByCompound` additionally namespaces its payload with a `bdb.` key prefix the parser did not recognise; normalising it also preserves `bdb.species`, which matters because many hits are non-human orthologs.

## Reactome accepted species/type filters it could never apply

`Reactome_query_by_ids` advertised `species` and `types` as optional filters. Neither was ever transmitted, and forwarding them would not have helped: a controlled probe returned a byte-identical 32,445-byte payload for every variant. The decisive negatives are `species=Mus musculus`, which still returned the human record, and `types=Reaction`, which still returned two Pathways. The endpoint does no server-side filtering, so both are rejected at input with a pointer to what does work.

## Orange Book reported empty searches as a dead endpoint

openFDA answers a well-formed query that matches nothing with HTTP 404 plus `{"error":{"code":"NOT_FOUND"}}`. With no branch for that, every empty search surfaced as a raw client error with the upstream URL leaked. Separately `drug_name` was an undeclared alias for the active-ingredient field, so a brand name passed to the most natural-sounding parameter could never match: `drug_name="Quviviq"` failed with a 404 while `brand_name="QUVIVIQ"` returned NDA214985 from the same index.

While fixing that, the worker found the multi-term join was itself broken: `"+AND+"` written into a `params` dict encodes to `%2BAND%2B`, which openFDA cannot parse, so *every* multi-term search was already failing. Adding the 404-as-empty branch without this would have converted those failures into confident, successful-looking "zero results" — strictly worse than the original defect.

## CTIS accepted a status filter it ignored

`status` admitted any string, so an invalid value reached the API unchecked and returned `status: success` with zero trials, echoing the bad value back in `search_criteria` as though honoured. The trap was sharpened by the tool's own description, which names statuses in words, so the most natural guess produced a confident "there are no such trials". Verified against the same query with a valid code: the garbage value returned 0 records where `status=[4]` returns 5. Status values are now validated at input. The config's legend was also wrong — it said `4=ended, 5=expired`; CTIS reports `8=Ended, 9=Expired`.

## Pharos omitted the diseases and ligands it promised

`Pharos_get_target`'s description promises "diseases, ligands, and druggability data". The GraphQL selection hard-coded eight scalar fields and requested none of them, so target-to-disease was unreachable from any Pharos tool. The data was one field name away: FOLH1 returns 240 ligands, 1 drug and ~23 disease associations led by prostate cancer at 1195. Schema introspection confirmed there is no `druggability` field — TDL *is* Pharos' druggability classification — so that promise was reworded rather than faked. The `return_schema` also declared four fields the tool has never returned, and is now accurate.

`Pharos_get_ligand_targets` reports pActivity (−log10 molar), not nanomolar, and said nothing about it. A Ki printed as 9.46 is pKi 9.46 = 0.35 nM, which cross-checks exactly against the suvorexant label's stated OX2R affinity; a reader taking it as nanomolar is wrong by ten orders of magnitude. The sibling tool already documents the convention, so its wording is carried across. Numbers are unchanged.

## DNA translation ignored the genetic code

`DNA_translate_reading_frames` declared a `genetic_code` parameter with no enum and never read it. `genetic_code=2` (vertebrate mitochondrial) returned the standard-code translation, so `ATGTGATTAAGATAA` came back as `M*LR*` where code 2 gives `MWL**` — a spurious stop at codon 2 and a missing stop at codon 4, exactly the codons that distinguish the two tables. Only the standard code is implemented, so anything else is now rejected. The constraint is declared in the schema and re-checked in the handler, so callers that construct the tool class directly get the same actionable error.

## Verification

All six reproduced and re-verified from the CLI before and after. 190 targeted tests pass across the affected areas, including 27 new offline regression tests with the transport mocked.

A full `tests/unit` run on the rebased branch is **4,397 passed, 2 failed, 28 skipped**. The two failures are `test_discovered_bugs.py::test_missing_api_key_returns_dict` (an ArXiv tool returning a list where the test wants a dict) and `test_documentation_core.py::test_tool_finder_llm_execution` (needs an API key). Both were confirmed pre-existing by stashing this branch's work, checking the affected paths out from `origin/main`, and re-running: they fail identically there, so neither is a regression from these changes.

The generator was never run; `src/tooluniverse/tools/` contains only two hand-edited wrappers for the tools whose schemas changed.

No contract change is spent here. Every change is either input validation that rejects a value which previously produced a wrong answer, or the surfacing of data the tool already promised. Callers passing valid arguments see no behaviour change.


gasvn:
Verified this independently before merge, and the upstream probe turned up something worth recording: **the wrong parameter name was not merely returning nothing — upstream was answering with a different protein.**

Same endpoint, same UniProt ID, only the parameter name differs:

```
?uniprots=O43613  ->  119 records,  "bdb.primary": "P35355"   ← not the protein asked for
?uniprot=O43613   -> 3943 records,  "bdb.primary": "O43613"   ← correct
```

So `uniprots=` is not being parsed as an empty-or-invalid filter; it is ignored, and BindingDB returns some other target's binding data with `bdb.primary` naming a protein the caller never asked about.

That did not surface as a wrong answer only because of a second, unrelated mismatch: BindingDB keys the payload `getLindsByUniprotResponse` (singular, and misspelled upstream), while the parser was looking for the plural form — so it found nothing and returned `affinities: []`. Two defects were cancelling out. Had the response key ever matched, the tool would have served P35355 affinity data as if it were O43613 — a confidently wrong answer rather than an obviously empty one.

Tool-level before/after on this branch, confirming the fix:

```
BindingDB_get_ligands_by_uniprot {"uniprot_id":"O43613"}
  merged main : status success, affinities []
  this branch : status success, affinities 10460 records
```

Worth noting because it changes how the "REST tier is down" claim in `broken_apis/bindingdb_rest.json` should be read — the endpoint was never down, and an empty success was the only thing standing between us and mis-attributed pharmacology.

PR #509, opened 2026-08-10
feat: add native remote tool sharing with standalone compatibility

gasvn:
## Summary

- restore first-class native remote-tool authoring and provider serving in ToolUniverse
- add resilient platform relay jobs while keeping the platform companion optional
- preserve standalone SDK behavior: no platform runtime or FAISS import on the ordinary local path
- make saved remote MCP connections recoverable when a provider is temporarily offline
- make Profile reloads deterministic and keep liveness checks independent from expensive SDK startup
- add built-wheel compatibility coverage for Python 3.10 through 3.14

## User flows covered

- a provider decorates a local function with `@remote_tool`, launches it on their own CPU/GPU server, and exposes it over MCP
- a consumer saves the provider URL with `tu connect`, discovers proxy tools, and invokes them through an ordinary ToolUniverse instance
- a stopped provider leaves the saved connection intact; restarting it restores calls without reconnecting manually
- users who never install or select TU Platform continue using the local SDK and generated tool functions normally

## Validation

- full unit and integration suite: passed
- focused HTTP liveness/concurrency, Profile reload, MCP stdout/recovery, and standalone-isolation tests: passed
- built and installed the candidate wheel on Python 3.10, 3.11, 3.12, 3.13, and 3.14
- dependency consistency checks passed on all supported Python versions; Python 3.14 excludes only the upstream-incompatible YouTube converter extra
- upgraded a clean environment from the published 1.4.0 wheel to the candidate wheel while preserving its workspace Profile
- final-wheel standalone smoke executed `ToolUniverse_get_usage_tips` without importing the platform companion or FAISS
- final-wheel `/health` returned immediately without constructing or loading a ToolUniverse instance
- real local provider/consumer MCP test passed discovery, tool execution, disconnect, automatic reconnect, and 32 concurrent calls (32 successful, 0 errors)

## Release guard

This is intentionally a **draft**. Merging or publishing a ToolUniverse release is out of scope until review and CI are complete. The TU Platform integration remains isolated so this branch can be merged without requiring standalone users to migrate to the platform.


gasvn:
Closing this in favor of #546 (already merged), which is the landed version of this same feature.

Verified before closing: the three core modules this PR introduces — `mcp_tool_registry.py`, `remote_connections.py`, `platform_remote_tool.py` — are byte-for-byte identical between this branch and current `main`. Every file this PR touches has also been touched on `main` since their common ancestor, and the only test difference between the two branches is a test that exists *only* on `main` (additive coverage `main` has that this branch doesn't). No capability, fix, or edge case unique to this branch was found.

PR #517, opened 2026-08-11
Fix Round 41: drug-name lookups answering with a different drug's label

gasvn:
## The defect

`FDA_get_boxed_warning_info_by_drug_name {"drug_name":"Albuminex"}` returned **BEIZRAY / docetaxel's** boxed warning — toxic deaths, hepatotoxicity, neutropenia, hypersensitivity — presented as the boxed warning of a plasma-derived human albumin product. A clinician asking about an albumin infusion was served a chemotherapy agent's warnings.

## Root cause, measured at the upstream API

All 77 `FDA_*_by_drug_name` tools map `drug_name` onto the OR-group `(openfda.brand_name, openfda.generic_name, spl_product_data_elements)`. The third is the raw product/ingredient blob, so the query matched any product merely *containing* albumin. AND-ing `_exists_:boxed_warning` then deleted both genuine Albuminex labels — which correctly have no boxed warning — leaving only the wrong drug's.

Measured at `api.fda.gov/drug/label.json`:

| query | total |
|---|---|
| `spl_product_data_elements:"ALBUMINEX"` | 3 |
| `… AND _exists_:boxed_warning` | **1** (BEIZRAY, docetaxel) |
| `openfda.brand_name:"ALBUMINEX"` | 2 |
| `… AND _exists_:boxed_warning` | **0** (the true answer) |
| `… AND _exists_:warnings_and_cautions` | **2** |

## The fix

Drop a blob-matched row when the record's **own resolved openFDA names** exist and none contains the queried name — that is an excipient match, not a product match.

Rows whose `openfda` block is **empty are kept**: they have no resolved identity to contradict, and an unresolved block is exactly how genuine products appear. This is what preserves the earlier denosumab 19→25 recall fix pinned by `test_fda_label_name_match_completeness.py`. The config field lists are unchanged.

Sampled live (100-record samples, match / unresolved / contradicted):

| query | blob | match | unres | dropped |
|---|---|---|---|---|
| Albuminex | 3 | 2 | 0 | **1** (BEIZRAY/DOCETAXEL) |
| denosumab | 25 | 19 | 6 | 0 |
| aspirin | 2223 | 43 | 57 | 0 |
| metformin | 1063 | 44 | 56 | 0 |
| warfarin | 285 | 26 | 74 | 0 |
| docetaxel | 58 | 32 | 26 | 0 |
| polysorbate 80 | 14275 | 0 | 55 | 45 |

Zero drops for ordinary drug-name lookups; drops confined to genuine excipient contamination.

When the filter empties a payload it becomes openFDA's own `NOT_FOUND` shape, so the existing `section_exists_clause` re-probe still engages. Albuminex now answers with its own 2 labels, `boxed_warning: null`, `related_sections_present: ["warnings_and_cautions"]`, and a `section_note` — the honest clinical answer rather than either silence or the wrong drug.

**No extra network round-trip on the success path.**

## On the disclosure note

An earlier revision of this branch emitted a note claiming *"…so this answer describes 'X' itself."* Two independent role-play personas caught that as false: a `minocycline` lookup's surviving rows are all zinc gluconate lozenges, and an `albumin human` lookup's are an allergenic diluent, paclitaxel protein-bound and Dysport. Rows with an unresolved `openfda` block cannot be checked by the guard, so the claim was false precisely where it mattered.

The note now states only what was actually done, and reports how many returned rows **could not** be checked. It is gated on the guard having dropped something, because an unresolved block is the normal shape for 6 of the 25 real denosumab labels — warning on every clean lookup would train readers to skip the note.

## Verification

- `pytest tests/unit -k "fda or openfda or faers"` under the repo's real addopts (`--cov`, `--maxfail=5`, `timeout=60`; **no** `--no-cov`): **276 passed, EXIT=0**. Baseline on this branch was 255 passed — unchanged, plus 21 new tests.
- CLI regression checks return full content with no `note`: denosumab 25 (XBRYK present), warfarin 282, docetaxel 58, aspirin 2188. `FDA_get_drug_names_by_ingredient` is correctly disarmed (blob-only lookup by design).
- New tests are hermetic — the `disable_network` fixture closes `requests.Session.request`, the door both `requests.get` and `request_with_retry`→`requests.request` use, with a positive control proving the guard bites. Includes a negative pin: with the guard disabled, the wrong drug's warning returns.
- `git status --short src/tooluniverse/tools/ src/tooluniverse/_lazy_registry_static.py` is empty; no generator was run.

## Base

Branched from `fix/fitz-dependency` (#516), which is required for CI to install at all — the yanked `fitz` stub blocks dependency resolution repo-wide. **This PR will show #516's commits until that merges, then only its own.**


gasvn:
Follow-up commit `b9e6d5ea`: an adversarial validation pass found four assertions in the disclosure note that the response does not support. The guard's filtering logic is unchanged (validated: zero recall loss across 30+ drug names sampled live; Albuminex still returns its own 2 labels with no docetaxel content). What was wrong was what it *told* the caller.

- **`meta.total` was page-local while the note asserted it had been corrected.** `albumin human` at `limit:10` reported **37** when the true post-filter total is **16** — a 2.3x overstatement — and the page-local caveat lived only in a source comment. The note now says "reduced accordingly, to N" only when the whole result set fits on one page; otherwise it declares the number an **UPPER BOUND** and quotes upstream hits and page size so the reader can audit it. At `limit:100` the same query now reports an exact `total: 16 == result_count: 16`.
- **`dedup_note` contradicted the guard note in the same payload**, still calling `meta.total` "openFDA's upstream hit count before local processing" after the guard had changed it.
- **The note named `spl_product_data_elements` as the sole match route.** On broadened queries that is false — for `minocycline` the term appears in no returned row's blob; those rows matched `indications_and_usage`/`description`. It now names the fields actually searched.
- **The unchecked-row count was taken pre-limit**, so it could say "12 of the returned label(s)" when 3 were returned.

Seven new tests cover the disclosure itself, which previously had **none** — which is why it regressed once already on this branch. A paged fake transport makes the page-local case observable; the earlier fake made the total correction exact by construction, so no test could have caught it.

`pytest tests/unit -k "fda or openfda or faers"` under real addopts: **283 passed, EXIT=0** (was 276).

Issue #521, opened 2026-08-12
USPTOPatentDocumentDownloader: easyocr and python-docx imported at module top level but never declared as dependencies


tskatell:
### Summary

`src/tooluniverse/remote/uspto_downloader/uspto_downloader_tool.py` imports `easyocr` and `docx` (python-docx) at module top level, but neither package appears in the `dependencies` list in `pyproject.toml` (root or MCPB bundle). In a clean environment resolved from the declared dependencies, both imports fail, so `USPTOPatentDocumentDownloader` raises `ModuleNotFoundError` the first time it is invoked.

This is unrelated to the `fitz` -> `pymupdf` problem in #519 / #516 — I hit it while applying the #516 change locally and checking what else that module needs.

### Location

`src/tooluniverse/remote/uspto_downloader/uspto_downloader_tool.py`, lines 1-6:

```python
import requests
import fitz
import easyocr
from io import BytesIO
from docx import Document
from PIL import Image
```

### Evidence

Resolved an environment from the declared dependencies only (with `fitz` replaced by `pymupdf`, per #516) and checked each import:

```
FAIL easyocr  ModuleNotFoundError  No module named 'easyocr'
FAIL docx     ModuleNotFoundError  No module named 'docx'
OK   PIL
OK   fitz      (1.28.2, via pymupdf)
OK   pymupdf   (1.28.2)
```

`PIL` happens to resolve transitively (pillow arrives via `pdfplumber` / `markitdown[all]`), which is presumably why this has gone unnoticed — but `easyocr` and `python-docx` have no transitive path.

### Impact

The tool registers fine — the lazy registry loads all 240 tools without error, so nothing surfaces at import or startup. The failure only appears when a caller actually invokes `USPTOPatentDocumentDownloader`, as an unhandled `ModuleNotFoundError` rather than a useful message.

### Suggested fix

I'd suggest *not* adding these to the base dependency list — `easyocr` pulls in torch, which is a very large addition to a package whose other 600+ tools don't need it. Something like:

1. Declare them in an optional extra, e.g. `[project.optional-dependencies]` -> `ocr = ["easyocr>=1.7", "python-docx>=1.1"]`.
2. Move both imports inside `ocr_pdf_bytes()` (or guard them at module level) and raise a clear error pointing at the extra, e.g. `"USPTOPatentDocumentDownloader requires the 'ocr' extra: pip install tooluniverse[ocr]"`.

[...]

PR #523, opened 2026-08-12
Fix USPTO downloader MCP service packaging

gasvn:
## Summary

- keep the USPTO downloader as a separately deployed MCP service with its own GPU/OCR dependency manifest instead of adding EasyOCR/PyTorch to ToolUniverse or MCPB
- fix package imports, unwrap the current USPTO response envelope, preserve structured API errors, and fall back from an unusable Word document to PDF
- OCR only PDF pages with insufficient embedded text, and try later USPTO document versions when an earlier matching version cannot be downloaded or parsed
- expose the three advertised MCP tools with the advertised `applicationNumberText` parameter
- add opt-in Bearer authentication to MCP clients, scoped to the explicitly configured token environment variable; if the server token is configured but auth setup fails, startup fails closed
- remove the three stale `RemoteTool` placeholders that were registered locally but had no executable implementation; the MCP auto-loader is the single integration path
- include the server config, README, and dependency manifest in built wheels, and correct the deployment documentation

## Root cause

The GPU-side service source was present, but it was not maintained as a complete deployable service. Its heavy dependencies were undeclared, imports pointed at modules that do not exist in the remote package, its metadata parser expected an older USPTO response shape, and required service files were omitted from wheels. The MCP server's names and schemas also disagreed with ToolUniverse's published definitions.

A second audit found four additional problems: enabling server authentication made `MCPAutoLoaderTool` receive HTTP 401 because it sent no token; one failed matching USPTO document prevented trying later versions; OCR was selected only when the entire PDF was empty rather than per page; and three non-executable `RemoteTool` placeholders remained in the default registry.

This PR does **not** add EasyOCR, PyTorch, python-docx, or PyMuPDF to the default ToolUniverse/MCPB dependency set.

## Validation

- `pytest tests/tools/test_uspto_tool.py tests/unit/test_uspto_downloader_tool.py tests/unit/test_mcp_client_auth.py -q --no-cov` — 29 passed
- MCP protocol/functionality integration suites — 38 passed
- Ruff, formatting, JSON validation, compile checks, and pre-commit hooks passed
- started an authenticated FastMCP 3.4.5 HTTP service and confirmed `MCPAutoLoaderTool` discovered all three remote tools using the configured Bearer token
- reproduced the multiple-document failure and confirmed the repaired implementation tries the second matching document successfully
- exercised real python-docx and PyMuPDF parsing paths
- built and installed the final wheel; confirmed the MCP loader, server files, and separate requirements are present and the stale remote placeholder config is absent

## Remaining merge gate

This remains a draft because the scanned-PDF EasyOCR path has not been exercised on a real CUDA GPU with a real USPTO API key. The non-GPU parsers, MCP contract, authenticated HTTP discovery, error handling, document-version retry, registry behavior, and wheel packaging are covered, but the PR should not be marked ready until that GPU end-to-end check is completed.

Refs #521


gasvn:
Closing this in favor of #546, which landed independently and already covers what this PR set out to fix:

- Renamed the three MCP tools to the advertised names with typed `applicationNumberText` parameters
- Word→PDF fallback, retrying other USPTO document versions when one fails to download/parse
- Structured API error propagation
- Opt-in Bearer authentication on the MCP server (via the shared `server_security` module)

On top of that, #546's implementation adds hardening this PR didn't have: SSRF/redirect validation restricted to `*.uspto.gov`, download size caps, DOCX zip-bomb protection, and PDF page/OCR pixel limits. It also uses EasyOCR in CPU-only mode (`gpu=False`), which removes the real-CUDA-GPU validation blocker this PR was stuck behind.

The one thing this PR still does better is OCR granularity: per-page OCR (only OCR pages lacking embedded text) vs. #546's whole-document fallback (OCR only when the entire PDF extracts to zero text). I'll follow up with a small, focused PR on top of main for just that improvement.

Issue #563, opened 2026-09-05
[Feature request] Add Exa (exa.ai) as a tool — search API and/or MCP server


LMayoGit:
### Request

Would you consider adding **Exa** (https://exa.ai/) as a tool in ToolUniverse?

### What it is (short version)

Exa is a search API built for LLMs/agents rather than for humans. It runs embedding-based (neural) search alongside keyword search over its own web index, and the endpoints return agent-ready output: search results, full page contents/text extraction, highlights, "find similar pages" for a given URL, and higher-level answer/research endpoints that synthesize across sources with citations. They also maintain an official MCP server on top of the same API.

### Why it could be a good fit

- Gives ToolUniverse a general web/open-internet retrieval channel, which is currently much thinner than its coverage of structured scientific databases.
- Semantic search plus content extraction in one call is convenient for agents that need grounding text, not just links.
- "Find similar" and the research/answer endpoints map nicely onto literature-adjacent and background-context steps in scientific workflows (preprints, lab and consortium pages, docs, news, datasets) that aren't indexed in PubMed-style sources.

### Scope question

I don't have a strong preference here, so I'd rather follow whatever fits the project best:

1. Wrapping the full Exa **search API** as native ToolUniverse tools (search, contents, find-similar, answer/research), or
2. Registering their **MCP server** through ToolUniverse's MCP support, or
3. Both, with the MCP server as the quick path and the API for finer-grained control over parameters.

Do the maintainers have a preference on scope (full search API vs. their MCP server specifically)? Happy to align with that before any implementation work starts.

### One more ask

Could the relevant workflow be updated as part of this as well — i.e. the tool registration/config files, the docs/tool listing, and the CI workflow entries — so the new tool is discovered, documented and tested the same way as the existing ones?

gasvn:
Thanks for the detailed request, @LMayoGit! Went with registering Exa's own MCP server (option 2) rather than wrapping their full REST API.

Opened #567 with the implementation — I've verified it live end-to-end already (real search and fetch calls against Exa's real server both returned real results), so this one should just work.

**Would appreciate your help confirming it on your end too**, since you're the one who'll actually be using it. To try it before it merges:

```bash
# Install ToolUniverse from the PR branch instead of PyPI:
uv pip install "git+https://github.com/mims-harvard/ToolUniverse.git@feature/exa-mcp-integration"
```

[...]

Issue #562, opened 2026-09-05
[Feature request] Add Consensus (consensus.app) as a tool — search API and/or MCP server


LMayoGit:
### Request

Would you consider adding **Consensus** (https://consensus.app/) as a tool in ToolUniverse?

### What it is (short version)

Consensus is an AI-powered academic search engine over a large corpus of peer-reviewed papers. Rather than just returning a ranked list of links, it pulls the relevant findings out of individual papers and aggregates them into an evidence view for a research question (e.g. their "consensus meter" for yes/no questions), along with study-level metadata such as study design, sample size and journal quality. They expose a search API, and they also publish an MCP server.

### Why it could be a good fit

- Complements the existing literature tools (PubMed, Europe PMC, OpenAlex, etc.) by adding evidence synthesis on top of raw retrieval.
- Useful for claim checking / hypothesis triage inside agentic workflows, where an agent wants a "what does the literature broadly say about X" answer in a single call instead of reading N abstracts.
- Corpus is not biomedical-only, so it broadens coverage relative to PubMed-centric tools.

### Scope question

I don't have a strong preference here, so I'd rather follow whatever fits the project best:

1. Wrapping the full Consensus **search API** as native ToolUniverse tools, or
2. Registering their **MCP server** through ToolUniverse's MCP support, or
3. Both, with the MCP server as the thin path and the API for finer-grained control.

Do the maintainers have a preference on scope (full search API vs. their MCP server specifically)? Happy to align with that before any implementation work starts.

### One more ask

Could the relevant workflow be updated as part of this as well — i.e. the tool registration/config files, the docs/tool listing, and the CI workflow entries — so the new tool is discovered, documented and tested the same way as the existing ones?

gasvn:
Thanks for the detailed request, @LMayoGit! For this one I went with wrapping Consensus's REST search API (option 1) rather than their MCP server — their MCP server turned out to require full OAuth 2.0, which isn't something ToolUniverse's MCP framework supports yet (that's a separate, bigger piece of infrastructure work, unrelated to Consensus specifically), while their REST API just needs a simple API key.

Opened #568 with the implementation. Being upfront about where it stands: **I don't have a Consensus account, so I haven't been able to test this against a real API call** — everything's been checked carefully against their documented schema and live (unauthenticated) probes of the real endpoint, but not an actual successful search.

**If you already have (or can get) a Consensus API key, I'd really appreciate you trying this before it merges:**

```bash
[...]

PR #568, opened 2026-09-08
Feature: add Consensus_search_papers tool (#562) — needs a real key to verify

gasvn:
## Summary

Addresses #562: adds a native `Consensus_search_papers` tool wrapping Consensus's REST search API (`GET https://api.consensus.app/v1/search`), returning papers from their 220M+ peer-reviewed corpus with an AI-generated per-paper takeaway and study-design metadata (study type, sample size, journal quality) — evidence triage rather than raw retrieval, complementing PubMed/Europe PMC/OpenAlex.

## Scope decision: native API tool, not their MCP server

#562 offered two options. Consensus also publishes an official MCP server (`https://mcp.consensus.app/mcp`), which was the initially-approved scope to match how #563 (Exa) was registered. Confirmed live that it requires full OAuth 2.0 (`WWW-Authenticate: Bearer resource_metadata=".../oauth-protected-resource"`), which ToolUniverse's MCP client framework doesn't support at all — only static header/Bearer-token auth. Building OAuth support is a separate, larger infrastructure task, not specific to Consensus, so this PR wraps their plain REST API instead — the other option #562 itself offered. Confirmed live that the REST API needs only a static `x-api-key` header (`401` + `WWW-Authenticate: APIKey` on an unauthenticated probe), matching dozens of other API-key-gated tools already in ToolUniverse.

## ⚠️ Not verified against a real API call — needs a reviewer with a Consensus account

I don't have a Consensus API key (their API requires a paid Pro/Deep/Teams account) and didn't want to sign up for a paid third-party account to test this myself. Everything checkable **without** a real key has been checked carefully:

- Confirmed live (unauthenticated probes against `api.consensus.app`): the endpoint exists, the method is `GET` (not `POST` — that returns `405`), the header name is `x-api-key`, and the auth-failure shape is a clean `401`.
- Resolved a real ambiguity: Consensus has two endpoints, `/v1/quick_search` (deprecated, removal 2027-02-07) and `/v1/search` (current). Confirmed via their own docs (`llms.txt`) that `/v1/search` — what this PR uses — is the correct, current one.
- Cross-checked all 22 documented request parameters and the full response shape (`results`, `page`, `page_size`, `is_end`, `next_page`, and every per-paper field) against [Consensus's own published API reference](https://github.com/Consensus-NLP/consensus-api) — all present, none missing, none extra.
- Verified `request_with_retry`'s actual contract by reading its source (never raises on non-2xx, doesn't retry `401`) matches what this tool's error handling assumes.

**One genuine open question:** how Consensus expects boolean parameters (`human`, `controlled`, `exclude_preprints`, `medical_mode`, `open_access`, `include_semantic_score`, `include_full_text_chunks`) serialized in the query string. No example anywhere I could access shows a real boolean value. This tool sends literal `"true"`/`"false"` strings, which is the standard FastAPI/Pydantic convention (their `server: uvicorn` header strongly implies FastAPI) — reasonably confident, not certain.

**If you have a Consensus API key, testing this and confirming (or fixing) the boolean handling would be genuinely useful before merging.**

## Test coverage

`tests/tools/test_consensus_tool.py` mocks the HTTP layer using Consensus's own documented example response as the fixture (following the same convention as `test_semantic_scholar_tool_resilience.py`): missing-query error, missing-API-key error, successful parse, boolean/passthrough filter forwarding, `401` handling, retryable-status marking. All pass, but none of them touch the real network.

## How to test this yourself

```bash
# Install ToolUniverse from this branch instead of PyPI:
uv pip install "git+https://github.com/mims-harvard/ToolUniverse.git@feature/consensus-search-tool"

# Or via uvx directly, no separate install step:
CONSENSUS_API_KEY=your_real_key uvx --from "git+https://github.com/mims-harvard/ToolUniverse.git@feature/consensus-search-tool" tooluniverse
```

Or point an existing checkout at this branch and `pip install -e .`, then with `CONSENSUS_API_KEY` set:

```python
from tooluniverse import ToolUniverse
tu = ToolUniverse()
tu.load_tools(categories=["consensus"])
print(tu.run({"name": "Consensus_search_papers", "arguments": {"query": "Does creatine improve cognition?"}}))
```

If the boolean filters (`human`, `open_access`, etc.) don't behave as expected, that's exactly the thing I couldn't verify without a key — a report or a fix would be very welcome.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -76,6 +76,10 @@ HF_TOKEN=
 
 # === Literature & Patents ===
 
+# CONSENSUS_API_KEY (required) — Search 220M+ peer-reviewed papers via the Consensus API. (https://consensus.app/home/api)
+# Without: Consensus tools are blocked without it. Requires a paid Consensus account (Pro, Deep, or Teams plan).
+CONSENSUS_API_KEY=
+
 # NCBI_API_KEY (optional) — Search PubMed and fetch citation metrics (iCite) and PubTator annotations. (https://www.ncbi.nlm.nih.gov/account/)
 # Without: Works at 3 requests/sec without it; the key raises the limit to 10/sec.
 NCBI_API_KEY=
--- a/src/tooluniverse/consensus_tool.py
+++ b/src/tooluniverse/consensus_tool.py
@@ -1,0 +1,120 @@
+import os
+
+import requests
+
+from .base_tool import BaseTool
+from .http_utils import request_with_retry
+from .tool_registry import register_tool
+
+_BOOL_PARAMS = (
+    "human",
+    "controlled",
+    "exclude_preprints",
+    "medical_mode",
+    "open_access",
+    "include_semantic_score",
+    "include_full_text_chunks",
+)
+_PASSTHROUGH_PARAMS = (
+    "year_min",
+    "year_max",
+    "month_min",
+    "month_max",
+    "study_types",
+    "sample_size_min",
+    "sjr_min",
+    "sjr_max",
+    "citation_min",
+    "domain",
+    "country",
+    "journal_name",
+    "publisher_name",
+    "page",
+)
+
+
+@register_tool("ConsensusTool")
+class ConsensusTool(BaseTool):
+    """
+    Search 220M+ peer-reviewed papers via the Consensus API.
+
+    API key is required and read from the CONSENSUS_API_KEY environment
+    variable. Request one at: https://consensus.app/home/api
+    """
+
+    def __init__(self, tool_config, base_url="https://api.consensus.app/v1/search"):
+        super().__init__(tool_config)
+        self.base_url = base_url
+        self.session = requests.Session()
+        self.session.headers.update({"Accept": "application/json"})
+
+    def run(self, arguments):
+        query = arguments.get("query")
+        if not query:
+            return {"status": "error", "error": "`query` parameter is required."}
+
+        api_key = os.environ.get("CONSENSUS_API_KEY", "")
+        if not api_key:
+            return {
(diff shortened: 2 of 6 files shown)


LMayoGit:
Tested this from `feature/consensus-search-tool` with a real `CONSENSUS_API_KEY` (installed via `uv pip install "git+https://github.com/mims-harvard/ToolUniverse.git@feature/consensus-search-tool"`).

**Basic query works:** `tu.run({"name": "Consensus_search_papers", "arguments": {"query": "Does creatine improve cognition?"}})` — 200 OK, real papers back with abstract, DOI, `takeaway`, `sjr_best_quartile`, `is_preprint`, `citation_count`, etc. Response shape matches exactly what's documented in the PR description, nothing missing or extra.

**Resolved the genuinely open question — boolean param serialization:** the tool sends literal lowercase `"true"`/`"false"` strings, and this is correct. Tried `human`, `open_access`, `controlled` individually and combined (`controlled=True, human=True`) against the real API — all returned 200, none rejected. So the FastAPI/Pydantic-convention guess in the PR description was right.

**Filter behavior (informational, not a bug):** compared `human=True` vs `human=False` on a mice/rats-heavy query — result count and ordering did change (20 vs 19, different top hits), so the param is doing *something* server-side, not being silently ignored. That said, a couple of animal-study papers still showed up under `human=True`, suggesting Consensus's `human` tag is a soft/AI-scored filter rather than a hard boolean gate. That's Consensus's own filter semantics, not something to fix in this tool.

No bugs found. This PR is good to merge as far as live-API behavior goes.

PR #569, opened 2026-09-09
Add SerpBase backend to web_search (#564)

### LINKED ISSUE #564: Feature Request: add SerpBase as an optional Google organic-results backend for WebSearchTool
Hi — WebSearchTool has quietly become the fallback I rely on inside ToolUniverse when a literature tool needs grounding, and the way `_search_with_fallback` chains DDGS backends before dropping to duckduckgo_html and wikipedia_api has saved me more than once in restricted environments.

The pain point I keep hitting is Google coverage. DDGS does expose `backend="google"`, but it routes through DuckDuckGo's scraping layer, and in practice it's the flakiest entry in the chain — which is presumably why `auto` leads with `duckduckgo`/`bing`/`brave` instead. On shared/VPS IPs it rate-limits quickly, and when it fails the fallback lands on Bing or DuckDuckGo HTML, so queries that genuinely need Google's fresh, rank-sensitive results (recent preprints, drug-pricing pages, niche regulatory docs) come back noticeably weaker. Parallel is a great addition, but it's an MCP call to a hosted endpoint and doesn't return true Google SERP ordering either.

What I'd find valuable is an optional, opt-in backend that returns real Google organic results as plain JSON over a single GET — something like https://serpbase.dev (a Google Search Results API: free tier, pay-as-you-go, no subscription, no scraping to maintain). The response is an `organic_results` array with title/link/snippet/position, which maps almost 1:1 onto the `{title, url, snippet, rank}` dicts WebSearchTool already returns.

Why it fits this codebase:

- `WebSearchTool` already has a clean provider pattern (`_search_with_ddgs`, `_search_with_parallel`, `_search_with_duckduckgo_html`, `_search_with_wikipedia_api`) — a SerpBase provider would just be another `_search_with_*` method returning the same dict shape, added to the `auto` fallback chain *after* the keyless providers or invoked explicitly via `backend="serpbase"`.
- The subprocess + JSON round-trip for DDGS exists precisely because native panics and scrapers are fragile; an HTTP GET backend is the opposite end of that spectrum — no subprocess, no selector maintenance, and the `requests` dependency is already there (used by the HTML fallbacks).
- It degrades gracefully: gate it on a `SERPBASE_API_KEY` env var like the other keyed tools in this repo (`.env.template` is auto-generated from `api_key_info` blocks, so it would slot into that catalog), and skip to the next provider when the key is absent — the keyless default chain stays exactly as it is today.

Happy to draft the provider method and the `.env.template`/config registration if there's interest — it's a contained change and I can test it against the existing test suite for `test_web_search_tool.py`.

gasvn:
## Summary

Fixes #564.

Adds `backend='serpbase'` to `web_search` / `web_api_documentation_search`: an opt-in real Google-SERP backend using SerpBase's API (https://api.serpbase.dev/google/search), for when DDGS's Google backend is unavailable or rate-limited.

## Root cause / motivation

DDGS's Google backend is unreliable/rate-limited in some environments. The issue proposed adding a real SERP API (SerpBase) as an alternative backend.

## Correction to the originally proposed request shape

The issue's proposed design didn't match SerpBase's actual live API. Verified directly against `https://api.serpbase.dev/google/search`:

| | Issue's proposal | Actual (verified live) |
|---|---|---|
| HTTP method | `GET` | `POST` (confirmed `GET` returns 405) |
| Results field | `organic_results` | `organic` |
| Result item fields | `position`, `url` | `rank`, `link` |
| Auth failure | (assumed non-2xx) | HTTP 200, with `error` in the JSON body |

Built against the verified-correct shape rather than the issue's description.

## Fix

- New `_search_with_serpbase()` method in `web_search_tool.py`: POSTs `{q, gl, hl}` (region split on `-` into country/language), checks the JSON body's `error` field explicitly (SerpBase returns 200 even on auth/quota failure), normalizes `organic` results (`link`→`url`, `rank`, `title`, `snippet`).
- Wired into `_search_with_fallback()`'s `serpbase` branch: falls back to the `auto` DDGS chain on any failure (missing key, API error, network error), mirroring the existing `parallel` backend's fallback behavior exactly.
- `run()` rejects non-default `safesearch` for `backend='serpbase'` (not supported by the API); `region` **is** supported (unlike `parallel`).
- `provider_notice` disclosure is attached whenever `serpbase` was attempted, same as for `parallel`.
- Added `SERPBASE_API_KEY` as an optional key in `web_search_tools.json` with `api_key_info`, and regenerated `.env.template` / `api_keys_catalog.json` via `scripts/gen_api_key_catalog.py`.

## Reproduction (before)

No `serpbase` backend existed; `backend='serpbase'` would have raised a schema validation error against the enum.

## Validation (after)

Live smoke test with no `SERPBASE_API_KEY` set (confirms clean fallback, no crash):

```
tu.run({"name": "web_search", "arguments": {"query": "test", "backend": "serpbase"}})
→ {'status': 'success', 'data': {..., 'backend_used': 'auto',
   'attempted_backends': ['serpbase', 'auto'],
   'provider_errors': {'serpbase': 'SERPBASE_API_KEY environment variable is not set...'},
   'provider_notice': "Query sent to SerpBase's hosted third-party Google-SERP API..."}}
```

Full `ToolUniverse().load_tools()` smoke test: 2716 tools loaded cleanly with `SERPBASE_API_KEY` unset (no regression from adding an optional key).

**Not yet verified**: an actual successful authenticated call. I wasn't able to complete SerpBase's account signup myself (no browser automation available on my end). Method, auth header, error-body shape, and field names were all confirmed live against the real (unauthenticated) endpoint plus their published docs — but a real search response has not been observed.

## Test coverage

13 new tests in `tests/tools/test_web_search_tool.py` (32 total, all passing): missing-key handling, response parsing against the documented shape, region splitting, `error`-field-with-200 handling, missing/malformed `organic` handling, truncation to `max_results`, `run()`-level success/fallback/safesearch-rejection, and that the `auto` chain never calls `serpbase` (matching the existing guarantee for `parallel`).

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -220,6 +220,10 @@ REPLICATE_API_TOKEN=
 
 # === Other ===
 
+# SERPBASE_API_KEY (optional) — Query the backend='serpbase' option for real Google SERP results (https://api.serpbase.dev), billed per request. (https://serpbase.dev)
+# Without: web_search still works via DDGS/Parallel/DuckDuckGo-HTML/Wikipedia; backend='serpbase' is blocked without it.
+SERPBASE_API_KEY=
+
 # WAQI_API_KEY (optional) — Fetch real per-city air quality data from the World Air Quality Index API. (https://aqicn.org/data-platform/token/)
 # Without: Falls back to the public 'demo' token, which returns a fixed sample station (Shanghai) for every city. The key is free.
 WAQI_API_KEY=
--- a/src/tooluniverse/data/api_keys_catalog.json
+++ b/src/tooluniverse/data/api_keys_catalog.json
@@ -537,6 +537,18 @@
     ]
   },
   {
+    "name": "SERPBASE_API_KEY",
+    "requirement": "optional",
+    "type": "secret",
+    "domain": "Other",
+    "register_url": "https://serpbase.dev",
+    "purpose": "Query the backend='serpbase' option for real Google SERP results (https://api.serpbase.dev), billed per request.",
+    "without": "web_search still works via DDGS/Parallel/DuckDuckGo-HTML/Wikipedia; backend='serpbase' is blocked without it.",
+    "used_by": [
+      "web_search_tools.json"
+    ]
+  },
+  {
     "name": "TXAGENT_MCP_SERVER_HOST",
     "requirement": "required",
     "type": "endpoint",
--- a/src/tooluniverse/data/web_search_tools.json
+++ b/src/tooluniverse/data/web_search_tools.json
@@ -2,7 +2,7 @@
   {
     "type": "WebSearchTool",
     "name": "web_search",
-    "description": "General web search using DDGS (Dux Distributed Global Search) supporting multiple search engines including Google, Bing, Brave, Yahoo, and DuckDuckGo. No API keys required. Parallel Search MCP is opt-in; selecting it sends the query to Parallel's hosted third-party service, so do not submit sensitive or patient-identifying information.",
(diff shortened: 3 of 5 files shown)


gefsikatsinelou:
Thanks for the quick merge, @gasvn — and for building it against the live endpoint rather than the write-up; that saved a round trip. Glad the real-key test results were useful. Happy to re-test if anything changes post-merge.

PR #576, opened 2026-09-11
fix: preserve all sections of guideline abstracts

krudo-taco:
## Summary

Preserve every AbstractText section and nested inline text for PubMed and Europe PMC guideline abstracts. The old paths could omit Results/Conclusions while still returning successful evidence. Use one shared XML flattening helper, retain entity decoding, and repair two incomplete synthetic XML fixtures.

## Validation

- [ ] Full touched-file `ruff check`: pre-existing diagnostics remain; no full-profile pass is claimed.
- [x] Relevant tests: 6 passed, 8 warnings.
- [x] Generated SDK/tool files: not applicable; tool metadata was not changed.
- [x] New tests isolate network requests with synthetic mocks.
- Targeted Ruff F/E9 and new-test formatting checks passed; `git diff --check` passed.

Commands:
```sh
uv run --no-project --with-editable . --with pytest pytest tests/unit/test_guideline_multi_section_abstracts.py tests/unit/test_pubmed_guidelines_entity_decoding.py tests/unit/test_pubmed_guidelines_book_records.py -q --disable-warnings -o addopts=
```

## Checklist

- [x] User-facing problem described above.
- [x] Existing schema/public API is retained; no separate documentation change is required for this correction.
- [x] Added regression tests for the affected behavior.

Nonempty malformed PubMed XML now produces a visible parse error. A valid response without an abstract can still have an empty abstract; no regex recovery is used.

### DIFF
--- a/src/tooluniverse/unified_guideline_tools.py
+++ b/src/tooluniverse/unified_guideline_tools.py
@@ -4,7 +4,6 @@ Unified Guideline Tools
 Consolidated clinical guidelines search tools from multiple sources.
 """
 
-import html
 import requests
 import time
 import re
@@ -87,6 +86,16 @@ def _extract_meaningful_terms(query):
     }
     meaningful = [token for token in tokens if token not in stop_terms]
     return meaningful if meaningful else tokens
+
+
+def _abstract_text_from_sections(element: ET.Element) -> str:
+    """Flatten every structured abstract section, preserving inline text."""
+    sections = element.findall(".//AbstractText")
+    return " ".join(
+        "".join(section.itertext()).strip()
+        for section in sections
+        if "".join(section.itertext()).strip()
+    )
 
 
 @register_tool()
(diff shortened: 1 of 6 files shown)


gasvn:
Thanks for this — the root-cause diagnosis (regex extraction only grabbing the first `AbstractText` node, and dropping structured sections) is right, and I like that you went and fixed the two malformed synthetic fixtures in `test_pubmed_guidelines_book_records.py` and `test_pubmed_guidelines_entity_decoding.py` while you were in there.

There's a third one that got missed, though, and it's failing CI: `tests/unit/test_pubmed_guidelines_total.py`'s `ABSTRACTS` fixture (lines 50-54) concatenates two `<PubmedArticle>` elements with no wrapping root:

```python
ABSTRACTS = "".join(
    f"<PubmedArticle><PMID>{pmid}</PMID>"
    f"<AbstractText>Abstract {pmid}</AbstractText></PubmedArticle>"
    for pmid in ("1", "2")
)
```

The old regex-based extraction didn't care whether this was well-formed XML, but the new `ET.fromstring(abstract_response.text)` call does — two sibling root elements with nothing wrapping them is a parse error ("junk after document element"), so `_search_pubmed_guidelines` now hits the new `except ET.ParseError` branch and returns `status: "error"` instead of `"success"`. That cascades into 4 failures in that file (`KeyError: 'metadata'` and `assert 'error' == 'success'`) — reproduced locally on this branch, not just a flaky CI run.

Should be the same fix as the other two fixtures: wrap it in a single root (e.g. `<PubmedArticleSet>...</PubmedArticleSet>`), same as `_MULTI_SECTION_XML` in your own new test file. Happy to merge once that's green.

krudo-taco:
Wrapped the remaining `ABSTRACTS` fixture in `tests/unit/test_pubmed_guidelines_total.py` in a `<PubmedArticleSet>` root, same as the other two. Also wrapped the sibling stub in `test_pubmed_guidelines_envelope.py` so a multi-PMID envelope test does not hit the same parse error.

Locally: 17 related tests passed.

gasvn:
Thanks for the quick turnaround — that's the one. Verified on your branch: the 4 previously-failing tests in `test_pubmed_guidelines_total.py` now pass, and CI is green across the board.

I also test-merged this against current `main` locally, since #577 and #578 landed in `unified_guideline_tools.py` after you cut this branch. It merges cleanly and the whole guideline suite (51 tests) passes on the merged result, so the three changes compose fine. Merging now.

PR #580, opened 2026-09-14
Fix: verify MCP tool contracts per-tool instead of aborting the whole category

gasvn:
## Summary

When `strict_tool_contracts` is enabled, `_verify_and_pin_contracts` raised on the **first** tool whose upstream schema had drifted, which aborted `discover_tools()` for the entire loader. One changed tool therefore deregistered every other reviewed tool in that category — including tools whose contracts still matched exactly.

Surfaced by @LMayoGit in #563.

## Root cause

`mcp_client_tool.py` validated all `selected_tools` in one pass and raised immediately on any mismatch:

```python
if reviewed_hash != remote_hash:
    raise ValueError(f"MCP contract changed for reviewed tool: {name}")
```

`discover_tools()` wraps that in a `try/except` that re-raises, so the whole category fails to load.

This happened in practice: Exa added a required `objective` field to `web_search_exa`, which also took `web_fetch_exa` offline even though its hash was unchanged (verified — `web_fetch_exa`'s live contract still hashes to the pinned `c52bc770…`).

## Fix

Each reviewed tool is now verified independently:

- Tools that fail verification (drifted contract, missing from server, no reviewed contract) are **excluded** and logged at `WARNING` with the specific reason per tool.
- The remaining verified tools still load.
- **The pin is not weakened** — a drifted tool is never exposed, only skipped. Excluding it is strictly safer than accepting it and strictly more available than taking down its siblings.
- If **no** reviewed tool passes verification, the loader still raises, so fail-closed behavior is preserved when nothing can be trusted.

## Validation

- Both pre-existing fail-closed tests (`test_strict_loader_fails_closed_on_contract_drift`, `test_strict_loader_fails_closed_when_reviewed_tool_is_missing`) still pass unchanged — they use single-tool configs, so nothing verifies and the loader still raises.
- 3 new regression tests. I confirmed each **fails against the old code** and passes with the fix, so they're genuine regression guards rather than vacuous assertions:
  - `test_one_drifted_tool_does_not_take_down_its_unaffected_siblings`
  - `test_a_missing_tool_does_not_take_down_its_siblings`
  - `test_drifted_tool_is_never_exposed_even_when_siblings_load`
- `tests/unit/test_mcp_contract_hardening.py`: 15 passed.
- All unit-level MCP tests: 88 passed.
- Folklore / Noodle / GI loader config tests: 5 passed (these share this code path).
- `ruff check` and `ruff format --check` clean on both touched files.

Pre-existing unrelated failures in `tests/integration/test_mcp_protocol.py` and `test_smcp_http_server.py` (they need a live server subprocess) reproduce identically on clean `main`.

## Note on blast radius

This changes shared behavior for every MCP auto-loader (Exa, Folklore, Noodle, GI), which is why it's split out from the Exa contract repin in #567 rather than bundled into it.

## Update: this is affecting three categories in production right now

After opening this PR I audited every contract-pinned MCP loader against its live server. Two more categories are **completely dead on current `main`** for the same reason — one drifted tool taking down all its siblings:

| Loader | Reviewed tools | Drifted | Status on `main` | With this PR |
|---|---|---|---|---|
| Exa | 2 | 1 | dead | 1 loads (2 after #567 repin) |
| **Folklore** | 4 | 1 (`get_publication_details`) | **dead — 0 tools** | **3 load** |
| **Genomic Intelligence** | 15 | 8 | **dead — 0 tools** | **7 load** |
| Noodle | 6 | 0 | fine | fine |

Verified live, e.g. Folklore on current `main`:

```
⚠️ Remote MCP server 'mcp_auto_loader_folklore' is unavailable:
   MCP contract changed for reviewed tool: get_publication_details.
LOADED: ['mcp_auto_loader_folklore']        # all 4 tools gone
```

and the same environment on this branch:

```
⚠️ Skipping reviewed MCP tools that failed contract verification:
   get_publication_details (contract changed). Re-review the upstream schema...
LOADED: [..., 'folklore_search_variant_evidence',
         'folklore_search_variant_literature', 'folklore_search_literature_corpus']
```

So this isn't a hypothetical robustness improvement — it currently restores 10 working tools across two categories that are entirely unavailable today.

Note the drifted contracts still need repinning to be fully restored (9 across Folklore and GI); that's a separate review decision, since accepting an upstream schema change is exactly what the pin is meant to gate.


gasvn:
Revised after review — the first version was too permissive.

It demoted **every** verification failure to a per-tool skip, including a reviewed tool being **absent** from the server. That's wrong: every reviewed tool being present is a check on the server's *identity*, not on any one tool's schema. An endpoint serving only a subset may be stale, rolled back, or simply the wrong host, and these schemas are public enough that a subset could match byte-for-byte. The effective bar dropped from "all N reviewed tools present and matching" to "at least one matches" — a real weakening, and well beyond what the outage actually required.

Changes in `7e571e55`:

- **Missing reviewed tools fail the whole category again**, exactly as before this PR.
- **A selected name with no reviewed contract raises loudly** instead of quietly shrinking the category — that's a local config typo, not upstream drift.
- **Only genuine contract drift is survivable**, which is what the Exa / Folklore / GI outages actually are.
- Dropped the overreaching "excluded tools are never exposed" comment; it now says only what's true — a drifted tool isn't pinned, so it produces no proxy config and isn't registered *by that load*.

Re-verified live against the real servers: Folklore still recovers 3 of 4 tools and Genomic Intelligence 7 of 15, unchanged. Added a test asserting the completeness property with a **two-tool** loader — the two pre-existing fail-closed tests use single-tool configs, so they only covered the "nothing survived" path and wouldn't have caught this.

PR #584, opened 2026-09-14
Keep MCP tools running when upstream drift cannot break the call

gasvn:
## Summary

Completes the pair started in #583. A pinned third-party MCP tool was disabled by **any** change to its schema, so a vendor rewording a parameter description took the tool away from users — who are the one party that cannot review a third-party schema or re-pin it.

Accepting the live contract instead is not an option: an agent fills in whatever parameter a server advertises, and reads whatever `return_schema` a server describes.

So this answers a narrower, decidable question:

> **Would the contract we reviewed still work against the live server?**

## The rule

Using the schemas recorded in `mcp_contracts.lock.json` (#583):

| Upstream change | Result |
|---|---|
| Description / title reworded (input or output) | **tool keeps working** |
| New **optional** input property | **tool keeps working** |
| New **required** property | disabled — reviewed calls would be rejected |
| Property removed or retyped | disabled |
| Constraint tightened (`pattern`, `additionalProperties`, …) | disabled |
| Output schema restructured | disabled |
| Recorded schema disagrees with the pinned hash | disabled |
| No recorded schema in the lockfile | disabled, exactly as before |

Conservative on purpose: anything it cannot prove harmless is refused, because a false reject costs availability (the status quo) while a false accept means a broken call or an unreviewed contract.

## Two properties make it safe to automate

**1. A tolerated tool is published from the reviewed contract in full — input *and* output.**

Nothing the server changed reaches the agent. A field it added is never advertised, so it is never filled in. And `outputSchema` is not merely descriptive: it reaches the model as `return_schema` via `get_tool_info` / `tool_specification`, so publishing a live one would let a server write arbitrary unreviewed text into the agent's context. This also keeps `contract_sha256` honest — the hash stamped on the proxy is the hash that was reviewed.

**2. The lockfile is a record of what was reviewed, never an authority of its own.**

A recorded contract is trusted only while it still hashes to the `contract_sha256` pinned in the loader config. Otherwise refreshing the lockfile alone — which `sync_mcp_contracts.py --update` does, re-recording straight from the server — would silently re-point the trust anchor at whatever is being served now, and drift would be compared live against live.

**The rule only ever decides availability, never trust.** Every re-pin stays a human decision on the CI diff from #583. A wrong answer costs a tool, not a secret.

## Verified live against Exa

Simulated a benign upstream rewording (reviewed copy differs from live by a description only):

```
ℹ️ Upstream changed these MCP tools compatibly, so they keep running on the
   reviewed schema: web_search_exa. The daily contract-drift check will raise
   the diff for review.

LOADED: ['exa_web_search_exa', 'exa_web_fetch_exa']      ← previously: 0 tools
agent sees query desc: "REVIEWED WORDING (upstream has since reworded…"   ← not upstream's
CALL OK: Title: AlphaFold Protein Structure Database
```

Also verified after the review fixes: Exa loads 2/2 and Genomic Intelligence 15/15 against the real servers, with a real search returning results.

## Review found three holes in the first version

Worth reading, because the first version's claimed safety property was breakable. Each was reproduced with a runnable probe before being fixed, and each now has a regression test that fails against the earlier commit.

**The lockfile had become the runtime trust anchor instead of `contract_sha256`.** Drift was compared live-against-live whenever the lockfile ran ahead of the config, so it always looked tolerable and the **live** input schema was published:

```
agent sees params: ['debug_context', 'q']    ← 'debug_context' added by the server
```

Reachable by ordinary use, not just by attack: `--update` re-records from the server and leaves updating the config hash as a separate manual step.

**`classify()` ignored `outputSchema` entirely**, so an output-only rewrite was tolerated unconditionally and the live schema went to the agent:

```
return_schema title: "IGNORE ALL PREVIOUS INSTRUCTIONS."
```

The original docstring claimed output "is not a path for data to leave" — true for *filling in fields*, but it missed that output is a path for text to **enter** the model's context.

**`_constraints_equal` stripped documentation keys from `properties` maps too**, so a property whose *name* was `description`/`title`/`examples` was never compared and a nested one could be retyped unnoticed. Comparison is now schema-aware and type-strict, so `enum: [1, 0]` is no longer equal to `enum: [true, false]` via Python's `1 == True`.

Also hardened `load_reviewed_contracts` against a structurally wrong but valid-JSON lockfile, which raised `AttributeError` and took down every tool in the category — including ones that had not drifted.

## Test coverage

23 tests in `tests/unit/test_mcp_compatible_drift.py`, including the end-to-end assertion that a tolerated tool's agent-facing `parameter` schema contains **only** reviewed properties, that server-authored output text never reaches the agent config, that a lockfile ahead of the pinned hash is refused, and that a missing lockfile entry still fails closed.

One assertion widened in `test_mcp_contract_hardening.py`: rejection reasons are now specific (`contract changed, and no reviewed schema is recorded`), so an exact-string match needed loosening. Behaviour is unchanged — still fails closed — and the phrase `contract changed` was deliberately kept in the message so the other pre-existing fail-closed tests pass untouched.

120 MCP unit tests and the full unit suite pass; `--check-config` clean; full load still 2716 tools; `ruff check`/`format` clean.

## Known limitation

Load-time only. A vendor can drift between a load and a call, and with no reload nothing notices. Verifying per call costs a round trip and is probably not worth it, since a structurally drifted tool generally rejects the call anyway — but it is a real boundary, not something this closes.


gasvn:
Adversarial review found the safety property was **breakable**. Three holes, all reproduced with runnable probes before fixing, all now covered by regression tests that fail against the previous commit. Pushed in `a68aa7dc`.

**1. The lockfile had become the trust anchor instead of `contract_sha256` (critical).**
`load_reviewed_contracts()` never checked the recorded schema against the pinned hash. Whenever the lockfile ran ahead of the config, drift was compared live-against-live, always looked tolerable, and the **live** input schema was published to the agent. Reproduced:

```
tool survived contract verification: ['search']
agent sees params: ['debug_context', 'q']      <- 'debug_context' was added by the server
leaked server field: True
```

This was reachable by ordinary use, not just attack: `sync_mcp_contracts.py --update` re-records straight from the server and leaves updating the config hash as a **separate manual step**. Forget it once and every tool in that loader silently moves onto the live schema. A recorded contract is now trusted only while it hashes to the pinned `contract_sha256`.

**2. `classify()` ignored `outputSchema` entirely (high).**
An output-only rewrite tripped the hash, reached `classify()` and was tolerated unconditionally — and the live `outputSchema` was then handed to the agent as `return_schema`, which `get_tool_info` surfaces verbatim. So a server could write arbitrary unreviewed text into the agent's context by touching only its output schema:

```
return_schema title: "IGNORE ALL PREVIOUS INSTRUCTIONS."
```

My docstring claimed output "is not a path for data to leave" — true for *filling in fields*, but I had missed that it is a path for text to **enter** the model's context. Output drift is now checked, and a tolerated tool publishes the reviewed contract **in full**. That also makes the stamped hash honest: it is now the reviewed hash, where before it named a hybrid contract nobody had ever seen.

**3. `_constraints_equal` stripped documentation keys from `properties` maps too.**
So a property whose *name* was `description`/`title`/`examples` was never compared — a nested one could be retyped or removed unnoticed. Comparison is now schema-aware (property names are names, not keywords) and type-strict, so `enum: [1, 0]` is no longer treated as equal to `enum: [true, false]` via Python's `1 == True`.

Also hardened `load_reviewed_contracts` against a structurally wrong but valid-JSON lockfile, which raised `AttributeError` and took down every tool in the category including ones that had not drifted.

**Verification:** all four probes now blocked; doc-only drift still tolerated with the reviewed text shipping; 120 MCP tests and the full unit suite pass; live Exa + GI load 2/2 and 15/15 with a real search succeeding; `--check-config` clean; 2716 tools.

The PR description above still describes the original design — the "keeps the live output schema" paragraph is now wrong. I'll leave correcting it until the approach here is confirmed.

PR #591, opened 2026-09-15
Fix Claude Desktop MCPB requirements and dependency setup

gasvn:
## Problem and resulting behavior

Installing the MCPB on a machine with Python 3.14 showed a Python requirement warning. Broadening the upper bound alone did not fix Desktop's version check: the manifest used PEP 440 commas, while Desktop expects a space-separated semver range. After clearing that warning, the old launcher downloaded dependencies during the MCP handshake and timed out on a cold install.

The bundle now passes the requirement check and uses MCPB 0.4's native UV runtime to install dependencies before connecting the server. Subsequent launches reuse the prepared environment.

## Changes

- Set manifest compatibility to `>=3.10 <3.15`; retain the equivalent PEP 440 range `>=3.10,<3.15` in the bundle pyproject, with the reason for the upper bound documented.
- Use `server.type: uv` so Desktop runs dependency setup during installation.
- Bundle `.python-version` selecting Python 3.12 for both installation and launch; use `uv run --no-sync` instead of the extra `--with .` installation during startup.
- Pin the manifest validator to MCPB CLI 2.1.2 and support an installed validator through `MCPB_CLI`.
- Add regression checks and document current-client requirements, first-install downloads, and enabling ToolUniverse under the chat's **+ → Connectors** menu.

## Validation

**Actual install and use:** macOS 27.0, Apple Silicon, Claude Desktop 2.110.0; system Python 3.14.6, extension Python 3.12.11.

- Installed the rebuilt bundle through Claude Desktop; requirements all met.
- Restarted Claude; the extension-managed server showed Running and successfully answered `initialize` and `tools/list`.
- Exercised the installed bundle's actual launcher: initialized in 2.354 seconds; `Sequence_gc_content(ATGC)` returned `gc_percent: 50.0` and `length: 4`; a live `UniProt_get_entry_by_accession(P04637)` returned the matching accession.
- Independently called `execute_tool` from a Claude chat and verified the GC result after enabling the connector for that chat.
- Successfully imported onnxruntime 1.20.1 with NumPy 2.5.3; CoreML and CPU providers were available.
- Targeted packaging tests: **38 passed, 1 skipped** (optional PyMuPDF unavailable). Ruff check/format and `git diff --check` passed. Official manifest validation and ZIP integrity checks passed.

The first installation was slow downloading large dependencies. The machine's old UV 0.7.12 also failed extracting a cache file; installation succeeded after updating it to the already-installed Homebrew UV 0.10.4. A duplicate manual `uvx --refresh tooluniverse` configuration was removed, leaving the extension-managed server. These are local environment repairs, not changes shipped inside the bundle.

This verifies the bundle running on **Python 3.12 on a machine with system Python 3.14**, not native Python 3.14 execution or every optional tool. The earlier Linux/aarch64 Python 3.14 validation is separate. The previously reported macOS 13 + native Python 3.14 onnxruntime wheel limitation is not resolved by widening a manifest bound. Older clients without MCPB 0.4 UV support must be updated.

## Delivery

Merged as `f7aeb281cc8e4f7fe5913944d7495d06d31debd9`. All PR CI checks passed.

The production [MCPB release](https://github.com/mims-harvard/ToolUniverse/releases/tag/mcpb) and its floating `mcpb` tag have been updated. [Download tooluniverse.mcpb](https://github.com/mims-harvard/ToolUniverse/releases/download/mcpb/tooluniverse.mcpb).

The final published artifact was rebuilt from the merge commit to include other changes already on main. The merged bundle's actual launcher and source passed the same GC calculation and live UniProt lookup using the already-installed dependency environment; initialization took 2.369 seconds. Earlier installation and real Claude chat verification above were performed with the PR-head build.

Published SHA256, verified against GitHub's asset digest:
`cffaf47382d050b605c40322512e79d77841c40932cf488c9a87406fe0ba0961`

The temporary `mcpb-test-py314` release and tag have been deleted after publication. Its old URL should no longer be used.


gasvn:
## Test build for macOS

Built from this branch so it can be tested on a real Mac — the manifest change only has an observable effect once it's inside an installed bundle.

**Download:** [tooluniverse.mcpb](https://github.com/mims-harvard/ToolUniverse/releases/download/mcpb-test-py314/tooluniverse.mcpb) (pre-release tag `mcpb-test-py314`, 8.7 MB)

`sha256: 2bbf31bade153d970e4f4c5d2a7f4786a9b706dd357a5fb7bb30fa89797c5496`

Verified inside the built bundle:

```
manifest version       : 1.4.1
python requirement     : >=3.10,<3.15      <- was >=3.10,<3.14
bundle requires-python : >=3.10,<3.15
```

The production `mcpb` release is untouched (still the 2026-09-07 asset).

### How to test

1. Remove any installed ToolUniverse extension first (Settings → Extensions)
2. Quit Claude Desktop completely
3. Reopen → Settings → Extensions → Advanced settings → Install Extension… → pick the downloaded file

**What to look at, in order:**

**(a) The requirements list in the install dialog.** This is the part that was broken — it should now read `Python >=3.10,<3.15` with a ✓ instead of ⚠, assuming your Python is 3.14.x. That alone confirms the manifest fix.

**(b) Whether it actually starts.** This is the part I could not verify. Passing the version check only means Claude Desktop lets it proceed — the dependencies still have to resolve on your machine.

If it fails at step (b), the log line worth grabbing is whatever mentions `onnxruntime`. On Apple Silicon its only Python 3.14 wheel is `macosx_14_0_arm64`, so:

- **macOS 14+** → expected to resolve (verified by resolution, not by a real install)
- **macOS 13** → expected to fail, and nothing in this PR can fix that

So `sw_vers -productVersion` and `python3 -V` alongside the result would tell us which case we're in.

### Cleanup

Delete the `mcpb-test-py314` tag once this merges and a real `.mcpb` ships.

PR #598, opened 2026-09-16
Fix: surface openFDA API errors instead of returning None, and send FDA_API_KEY

gasvn:
## Summary

openFDA reports quota, authentication and server problems with a JSON error
envelope. `search_openfda()` only special-cased `NOT_FOUND`; every other code
fell through to a bare `return None`, which a caller cannot distinguish from
"this label genuinely has no such field". For an LLM agent the consequence is
invisible: it receives an empty tool result, answers from parametric knowledge,
and nothing appears in the logs.

Separately, `fda_label_tool.py` built its requests as
`params={"search": q, "limit": limit}` with no `api_key`, so every `FDA_*` label
tool ran on openFDA's anonymous quota even when `FDA_API_KEY` was configured —
**1,000 requests/day** instead of 120,000.

## Impact observed

Measured while evaluating an agent against these tools:

| | rate-limited run | unthrottled run |
|---|---:|---:|
| FDA tool calls returning `None` | 1,995 / 8,888 (**22%**) | 17 / 7,884 (0.2%) |

Across a 10,130-question benchmark, 17.5% of biomedical tool results came back
empty. Accuracy tracked evidence completeness closely:

| Empty tool results per question | n | accuracy |
|---|---:|---:|
| 0% | 2,525 | 92.8% |
| <50% | 187 | 84.5% |
| ≥50% | 431 | 81.4% |
| no tool consulted | 25 | 48.0% |

Re-running 150 affected questions against this branch: empty results
**0 / 406**, accuracy 78.0% → 87.3%. No failure was visible without inspecting
saved trajectories.

## Changes

`openfda_tool.py`
- New `_openfda_request()`: adds a timeout (previously **none**, so a hung
  connection stalled the caller indefinitely) and bounded retry with
  exponential backoff on 429/5xx, honouring `Retry-After`.
- Returns a structured error for every non-`NOT_FOUND` code, carrying
  `error_details.code` and, for `OVER_RATE_LIMIT` / `API_KEY_INVALID`,
  `next_steps` naming the limit and where to get a key.
- `NOT_FOUND` behaviour is unchanged — that one really is absent data.
- Error returns carry the same `results` / `meta` shape as success, so callers
  cannot mistake one for the other.
- Warns once when `FDA_API_KEY` is rejected instead of silently downgrading.

`fda_label_tool.py`
- All three call sites now go through `_fda_get()`, which attaches the key,
  applies a timeout and retries 429/5xx with backoff.
- Placeholder key values (`your_api_key_here` and friends) are rejected rather
  than sent.
- An exhausted quota becomes an actionable error instead of a bare
  `RequestException` string.
- 404 is still left to the caller: openFDA uses it for "no matching records",
  a legitimate empty result.

Both honour `OPENFDA_TIMEOUT` (default 30s) and `OPENFDA_MAX_RETRIES`
(default 3).

## Behaviour change

```
before:  FDA_get_clinical_pharmacology_by_drug_name(...)  ->  None

after:   {"status": "error",
          "error": "openFDA OVER_RATE_LIMIT: You have exceeded your rate limit.",
          "error_details": {"code": "OVER_RATE_LIMIT", "http_status": 429,
                            "retriable": true},
          "next_steps": [...], "results": [], "meta": {...}}
```

Callers that treated `None` as "no data" will now receive a dict with
`status == "error"` and an empty `results` list.

## Tests

`tests/test_openfda_error_surfacing.py` — 11 tests covering rate limit, invalid
key, server error, non-JSON body, timeout, `NOT_FOUND` passthrough, actionable
`next_steps`, the key actually reaching openFDA, placeholder rejection, 429
retry, and that a timeout is always passed to `requests.get`.

Verified additionally by loading the pre-change module from git and running both
against an identical simulated 429 (`None` vs structured error), through the
HTTP API server, and on the live API.


gasvn:
Closing: already fixed upstream.

I raised this against a local clone that turned out to be 471 commits behind `main` (installed version 1.0.16). Both problems are already resolved in current `main`:

- `openfda_tool.py` — non-`NOT_FOUND` openFDA errors now return the shared `status`/`results`/`meta` error envelope instead of a bare `None`, and `_openfda_get()` adds the timeout and non-JSON handling (Fix-47-5).
- `fda_label_tool.py` — `_params()` attaches `FDA_API_KEY` via `_valid_api_key()`, with placeholder values rejected.

Verified by running current `main` against a simulated openFDA 429: it returns `status='error'` with `results=[]`, not `None`.

The degradation I measured (22% of FDA tool calls returning `None` under rate pressure, ~2pp accuracy loss on a 10k-question benchmark, invisible in logs) was real, but it was caused by running the stale version, not by a defect in `main`. Apologies for the noise.

One observation that may still be worth acting on: the failure was completely silent from the caller's side — an agent received empty tool results and answered from parametric knowledge with nothing in the logs. Anyone pinned to <=1.0.16 is exposed to that, so it may deserve a release note or a minimum-version warning.

gasvn:
Heads-up before anyone spends time resolving the conflicts here: **both fixes in this PR already landed on `main`**, so this looks superseded rather than mergeable.

This branch forks from `3e981b10` (2026-05-05) and is **474 commits behind** `main`. The two files it touches were reworked several times since, by #508, #514 and #517 among others — hence the 11 conflicts.

**1. Surfacing openFDA errors** — `openfda_tool.py` on `main`, in `search_openfda`:

```python
# Fix-47-1: every openFDA error *other* than NOT_FOUND used to fall
# ... The upstream error is reported instead, in the same shape the
suggestion = _build_request_error_suggestion(err)
return {"status": "error", "error": err, ...}
```

`API_KEY_INVALID` is special-cased separately. The bare `return None` on an error envelope is gone; the 8 remaining `return None` calls in that module are in unrelated helpers (`_execute_opentargets_query`, `_excipient_guard_term`, `_filter_exists`, `_build_section_note`, `_convert_id_to_drug_name`).

**2. Sending `FDA_API_KEY`** — `fda_label_tool.py` on `main`:

```python
self.api_key = os.getenv("FDA_API_KEY")

def _params(self, **kwargs):
    if _valid_api_key(self.api_key):
        kwargs["api_key"] = self.api_key
    return kwargs
```

`main` also goes further than this branch: `_valid_api_key` rejects placeholder values people leave in the env var, which this branch does not do.

So merging this as-is would risk **reverting** the more thorough versions. Suggest closing it unless there's a specific behaviour here that `main` still misses — happy to re-check if you think something is still uncovered.

The measured impact in the description (22% of FDA calls returning `None`) is worth keeping somewhere though: if that measurement predates those fixes it would be good to re-run it against current `main` to confirm the number actually dropped.

gasvn:
Followed up on this properly rather than just reading the diff — ran your test file's *behaviour* against current `main`.

Five of the six things this PR fixes are already there, and I verified each by probing `search_openfda` directly rather than trusting the code comments:

```
OVER_RATE_LIMIT  -> status='error'
API_KEY_INVALID  -> status='error'
SERVER_ERROR     -> status='error'
NOT_FOUND        -> status='error' + results:[] + suggestion
non-JSON body    -> status='error' (BAD_JSON)
```

No bare `None` on any path. `main`'s `NOT_FOUND` is actually richer than this PR's test expects — it carries `results: []`, `meta` and a suggestion pointing at alternative tools, so legitimate absence stays distinguishable from a real failure. `FDA_API_KEY` injection, placeholder rejection and `timeout=20` are all on `main` too.

**The one real gap was 429 retry** — `main` has no retry or backoff anywhere in `fda_label_tool.py`. I've opened #599 for exactly that piece, routed through the shared `request_with_retry` helper instead of a bespoke retry loop, since a dozen other modules already use it.

Your test file couldn't run as-is against `main` (it patches `OPENFDA_MAX_RETRIES` and `_fda_get`, which are introduced by this branch), which is why I probed behaviour separately.

Suggest closing this in favour of #599 — merging a 474-commit-stale branch would risk reverting the more thorough versions already on `main`. Leaving that call to you.

Still worth re-running the 22%-`None` measurement against current `main` though: if it was taken before those fixes landed, the number should have dropped a lot; if it was taken after, something is still uncovered and I'd want to chase it.

PR #607, opened 2026-09-17
Update stale tool and skill counts in the setup skill

gasvn:
## Why

`aiscientist.tools/setup.md` is the canonical install entry point — the project's own docs convention names it as the single setup path. It was telling every new user:

> ToolUniverse has **1200+ tools** … the MCP server plus **100+ skills**

v1.5.0 ships **2,716 tools and 187 skills**. The first thing a new user reads understated the project by more than half.

Counts were verified by loading each release and counting, not estimated:

| | tools | skills |
|---|---|---|
| v1.4.1 | 2,602 | 155 |
| v1.5.0 | 2,716 | 187 |

Rounded down to `2,700+` and `180+` so they stay true for a while rather than needing an edit every release. Applied to all three synced copies (`skills/`, `plugin/`, `plugins/`).

## Checked and deliberately left alone

- **Every external link still resolves.** `omim.org/api` returns 403 to scripted requests, but that is bot filtering, not a dead link — same behaviour as several other sites that work fine in a browser. Not "fixed".
- **All nine documented CLI subcommands still exist** (`list grep find info run test status build serve`).
- The skill does not mention ToolUniverse Platform or Connect anywhere. The `connect`/`share` hits in it are ordinary English (shared API keys, shared skills).

## Two things I did not change, because they are your call

**1. The CLI now has 15 subcommands, not 9.** The frontmatter says "CLI (command line with 9 subcommands)". The nine listed all still work, but `remote`, `doctor`, `connect`, `connections` and `disconnect` were added since.

I left it, because those additions are mostly Platform-related and Platform is not released yet — updating the count would advertise it early. Worth revisiting once Platform ships.

**2. The advertised tool count is inconsistent across the repo**, and unifying it is a product decision rather than a typo fix:

| where | says |
|---|---|
| `README.md` | 1000+ tools |
| `docs/` (44 occurrences) | 1000+ tools |
| `mcpb/manifest.json` | 2,500+ scientific tools |
| setup skill (before this PR) | 1200+ |
| **actual** | **2,716** |

`skills/devtu-docs-quality/SKILL.md` even codifies a rule — *Always "1000+ tools". Never "600+", "750+", "1200+"…* — which the setup skill was violating, and which is itself now badly understated. It is a docs convention only; nothing in CI enforces it.

I fixed only the file you asked about rather than sweeping 44 docs files and rewriting that rule, since which number the project advertises is yours to decide. Happy to do the sweep once you pick one.


gasvn:
Updated the CLI subcommand count too, in `c0368a3a`.

**It's 14, not 15** — I said 15 earlier from memory and that was wrong. Counted properly from `tu --help` in the **published 1.5.0 wheel from PyPI** (not the source tree):

```
list grep info find run test status build remote doctor serve connect connections disconnect
```

`remote`, `doctor`, `connect`, `connections` and `disconnect` were added since the skill was written. Each of the fourteen in the new list was checked back against that output.

**Why I originally held this back, and why that was wrong.** I assumed these additions were tied to ToolUniverse Platform and that updating the count would advertise it before release. I checked, and the assumption doesn't hold — the commands are already visible in `tu --help` for anyone who installed 1.5.0, and the backend they talk to is live (`POST /tools/call` returns `401 Not authenticated` rather than a 404 or a connection failure, and `connect.aiscientist.tools/api-keys` serves a 200). So the docs were simply understating what already shipped.

The advertised-tool-count inconsistency in the PR description still stands as a separate decision — 44 docs files say `1000+`, the mcpb manifest says `2,500+`, and `devtu-docs-quality` codifies a rule that the setup skill was violating.

PR #611, opened 2026-09-18
fix: coerce HumanBase gene_list strings into gene names

krudo-taco:
## Summary

HumanBase iterated `gene_list` directly. Agents often pass a string (`"TP53"` or `"BRCA1,TP53"`), so `get_official_gene_name` ran on each character and the network query was nonsense.

Coerce a string (comma or semicolon separated) into gene symbols; keep real lists. Missing/blank input returns a local error.

## Validation

- [x] Relevant unit tests: 8 passed (`tests/unit/test_humanbase_gene_list_string.py`).
- [x] `humanbase_ppi_retrieve` is mocked; the test asserts the coerced list is `["TP53"]`.
- `git diff --check`: passed.

Full-file `ruff check` of `humanbase_tool.py` still reports pre-existing line-length diagnostics; no full-file lint pass is claimed.

## Checklist

- [x] No docs/schema change needed: `gene_list` / `genes` names are unchanged.
- [x] Added coercion unit tests and a run-path regression.


gasvn:
Merged — thanks @krudo-taco. The bug you identified is real: an agent passing `"TP53"` instead of `["TP53"]` had every character queried separately.

One thing worth flagging, because it applies to several of your PRs: the coercion could not actually run as merged. `gene_list` is declared `"type": "array"`, so validation rejected a string before `run()` was entered:

```
$ tu run humanbase_ppi_analysis '{"gene_list":"TP53","tissue":"brain"}'
Error: Parameter validation failed for 'gene_list': 'TP53' is not of type 'array'.
```

I've merged #621 on top of yours to finish it — it widens the schema to accept a string, which is what makes `_coerce_gene_list` reachable. It also handles a case yours didn't: `["BRCA1,TP53"]`, where the comma sits inside a list element. That was resolved as a single symbol, fuzzy-matched to **BRIP1**, and returned a BRCA1/BRCA2/BRIP1/CCNE2/CHEK1/WDHD1 network with no TP53 in it — under a header still reading `Query Genes: BRCA1,TP53`. Both now work:

```
gene_list "TP53"          -> TP53 network
gene_list ["BRCA1,TP53"]  -> BRCA1 and TP53 both present
```

The general lesson is worth carrying into future PRs: when a fix handles an input type, check the schema permits that type, or the code is unreachable. That is also why I closed #613 — `max_node` carries `"minimum": 1`, so its zero-handling could never execute either.

PR #612, opened 2026-09-18
fix: honor zero OmniPath interaction limits

krudo-taco:
## Summary

OmniPath dropped `limit=0` because `if arguments.get("limit")` is false for zero. Ligand-receptor and the other endpoints then omitted the API parameter and returned the untruncated dump.

Forward any explicit `limit`, including zero.

## Validation

- [x] Relevant unit tests: 1 passed (`tests/unit/test_omnipath_zero_limit.py`).
- [x] `_make_request` is mocked; the test asserts `params["limit"] == "0"`.
- `git diff --check`: passed.

## Checklist

- [x] No docs/schema change needed: public `limit` is unchanged.
- [x] Added a regression for `limit=0` on ligand-receptor.


gasvn:
Thanks @krudo-taco — you found a real inconsistency here, and I checked the upstream behaviour you inferred: OmniPath genuinely honours `limit=0`, returning 0 rows where an unlimited query returns 37. So the PR does what it claims, unlike #614.

I'd still rather go the other way, and refuse zero instead of forwarding it.

`limit=0` returns `{"status": "success", "data": []}` with `metadata.total_interactions: 0`. For EGFR — one of the best-characterised receptors in the dataset — an agent reads that as "no ligand-receptor interactions", and nothing in the payload shows the emptiness was self-inflicted. `limit=-1` is worse: it goes upstream and comes back a bare HTTP 500 naming no parameter.

`humanbase_ppi_analysis` already declares `"minimum": 1` on `max_node` and refuses zero with a clear message. Matching that is the more consistent answer, and it makes the false negative impossible rather than merely correct.

So I've put `minimum: 1` on `limit` across the four OmniPath tools that expose it, in #622, now merged. That makes the change here unreachable — zero can no longer arrive — so I'm closing this one.

#622 also makes the gene parameters accept a list. `partners: ["EGFR"]` was rejected as `not of type 'string', 'null'`, while the other interactome tool in the plugin takes a list and refuses a string, so moving between them costs a round trip in each direction. They now take either.

Thanks for pushing on the zero cases across these tools — the pattern was worth surfacing even where the resolution went the other way.

PR #613, opened 2026-09-18
fix: honor zero HumanBase max_node

krudo-taco:
## Summary

`max_node = arguments.get("max_node") or arguments.get("top_n") or 10` treated `0` as missing, so HumanBase still requested a default-sized network. Keep an explicit zero.

## Validation

- [x] Relevant unit tests: 1 passed (`tests/unit/test_humanbase_max_node_zero.py`).
- [x] `humanbase_ppi_retrieve` is mocked; the test asserts `max_node == 0`.
- `git diff --check`: passed.

## Checklist

- [x] No docs/schema change needed: `max_node` / `top_n` names are unchanged.
- [x] Added a regression for `max_node=0`.


gasvn:
Thanks @krudo-taco. This one I'm going to close, and the reason is the same as on #611.

`max_node` carries `"minimum": 1` in the schema, so a zero never reaches `run()`:

```
$ tu run humanbase_ppi_analysis '{"gene_list":["TP53"],"tissue":"brain","max_node":0}'
Error: Parameter validation failed for 'max_node': 0 is less than the minimum of 1
```

The zero-handling can't execute through the CLI or MCP. You're right that the API honours `node_size=0` — I checked, and it returns just the query gene with no partners — but the constraint is deliberate: a network of zero partners isn't a useful request, and refusing it up front is better than answering it.

Worth noting the error you'd get is already a good one. It names the parameter, the value and the bound, which is more than several other tools manage — OmniPath's `limit` had no minimum at all, so `limit=0` there returned an empty list with `status: success`, reading as "EGFR has no ligand-receptor interactions". That one now has the same `minimum: 1` this tool already carried, in #622.

Closing with thanks — the surrounding investigation was useful even though this particular change is a no-op.

PR #614, opened 2026-09-18
fix: send Europe PMC annotation pageSize including zero

krudo-taco:
## Summary

Europe PMC batch annotation search treated `page_size=0` as missing (`if page_size:`), so the request used the API default page and looked complete. Forward any explicit `pageSize`, including zero.

## Validation

- [x] Relevant unit tests: 1 passed (`tests/unit/test_europepmc_annotations_page_size.py`).
- [x] `requests.get` is mocked; the test asserts `pageSize == 0`.
- `git diff --check`: passed.

## Checklist

- [x] No docs/schema change needed: public `page_size` is unchanged.
- [x] Added a regression for `page_size=0`.


gasvn:
Thanks @krudo-taco — but I don't think this one is doing what the description says, and I'd like to point at where the real problem is instead.

**`annotationsByArticleIds` does not paginate at all.** The test mocks `requests.get`, so it proves `pageSize` reaches the query string — but not that the endpoint acts on it. Against the live API with four article IDs:

| pageSize | articles | annotations per article |
|---|---|---|
| unset | 3 | 274 / 201 / 67 |
| 0 | 3 | 274 / 201 / 67 |
| 1 | 3 | 274 / 201 / 67 |
| 2 | 3 | 274 / 201 / 67 |

`pageSize` pages neither articles nor annotations. So forwarding an explicit zero is indistinguishable from dropping it, and the premise — that the request "used the API default page and looked complete" — doesn't hold here. There is no default page to fall back to.

**There is a real bug next to it, and it's worse than the one described.** The schema advertises `Number of results per page. Default: 25. Max: 1000.`, none of which is true, and `_batch_by_type` truncates with a hard-coded `annotations[:100]` the caller can't reach. Asking for 25 gets 100; asking for 500 also gets 100. `annotation_count` is computed before that slice, so an article can report `274` next to a list of 100.

I've merged #620 to fix that: `page_size` now sets the per-article cap — the only place the number can act, given the endpoint returns everything in one response — the description matches the behaviour, the dead parameter is dropped, and `annotation_count` counts what you actually receive while each article reports its own total.

Chasing this turned up two more faults in the same tools, both in that PR. `'Genes & Proteins'` has been a documented `annotation_type` all along, but Europe PMC spells it `Gene_Proteins` and answers an unrecognised type with an empty result rather than an error — so the documented value returned 0 where the real one returns 131. And every HTTP failure reported `HTTP error: unknown`, because `requests.Response.__bool__` is `ok`, so `if e.response` is False for precisely the 4xx responses that carry the explanation. That one is relevant to your #608: the error now reads

```
HTTP error 400: The articleIds list parameter contains one or more wrongly
formatted values (i.e. PMC:PMC:PMC4353746)...
```

which names the double-prefix bug your PR fixes, instead of hiding it.

So I'm going to close this one, with thanks — finding the inert parameter is what led to the rest.

**One request, since this keeps coming up.** Comparing this with your #578 shows the difference exactly. There you checked what the API does with zero and short-circuited before the request — and you were right: OpenAlex rejects `per_page=0` with `400 Pagination error: per-page parameter must be between 1 and 200`. That guard is load-bearing. Here the same shape of claim went the other way because only a mock was consulted.

When a PR asserts what an upstream API does with a parameter, could you include a live request and its response in the body? A single `curl` with the before/after counts would have settled both cases in a minute. Your local-logic fixes have been consistently strong — #596 in particular, where `size` was silently ignored and search was returning 1000 records, was a genuinely valuable catch — so this is narrowly about the API-behaviour claims.

For transparency on the same standard applied to merged work: #573's `failedIds` condition never fires, because UniProt's status response always carries a `results` key (checked across four failed-only job shapes), so the existing `"results" in status_data` already exits the loop. It's harmless and I'm leaving it. #577 did land a real regression — the two-letter tokens meant `of` and `in` survived as search terms and silently disabled guideline relevance filtering — which I fixed in #581.

On the rest of your open set: #608, #609 and #610 are merged — #609/#610 in particular fail fast in ~1.3s where the alternative is a 22s bulk download, which is exactly right. #611 is merged too, with #621 on top to finish it.

#612 and #613 I've closed, and the reason is the same one as here. `humanbase_ppi_analysis` declares `max_node` as `"minimum": 1`, so `max_node: 0` never reaches `run()`, and OmniPath's `limit` now carries the same bound. In both cases the zero-handling cannot execute through the CLI or MCP. Details are on each PR.

PR #624, opened 2026-09-21
Feature: exact RxNorm lookup that reports every match instead of choosing one

gasvn:
## Why

`RxNorm_find_rxcui` searches with normalization (`search=2`) and returns the first identifier as `primary_rxcui`, without saying what kind of concept it is. That's fine for "give me an RXCUI", but it's wrong for a caller that has to know what a name refers to before acting on it. Observed on the live API:

| input | `RxNorm_find_rxcui` | what the name actually is |
|---|---|---|
| `ceftriaxon` (typo) | 2193 | not an RxNorm string; silently corrected to ceftriaxone |
| `Tylenol` | 202433 | a brand name (BN), not an ingredient |
| `heparin` | first of 2 | ingredient 5224 (IN) **and** `heparin, porcine` 235473 (PIN) |

## What

A new `get_exact_concepts` operation on `RxNormExtendedTool`, exposed as `RxNorm_get_exact_concepts`. It:

- looks the name up in exact mode (`search=0`, `allsrc=0`);
- fetches properties for **every** match (up to 5) and returns them all, with `tty`, `term_type_label` and `suppress`;
- returns an explicit `resolution` and never picks a match:
  - `unique_active_IN_concept`: exactly one active ingredient; `canonical_ingredient_name` is its RxNorm name (`paracetamol` → `acetaminophen`)
  - `requires_identity_review`: a single concept that is not an active ingredient (brand, salt, …)
  - `multiple_concepts`: more than one match, none selected
  - `no_exact_concept`

The only normalization is trimming surrounding whitespace; RxNorm's exact mode is already case-insensitive. A malformed identifier list, or properties that belong to a different RXCUI, is an error rather than an empty answer, and every error keeps the request trail in `metadata.requests`.

No new tool class and no new dependency. `RxNorm_find_rxcui` is unchanged.

## Verification

- `tests/unit/test_rxnorm_exact_concepts.py` (16 tests, mocked HTTP with payloads taken from the live responses): exact-mode parameters, each resolution, synonym → canonical name, too many matches, whitespace, blank input, network failure, mismatched properties, malformed IDs. Every returned payload is validated against the declared `return_schema`.
- `tu test RxNorm_get_exact_concepts`: all 3 examples pass live.
- Live CLI: `paracetamol` → IN 161 / acetaminophen; `Tylenol` → BN, review; `heparin` → IN + PIN, multiple; `ceftriaxon` → no exact concept.
- `tests/unit` fails on the same set of tests as `main` (remote-tool and Boltz tests that need services not available locally).


gasvn:
Rebased onto `main` and added the generated SDK wrapper.

The green CI on this branch predated #629, which added a check that every built-in tool has a generated wrapper. Against current `main` this branch failed it:

```
1 built-in tools have no generated wrapper: ['RxNorm_get_exact_concepts']
from tooluniverse.tools import RxNorm_get_exact_concepts  -> ImportError
```

Without the wrapper the tool would be served by the CLI and MCP while being invisible to the Python SDK — the same gap that left 119 tools unreachable before #629.

`python scripts/build_tools.py` generated exactly one tool. Because the rebase also picks up #630, the metadata gains a single line instead of being rewritten:

```
+  "RxNorm_get_exact_concepts": "d6d4cfbb5ca9539101131cd12eea1d92",
```

Verified: guard tests pass, full unit suite passes, a second build produces no diff, and the four live cases still behave as described — `paracetamol` → IN/acetaminophen, `Tylenol` → BN/review, `heparin` → PIN+IN/multiple, `ceftriaxon` → no exact concept.

PR #632, opened 2026-09-21
Add Firecrawl Search as an opt-in web search backend

JuampiHernandez:
## Summary

I'm Juan from Firecrawl. This adds our Search API as another opt-in backend for `web_search`, next to Parallel and SerpBase. It runs keyless like Parallel, with an optional key for higher limits, and it takes both the `region` and `safesearch` controls (Parallel rejects them, SerpBase takes `region` only).

- add `backend="firecrawl"` to `web_search` and `web_api_documentation_search`, calling Firecrawl Search (`POST https://api.firecrawl.dev/v2/search`) over REST
- opt-in only. The DDGS default and the `auto` chain are unchanged, and `firecrawl` never enters the default chain
- works without a key (documented keyless tier, capped per IP per day, https://docs.firecrawl.dev/rate-limits). `FIRECRAWL_API_KEY` is optional; with it, searches bill the account's credits and use the account's rate limits
- `region` maps to Firecrawl's ISO `country` (country half only, `uk` to `GB`, `ja` to `JP`; the language half is not sent). `safesearch` maps to Firecrawl's boolean `safe` (`on` and `moderate` enable it, `off` leaves it unset). Parameter descriptions say so
- on any failure it falls back to `auto` and keeps Firecrawl's own error text in `provider_errors`, same as the `parallel` and `serpbase` branches. The 429 hint depends on whether a key was set
- `provider_notice` is attached whenever `firecrawl` was attempted, fallback included
- `FIRECRAWL_API_KEY` registered under `optional_api_keys` with `api_key_info`; `.env.template`, `api_keys_catalog.json` and tool hashes regenerated with the scripts, not hand-edited. The "Other" block in `.env.template` moves because the generator orders unlisted domains by first key alphabetically

## API contract

Request `{"query", "limit", "sources": ["web"], "country", "safe", "highlights": false}`, Bearer header optional (`highlights: false` asks for plain provider snippets instead of the default multi-KB highlight blocks, so `snippet` stays comparable to the other backends). Success is `{"success": true, "data": {"web": [{"url", "title", "description", "position"}]}}`. Bad key is 401, out of credits is 402, cap exceeded is 429, unknown keys are 400, all with `{"success": false, "error": ...}` in the body, so the body is read before the status is enforced. I ran all six `region` values against the live endpoint to confirm the `country` mapping.

The key is read with `os.environ.get`, same as SerpBase on `main`. If #617 lands first I'll rebase onto `self.credential()`. And if you'd rather reject non-default `region` the way Parallel does instead of mapping the country half, happy to switch.

## Validation

- [x] `ruff check` passed for touched Python files (and `ruff format --check`, with the pinned 0.14.7)
- [x] `pytest` passed for relevant unit/integration tests: `tests/tools/test_web_search_tool.py -m "not network"` 68 passed (33 existing, 35 new); `ToolUniverse().load_tools()` loads all 2718 tools with the key unset
- [x] Generated SDK/tool files were rebuilt when tool metadata changed (`scripts/gen_api_key_catalog.py`, `scripts/build_tools.py`)
- [x] Network/API-key tests are marked or isolated from default PR checks: the live test carries `integration` + `network` and forces keyless so CI never spends a real key. It passes from a residential IP and skips with a clear reason if the keyless per-IP cap is already hit, so it won't go red from a shared runner. Live calls through ToolUniverse keyless and with a key also pass

## Checklist

- [x] Linked the relevant issue or described the user-facing problem
- [x] Updated docs, examples, or tool metadata for behavior changes
- [x] Added or updated tests for new tools, providers, or schema behavior
- [x] Avoided committing credentials, cached API responses, or large generated artifacts

One more thing. Firecrawl also has a hosted MCP server (search, scrape, crawl, PDF parsing), similar to what the Exa loader does here. If that's useful I can send it as a separate PR. Wanted to keep this one to the search backend.


gasvn:
Verified locally against a clean merge of this branch into current `main` (no conflicts).

**Behavior, live:**

* Keyless `backend='firecrawl'`: `status: success`, `backend_used: firecrawl`, 3 real results with plain snippets, `provider_notice` attached, `provider_errors` empty.
* With an invalid key: falls back to `auto`, returns results, and keeps the provider message — `{'firecrawl': 'Firecrawl Search API error (HTTP 401): Unauthorized: Invalid token'}`.
* Default call (no `backend`): `attempted_backends` is `['duckduckgo', 'auto']` and no `provider_notice` — the default chain is untouched, as the PR says.
* `web_api_documentation_search` with `backend='firecrawl'` routes through and returns results, so the new enum value is real on both tools.
* `_firecrawl_country_from_region` over the full `region` enum: `us-en→US, cn-zh→CN, uk-en→GB, de-de→DE, fr-fr→FR, ja-jp→JP`. Every enum value is a valid ISO code after mapping; nothing in the enum falls through to the `US` default.

**Checks:**

* `tests/tools/test_web_search_tool.py -m "not network"`: 68 passed.
* `scripts/gen_api_key_catalog.py` and `scripts/build_tools.py` rerun on the merge result leave the tree clean (2719 tools unchanged), so the generated files in the diff match what the generators produce.

**Two notes, neither blocking:**

1. `safesearch='moderate'` (the default) maps to Firecrawl's `safe: true`, so an opt-in Firecrawl search filters more aggressively than the same call on a DDGS backend. The conservative direction is the right one and the parameter description says so — flagging it only because it is the one place where the same arguments mean something different per backend.
2. On the credential question: read the key through `self.credential()` rather than `os.environ.get` once #617 lands, as you offered. Whichever of the two merges second should carry that change.

The hosted MCP server is worth a separate PR, yes — send it standalone as you suggested.

Thanks for the thorough write-up, especially the API-contract section and having run all six `region` values against the live endpoint.

gasvn:
Pushed one commit to this branch (8f15383) rather than leaving it as a request, since you have maintainer edits on — revert it if you disagree.

It widens the live keyless check's skip. It skipped only on a 429 in `provider_errors`, so every other provider-side outcome failed the test instead: a 402, a blocked or shared runner IP, a Firecrawl outage. `tests/tools` runs in this repo's optional network job, so those would have turned that job red for reasons outside the repository. The skip now keys on `backend_used`, which is the signal the fallback already reports, and carries Firecrawl's own message into the skip reason. Everything after the guard is unchanged.

Verified both branches locally: against the real endpoint the test passes, and with the endpoint pointed at a closed port it skips with `Firecrawl did not serve this search: HTTPConnectionPool(...) Connection refused` where it previously failed. `ruff check`, `ruff format --check`, and the full file (68 passed) are clean.

Still open on my side: the `self.credential()` change once #617 lands, as you offered. I'd rather land #632 after #617 for that reason — there's no rush from this end.

gasvn:
#617 is merged, so I pushed the rebase-equivalent here: `main` merged into the branch (no conflicts) plus the credential change you offered (85353d6).

It was not optional in the end. #617 removed `import os` from `web_search_tool.py` because SerpBase now reads its key through `self.credential()`, so once `main` merged in, the Firecrawl branch's `os.environ.get` was an undefined name — `ruff` reports F821 and the first keyed search would have raised. The key now resolves with `self.credential("FIRECRAWL_API_KEY")`, same as the other backends: outside a request scope it still falls back to the environment, so local use is unchanged, and inside one the request's own key is used.

Added two tests for the new path, in the style of the ones already in the file: the request-scoped key wins over an environment key, and an active scope that carries no Firecrawl key sends no `Authorization` header while the keyless search still runs (fail closed — a request must not spend the operator's key).

Checks after the merge: `tests/tools/test_web_search_tool.py` 70 passed (68 plus the two new), the live keyless check passes against the real endpoint, `ruff check` and `ruff format --check` clean, `ToolUniverse().load_tools()` loads 2717 tools. The PR is `MERGEABLE` again.

I left `.tool_metadata.json` as the merge produced it. Running `build_tools.py` on this branch also rewrites eight SemanticScholar hashes, but that drift is inherited from `main` and is not yours to carry — it is being fixed separately.

Nothing outstanding from my side on the code. Whether ToolUniverse takes a fourth search backend is a maintainer call I'll leave to the repo.

JuampiHernandez:
Thanks for landing this and for carrying the `self.credential()` change, I would have missed the `import os` removal. The wider skip on the live test is the right call too.

I'll put the MCP loader together as a separate PR, following the Exa loader shape.

